PR #43, opened 2026-04-02
hotfix: restore local data path compatibility

TheLonelyDevil9:
## Summary
- restore local/git installs to the pre-regression storage root under `packages/server/data`
- resolve relative `DATABASE_URL=file:...` paths from the server package root again so the app, backups, encryption key, and Drizzle CLI stay coherent
- add startup diagnostics and recovery guidance for users who now have both `packages/server/data/` and `data/`

## Why
A recent path-selection regression made some installs open a different SQLite database than the one users had been using before. That looks like data loss even when nothing was deleted.

The core objective of this hotfix is to restore visibility of existing user data immediately and safely. It deliberately avoids automatic migration so we do not overwrite, merge, or move the wrong database under uncertainty.

## Validation
- verified runtime resolution with no env override: `DATA_DIR` and DB path resolve to `packages/server/data`
- verified copied `.env` case with `DATABASE_URL=file:./data/marinara-engine.db` still resolves to `packages/server/data`
- verified explicit `DATA_DIR` override still controls both data dir and DB path
- verified dual-location warning when both `packages/server/data/marinara-engine.db` and `data/marinara-engine.db` exist
- ran `pnpm check`

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -13,7 +13,7 @@ android {
         minSdk 24
         targetSdk 34
         versionCode 2
-        versionName "1.4.6"
+        versionName "1.4.7"
         buildConfigField "String", "MARINARA_SERVER_URL", "\"http://localhost:${marinaraPort}\""
     }
 


TheLonelyDevil9:
Added some context as to why

PR #44, opened 2026-04-02
fix: stabilize storage paths and release metadata

TheLonelyDevil9:
## Summary
- restore local data-path compatibility so existing installs keep resolving storage under `packages/server/data`
- make update checks use the newest GitHub `v*` tag with release metadata fallback instead of trusting `releases/latest`
- add release guardrails for tracked installer binaries and changelog-driven GitHub Release publishing

## Why
This PR exists to remove two classes of operational drift that directly affect user trust and maintainer reliability.

First, local installs need stable storage resolution so updates do not make user data appear to vanish by switching databases. Second, release and update metadata need to reflect the real newest version even when GitHub Releases lag behind tags, otherwise the in-app updater and release process can report the wrong state.

The core objective is to make local installs safer and the release pipeline more truthful, so the app behaves consistently and the project publishes version information from a single reliable flow.

## Changes
- reapply the storage compatibility hotfix from PR #43, including startup storage diagnostics and aligned Drizzle/runtime path resolution
- route `ADMIN_SECRET` and `ENCRYPTION_KEY` through the runtime config layer
- sync Android `versionName` to `1.4.7`
- remove the tracked installer `.exe` from source control
- add `guard:installer-artifacts` and `release:notes` scripts
- add PR CI enforcement for tracked installer artifacts
- add a tag-triggered GitHub Release workflow that renders release notes from `CHANGELOG.md`
- update README and contributor docs to match the new update and release behavior

## Validation
- `pnpm guard:installer-artifacts`
- `pnpm version:check`
- `pnpm release:notes -- 1.4.7`
- `pnpm check`

### DIFF
--- a/packages/server/src/routes/generate.routes.ts
+++ b/packages/server/src/routes/generate.routes.ts
@@ -38,7 +38,7 @@ import type { LLMToolDefinition, ChatMessage, LLMUsage } from "../services/llm/b
 import { executeToolCalls } from "../services/tools/tool-executor.js";
 import { createAgentPipeline, type ResolvedAgent } from "../services/agents/agent-pipeline.js";
 import { DATA_DIR } from "../utils/data-dir.js";
-import { executeAgent, executeAgentBatch } from "../services/agents/agent-executor.js";
+import { executeAgent } from "../services/agents/agent-executor.js";
 import {
   parseCharacterCommands,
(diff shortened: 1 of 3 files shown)


TheLonelyDevil9:
Added some context as to why

PR #102, opened 2026-04-19
feat(game): movable lockable HUD panels + layout/drag fixes

LukaTheHero:
## Summary

Adds a small lock icon next to the collapse toggle on every Game Mode HUD panel (map, Bonds, Leads, Mana Stability, Exposure Risk). Panels are locked by default; clicking the lock swaps it to an open padlock, draws a faint ring around the card, and makes it free-draggable within the game surface. Lock state and position persist **per chat** so layouts don't leak between games.

## What's in the box

**New:** `packages/client/src/components/game/DraggablePanel.tsx`
- `useDraggablePanel(chatId, panelId)` hook — returns `{ locked, toggleLocked, x, y, handleDragEnd }`. Storage key is `marinara-game-panel:<chatId>:<panelId>`, so every game gets its own HUD layout.
- Reads localStorage synchronously via a `useRef` seed on first render, so moved panels paint at their saved position with no origin-flash flicker.
- `PanelLockButton` — lucide `Lock`/`Unlock` toggle styled to match the existing chevron/collapse indicators (`text-white/30 hover:text-white/70`), with `stopPropagation` so it doesn't fire the header's collapse click.

**Updated:** `GameWidgetPanel.tsx`, `GameMap.tsx`
- Cards render as `motion.div` with `drag={!locked}`, `dragMomentum={false}`, `dragElastic={0}`, and `dragConstraints` pointed at the game surface — **panels can't be dragged off-screen**.
- `WidgetCard` keyed by `${chatId}:${widget.id}` so switching games remounts with fresh motion values.
- Map card styling (`w-52`, border, bg, shadow, blur) moved from the `GameSurface` wrapper onto `GameMapPanel`'s draggable root, so the whole visual card translates together.
- `data-tour="game-map"` moved from the static wrapper onto the draggable map root, so the onboarding tutorial highlight tracks it wherever it's dragged.

**Updated:** `GameSurface.tsx`
- New `hudSurfaceRef` on the main game content `<div>`, passed as `constraintsRef` to both map and widget panels.
- `activeChatId` threaded down as `chatId` to both panels for storage scoping.
- HUD flex container: `items-start` → `items-end` (fix below).

## Pre-existing bugs fixed along the way

- **Column reshuffle on expand/collapse** — the HUD flex container was `items-start` but anchored at `bottom-24` with no fixed height, so growing the left column pushed the container top upward and dragged the top-aligned right column with it. Switched to `items-end` so both columns bottom-anchor and grow independently upward.
- **Drag lag** — `WidgetCard` had `transition-all`, which made the browser tween every transform update framer-motion pushed during drag. Narrowed to `transition-colors` so only color state animates; transform updates are immediate.

## Test plan

- [x] `pnpm check` passes (client build + typecheck clean)
- [x] All five desktop panels show a lock icon and are locked by default
- [x] Clicking lock → open padlock + ring + grab cursor; drag works; re-lock restores normal behavior
- [x] Collapse/expand still toggles whether locked or unlocked
- [x] Lock state and drag position persist across reload
- [x] Switching chats resets layout to that chat's saved positions (doesn't carry over)
- [x] Cannot drag any panel off-screen (clamped to game surface)
- [x] No paint-flicker when reloading a chat with moved panels
- [x] Expanding/collapsing a widget on the left no longer shifts the right column (and vice versa)
- [x] Drag feels smooth — no CSS-transition lag
- [x] Tutorial highlight still targets the map after it's been moved

### DIFF
--- a/packages/client/src/components/game/DraggablePanel.tsx
+++ b/packages/client/src/components/game/DraggablePanel.tsx
@@ -1,0 +1,115 @@
+// ──────────────────────────────────────────────
+// Game: Lock + drag helpers for HUD panels
+//
+// Each panel (widget cards, map) uses `useDraggablePanel`
+// to persist a lock flag and {x,y} offset. State is scoped
+// by chatId so positions don't bleed across games.
+// `PanelLockButton` renders the lock toggle in headers.
+// ──────────────────────────────────────────────
+import { useCallback, useRef, useState } from "react";
+import { useMotionValue } from "framer-motion";
+import { Lock, Unlock } from "lucide-react";
+import { cn } from "../../lib/utils";
+
+const STORAGE_PREFIX = "marinara-game-panel:";
+
+interface PanelState {
+  locked: boolean;
+  x: number;
+  y: number;
+}
+
+function storageKey(scopeId: string, panelId: string): string {
+  return `${STORAGE_PREFIX}${scopeId}:${panelId}`;
+}
+
+function readPanelState(key: string): PanelState {
+  if (typeof window === "undefined") return { locked: true, x: 0, y: 0 };
+  try {
+    const raw = window.localStorage.getItem(key);
+    if (!raw) return { locked: true, x: 0, y: 0 };
+    const parsed = JSON.parse(raw) as Partial<PanelState>;
+    return {
+      locked: parsed.locked !== false,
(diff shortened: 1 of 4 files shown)


SpicyMarinara:
Thank you!

PR #106, opened 2026-04-19
fix(sidecar): Replace native llama addon rebuilds with downloadable llama-server runtime

munimunigamer:
## Summary

This replaces the old `node-llama-cpp` sidecar path with a downloaded `llama-server` runtime and keeps the local-model flow working without native compilation on end-user machines.

It also adds support for using curated Gemma presets or custom GGUFs from HuggingFace, plus a clearer setup flow in the download modal.

## Why

The previous sidecar flow depended on rebuilding native llama bindings locally, which was fragile across user machines and especially painful on Windows and other environments without the right toolchain.

This change moves Marinara to the official upstream `llama.cpp` server binaries instead:
- no local C++ build step
- no `node-llama-cpp` dependency
- cleaner crash isolation via subprocess
- easier support for newer llama.cpp model/runtime updates
- better platform coverage

### Server
- Replaced in-process sidecar inference with a spawned `llama-server` subprocess
- Added runtime install management for official `llama.cpp` release assets
- Added runtime detection/status reporting
- Added BYO GGUF listing + download flow from HuggingFace repos
- Migrated sidecar config to explicit `modelPath`
- Preserved sidecar inference API shape for existing callers where possible
- Added restart/unload/health handling for the sidecar process
- Added cleanup for legacy sidecar runtime stamp behavior

### Client
- Updated sidecar store for new runtime/model lifecycle states
- Added runtime-aware modal UI for:
  - curated model downloads
  - custom HuggingFace GGUF selection
  - active setup progress
  - runtime/server error visibility
- Disabled finishing setup until the sidecar is actually ready
- Added a clearer “setup in progress” flow with cancel behavior while runtime/model startup is happening

### Launcher / packaging
- Removed sidecar native rebuild steps from:
  - `start.bat`
  - `start.sh`
  - `start-termux.sh`
- Removed `node-llama-cpp` from package manifests and lockfile
- Kept app versioning unchanged

## Platform/runtime support

This runtime-selection flow now supports official download/install paths for:
- Windows
- macOS
- Linux
- Android/Termux

With runtime matching/fallback for available upstream variants such as:
- CPU
- CUDA
- Vulkan
- ROCm
- ARM64 where upstream assets exist

## User impact

Users should now be able to enable the local sidecar without needing to compile llama.cpp locally.

They can also:
- keep using curated Gemma sidecar presets
- use their own GGUF from a HuggingFace repo
- see clearer setup progress while the runtime is downloading/starting

## Validation

Ran:
- `pnpm --filter @marinara-engine/shared build`
- `pnpm --filter @marinara-engine/server build`
- `pnpm --filter @marinara-engine/client build`

Manual:
- verified the updated sidecar setup modal flow on the cleaned branch

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,8 +31,7 @@
     "db:push": "pnpm --filter @marinara-engine/server db:push",
     "db:generate": "pnpm --filter @marinara-engine/server db:generate",
     "db:migrate": "pnpm --filter @marinara-engine/server db:migrate",
-    "clean": "pnpm -r --if-present run clean && node ./scripts/clean-workspace.mjs",
-    "sidecar:build": "node ./scripts/build-sidecar-runtime.mjs"
+    "clean": "pnpm -r --if-present run clean && node ./scripts/clean-workspace.mjs"
   },
   "devDependencies": {
     "esbuild": "^0.27.3",
@@ -42,7 +41,6 @@
     "onlyBuiltDependencies": [
       "better-sqlite3",
       "esbuild",
-      "node-llama-cpp",
       "sharp"
     ],
     "overrides": {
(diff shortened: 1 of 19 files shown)


munimunigamer:
lookin into copilot suggestions

munimunigamer:
Should be good now!

SpicyMarinara:
Thank you!

PR #116, opened 2026-04-19
Add -sm none to llama-server args

guybrush01:
There is a [known issue with multiple GPUs and Gemma 4 regarding split mode](https://github.com/ggml-org/llama.cpp/issues/21730).  This pull request adds the no split mode flag to prevent the looping crash.


SpicyMarinara:
Good catch, I had no idea that was a thing. Thank you for the PR, adding you as a contributor!

PR #156, opened 2026-04-21
Fix {{user}} / {{char}} macro substitution in Game Mode

Romuromylus:
## Summary

In Game Mode, `{{user}}` and `{{char}}` macros appear literally in the rendered narration instead of resolving to the active persona/character.

## Root cause

`GameNarration.tsx` is the rendering chokepoint for Game Mode narration and player-action segments, but it never runs display-time macro replacement. The chat path handles this in `ChatMessage.tsx` (around line 650, `text.replaceAll(\"{{user}}\", userName)`), and `GameNarration.tsx` does not call into that path. The server-side `resolveMacros` engine doesn't catch it either, since Game Mode replaces the system prompt entirely with `buildGmSystemPrompt(gmCtx)` (in `generate.routes.ts`), whose output never goes through `resolveMacros`. Net effect: `{{user}}` / `{{char}}` survive end-to-end and render verbatim in the UI.

This is a missing feature, not a regression — Game Mode shipped without macro display-substitution from v1.5.0 onward.

## Fix

Apply the substitution at segment-build time inside the `segments` and `sideLineMap` useMemos in `GameNarration.tsx`. Doing it there (rather than inside `formatNarration`) means every downstream render path — `formatNarration`, `animateTextHtml`, side-line overlays — receives already-resolved text without changing the renderer's signature.

- `{{user}}` → `personaInfo.name` (fallback `"You"`), matching `ChatMessage.tsx`.
- `{{char}}` → `seg.speaker` when present, otherwise the first character in `characterMap`. Left literal if neither is available.

### DIFF
--- a/packages/client/src/components/game/GameNarration.tsx
+++ b/packages/client/src/components/game/GameNarration.tsx
@@ -450,6 +450,23 @@ export function GameNarration({
       }
     }
 
+    // Resolve display macros ({{user}}, {{char}}) on every segment's content so
+    // downstream renderers (formatNarration / animateTextHtml) receive the
+    // already-substituted text. Mirrors ChatMessage.tsx display-time logic.
+    const userName = personaInfo?.name || "You";
+    const fallbackCharName = (() => {
+      if (!characterMap || characterMap.size === 0) return null;
+      const first = characterMap.values().next().value;
+      return first?.name ?? null;
+    })();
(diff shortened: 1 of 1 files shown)


SpicyMarinara:
Thank you, was planning to address it myself, but you were faster!

PR #162, opened 2026-04-21
Fix comfyui generation for sprites and fix default workflow

felorhik:
## Why this change

When generating sprites, the comfyui workflow from the connection settings is not passed through. So it will always default to the default comfyui workflow. This is fine for most SDXL setups, but if using models like flux or ernie, they require different text encoders / vae.

Also the default comfyui workflow does not use the generated seed, so if using the same connection, the output would always be the same, regardless of regeneration. So I have updated the workflow to use the generated seed during the generation request. Width / Height were also not used, so I fixed that as well.

## What changed

Update sprite generation to pass comfyui workflow.
Update pose generation to pass comfyui workflow.
Update default comfyui workflow to add in placeholders for seed/width/height.
Added an env variable to allow customisation of the comfyui timeout (120 seconds is not always sufficient for some models)

## Validation

Setup a comfyui connection, and use a custom workflow and attempt to do sprite / pose generation.
Before update, will fail.

Retry a generation without any custom workflow
Before update, the results will be the same on each generation.

### Manual verification notes

I had manually checked with SDXL / Chroma / Flux / Ernie based workflows. With / Without custom workflows added.

First PR here, and a simple one so have not done all the linting / tooling.

### DIFF
--- a/packages/server/src/routes/sprites.routes.ts
+++ b/packages/server/src/routes/sprites.routes.ts
@@ -428,6 +428,7 @@ export async function spritesRoutes(app: FastifyInstance) {
               height: targetHeight,
               referenceImage: resolvedRefs[0],
               referenceImages: resolvedRefs.length > 1 ? resolvedRefs : undefined,
+              comfyWorkflow: conn.comfyuiWorkflow || undefined,
             });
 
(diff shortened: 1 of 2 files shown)


SpicyMarinara:
Amazing job, thank you!

PR #170, opened 2026-04-21
Fix/sidecar separation and bugs

munimunigamer:
Fixes sidecar startup state and enables logs for ollama to see whats going on
Also separates sidecar from normal install fully


SpicyMarinara:
Amazing work, thank you!

PR #177, opened 2026-04-22
Fix comfyui game asset generation

felorhik:
## Why this change

Fixes background and portrait generation for comfyui connections.

## What changed

Portrait and Background game generation

Note: This still needs some testing. It seems to work, but its really hard to trigger tests of game portrait and background generation reliably. 

Current issue I am having is that when a generation does happen, with a custom workflow not including `%model%` it gives an error saying it tried to use that as the model itself (since none are selected in the connection settings). The generation does work however, and generated using the correct workflow. So not sure what is triggering it yet.

The issue is likely separate to getting the game asset generation from utilising the comfyui workflow. So the fixes here may be able to be merged, though more testers for it would be good since I am having difficulty triggering the asset generation reliably.


felorhik:
Tested further, may of had lingering dist files in my failed test / adjusted my connection settings / sidecar. Seems to generate fine without errors now.

felorhik:
Added additional adjustment for escaping prompts to ensure json remains valid.

SpicyMarinara:
Thank you so much!

PR #178, opened 2026-04-22
Append "Respond ONLY as [name]" on group chat individual mode regens

mm14141:
## Why this change

In group chats set to individual mode, the engine appends a "Respond ONLY as [name]" instruction on the initial generation of a message, but does not do so when regenerating, often causing the AI to play as only the first character in the group or multiple characters at once, inconsistent with the first generation. 

## What changed

- In `generate.routes.ts`, before running the generation, if in a group chat, in individual mode, and regenerating, it appends the "Respond ONLY as [name]" instruction as it does for the individual loop. 

## Validation

- [x] `pnpm check`

### Manual verification notes

- Verified group chat, individual mode regenerations have the instruction appended.
- Verified that generations and regenerations weren't affected in conversations, 1-on-1 roleplays or in merged/narrator mode roleplays.

## Docs and release impact

- [x] No docs changes needed

N/A


mm14141:
Addressed the review. I have it using the regenerated message from chatMessages so it doesn't do a new DB read, and it replies with 404 if the regenerated message isn't found, or with 400 if the message has the wrong chatId or is missing its characterId.

SpicyMarinara:
Thank you for the PR, adding you to the contributors!

PR #190, opened 2026-04-22
Fix Android intro typewriter blocking touch-scroll in Game Mode

Romuromylus:
## Summary

- In Game Mode, the intro world-overview screen was unscrollable on Android after a new game was started. Desktop mouse-wheel worked fine.
- Root cause: `IntroTypewriter` in [GameSurface.tsx](packages/client/src/components/game/GameSurface.tsx) wrapped its text in a `<div className="overflow-y-auto">` and auto-scrolled that inner div. The inner div had no height constraint so it never actually scrolled — but on Android Chrome the nested overflow container still captured touch events and blocked scroll-chaining up to the parent (`min-h-0 flex-1 overflow-y-auto`). Desktop mouse-wheel chains through nested overflow naturally, which is why Windows was unaffected.
- Fix: drop the inner `overflow-y-auto` and track the reveal edge with a sentinel `<span>` + `scrollIntoView({ block: "end" })`. The outer scroll container now handles touch input unimpeded, and the reveal-follow behavior is preserved.

## Test plan

- [x] TypeScript check passes (`pnpm --filter @marinara-engine/client exec tsc --noEmit`).
- [x] Tested on Android in a live deployment: scroll works during and after typewriter reveal.
- [x] Reveal edge still scrolls into view as characters appear.
- [x] Desktop (Chrome on Windows) still scrolls fine with mouse wheel.

### DIFF
--- a/packages/client/src/components/game/GameSurface.tsx
+++ b/packages/client/src/components/game/GameSurface.tsx
@@ -81,7 +81,7 @@ import type { PersonaInfo } from "../chat/chat-area.types";
 /** Typewriter component for the intro screen — reveals text character-by-character. */
 function IntroTypewriter({ text, onComplete }: { text: string; onComplete?: () => void }) {
   const [visible, setVisible] = useState(0);
-  const containerRef = useRef<HTMLDivElement>(null);
+  const endRef = useRef<HTMLSpanElement>(null);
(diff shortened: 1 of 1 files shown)


SpicyMarinara:
Thank you for the PR, adding you to the contributors!

PR #188, opened 2026-04-22
universal agent tool support

adunato:
## Why this change

  Problem Statement:
  Currently, the generation pipeline in the Marinara Engine contains a hardcoded restriction that only allows the
  spotify agent to utilize function calling (tools). For all other agents—built-in or custom—the server ignores their
  enabledTools configuration and sends them into the pipeline as "batchable" (text-only) agents. This prevents users
  from extending the engine's capabilities via custom agent-tool interactions.

  Solution:
  This PR generalizes the tool resolution logic within the generate route. It relocates tool definition and Spotify
  credential resolution to the beginning of the pipeline and iterates through all agents. For every agent with tools
  enabled in its settings, the server now correctly attaches a toolContext, enabling true function-calling capabilities
  for both custom and built-in agents.

## What changed

  Key Changes
   - generate.routes.ts:
       - Relocated tool resolution and custom tool fetching to the start of the generation flow.
       - Decoupled tool assignment from the hardcoded spotify type check.
       - Implemented a universal loop that builds and attaches toolContext to any resolvedAgent with active tools.
       - Optimized Spotify token refresh to occur early in the request lifecycle.
   - agent-executor.ts:
       - Added real-time terminal logging for tool calls and results using the [agent-tools] tag, following existing
         project conventions.

## Validation

- [x] `pnpm check`
- [x] Manual verification completed (describe below)

### Manual verification notes

  Verification Performed
   - Prerequisites: Test Agent enabled with instruction to invoke roll_dice tool (tool enabled in test agent configuration), tools enabled in chat.
   - Orchestration Test: Confirmed via server logs that a custom agent now reports as 1 tool-using instead of 1
     batchable.
   - Functionality Test: Verified that a custom agent can successfully call the roll_dice tool, receive the JSON result,
     and incorporate it into its final response.
   - Logging Test: Verified that the new [agent-tools] logs correctly display the function name, arguments, and raw
     results in the terminal.
   - Cross-check: compared before / after change server logging with the same test to demonstrate tool calling behaviour (before:[agent-pipeline] executeGroup: 1 batchable, 0 tool-using [ 'custom-test-agent' ]. after: [agent-pipeline] executeGroup: 0 batchable, 1 tool-using []
   - OUT of scope of validation: existing tools. The only agent that seem to be enabled for tools and has tools in its prompt is "Spotify DJ", the others either don't have tools enabled and/or do not mention tools in their prompt.
[...]

PR #212, opened 2026-04-24
Universal agent tool support

### LINKED ISSUE #279: [Feature]:
NOTE - placeholder issue to support PR #212 

# Issue: Agents need configurable access to tools at runtime

## Summary

Agents should be able to use tools according to the tool selections configured in the UI. This should work consistently for built-in agents and custom agents, while still respecting chat-level tool settings.

## Problem

The app exposes tool/function-calling configuration in the UI, but agent runtime tool access is not consistently driven by that configuration. In particular, custom agents cannot reliably use tools through the same path as built-in/special-cased agents.

This limits agent extensibility:

- Users can create custom agents, but cannot reliably grant them selected tools at runtime.
- Tool behavior is harder to reason about because some paths are special-cased instead of using one consistent resolver.
- Chat-level tool settings and per-agent tool settings are not clearly enforced together.

## Desired Behavior

- Chat settings control whether tools are available in a chat.
- Chat settings control which tools are available in that chat.
- Agent settings control which of those available tools a specific agent may use.
- Built-in and custom agents use the same runtime tool resolution path.
- Tool calls from agents are checked against the agent's allowed tools before execution.
- Invalid or conflicting custom tool definitions should not break generation for otherwise valid tools.

## Acceptance Criteria

- A custom agent can be configured in the UI with selected tools and can use those tools during generation.
- A built-in agent and a custom agent follow the same runtime tool-selection rules.
- If chat-level tools are disabled, agents cannot use tools.
- If a tool is not selected for an agent, that agent cannot execute it.
- If a custom tool is invalid or conflicts with another tool name, it is skipped cleanly and does not prevent other valid tools from being available.

adunato:
Closes #279

Resubmission of universal agent tool support #188 after clean up and addressing previous copilot feedback

# CR002 Universal Agent Tool Support - Upstream PR Description

### Problem statement

The generation pipeline previously only attached function-calling support to the built-in Spotify agent. Other built-in and custom agents could have `enabledTools` configured, but they were still treated as text-only agents in the pipeline and executed as `batchable`.

This meant non-Spotify agents could not actually invoke tools during their turn, even when tool usage was configured for that agent.

### Solution

This change generalizes agent tool support in `generate.routes.ts`.

When chat-level tools are enabled, the route now resolves tool definitions once, filters them per agent allowlist, and attaches `toolContext` to any resolved agent with active tools. This preserves chat-level tool gating while enabling both built-in and custom agents to use tools correctly.

### Core changes

- `generate.routes.ts`
  - moved tool definition and custom-tool resolution earlier in the request flow
  - removed the hardcoded Spotify-only tool assignment path
  - added universal per-agent `toolContext` attachment for resolved agents with allowed tools
  - kept chat-level tool enablement as the master gate for all tool use in the request
  - reused a single Spotify credential resolution path and only resolves/refershes Spotify credentials when Spotify tools are actually relevant to the request
  - added server-side filtering for main-generation tool calls so only allowed tools are executed
  - added server-side allowlist enforcement for agent tool calls
  - hardened custom tool registration by rejecting duplicate tool names and skipping invalid parameter schemas safely

- `agent-pipeline.ts`
  - improved diagnostics so pipeline logs distinguish `batchable` agents from `tool-using` agents

- `agent-executor.ts`
  - added `[agent-tools]` diagnostics for tool invocation and completion
  - gated raw args/results behind `DEBUG_AGENTS=true`
  - redacts and truncates debug payload logging

## Validation

- [x] `pnpm check`
- [x] Manual verification completed

### Manual verification notes

- Custom-agent tool path:
  - configured a custom test agent with `roll_dice` enabled
  - confirmed the pipeline now reports the agent as `tool-using` instead of `batchable`
  - confirmed the agent successfully invoked `roll_dice` and received the JSON result for the follow-up model turn

- Spotify regression check:
  - confirmed the Spotify DJ agent still executes through the tool loop
  - confirmed successful invocation of `spotify_get_playlist_tracks`, `spotify_play`, and `spotify_set_volume`
  - confirmed the earlier regression where Spotify tool execution stopped before playback is no longer present

- Logging verification:
  - confirmed normal logs show tool invocation/completion with the `[agent-tools]` tag
  - confirmed debug mode (`DEBUG_AGENTS=true`) includes redacted/truncated tool args and results

### Before / after evidence

Before:

```text
[agent-pipeline] executeGroup: 1 batchable, 0 tool-using [ 'custom-test-agent' ]
```

After:

```text
[agent-pipeline] executeGroup: 0 batchable, 1 tool-using { batch: [], tools: [ 'custom-test-agent' ] }
[agent-tools] custom-test-agent calling: roll_dice
[agent-tools] custom-test-agent roll_dice completed
```

## Docs and release impact

- [x] No user-facing docs changes needed

## UI evidence

Not applicable.

## Additional notes

- This PR intentionally keeps chat-level tool enablement as the master switch.
- Agent `enabledTools` determines the subset of tools an agent may use once tools are enabled for the chat.

## Summary by CodeRabbit

* **Refactor**
  * Unified tool resolution and execution for agents and main generation, with earlier Spotify credential refresh and per-agent tool permission enforcement
* **Bug Fixes**
  * Skips invalid custom tools and prevents duplicate tool names across sources
* **Diagnostics**
  * Safer tool-call logging with sensitive-data redaction and improved debug information for agent grouping and tool calls


LukaTheHero:
Thanks for resubmitting this — the cleanup compared to #188 is real. The single resolution path for tool defs, the unified denial behavior in main-gen + agent paths, and the Spotify-only-when-needed refresh are all clear wins. A few things stood out on a careful read:

## Things that look good

- **Master gate still holds.** `enableTools = inputBody.enableTools === true || chatMeta.enableTools === true` is the only entry point into the resolver block in `packages/server/src/routes/generate.routes.ts`. Nothing downstream creates `toolContext` outside that gate. Good.
- **Symmetric allowlist enforcement.** Both the per-agent `executeToolCall` closure and the main-generation path (`permittedToolCalls` / `deniedToolResults`) reject unknown tool names and return a structured `{ error, allowed }` payload with `success: false`. The denied results are merged back in id-order so the model sees a tool message for every call it made — that's the right shape for OpenAI tool loops.
- **Redaction.** `redactSensitiveValue` + `formatToolPayloadForLog` in `agent-executor.ts` cover the obvious vectors (token/secret/password/api[_-]?key/authorization/cookie/credential, plus `Bearer …` regex), and payloads are truncated at 400 chars. Logs at INFO level only emit `[agent-tools] <type> calling: <name>` / `completed`, with raw args/results gated behind `isDebugAgentsEnabled()`.

## Risks worth confirming

1. **Existing chats with `chat.enableTools=true` and no per-agent config will silently lose batching.** In `generate.routes.ts` the per-agent resolver falls back to `DEFAULT_AGENT_TOOLS[agent.type]` when `agentSettings.enabledTools` is undefined. Looking at `packages/shared/src/types/agent.ts:491`, that map populates non-empty defaults for `world-state`, `continuity`, `expression`, `director`, `quest`, `lorebook-keeper`, `combat`, `character-tracker`, `persona-stats`, plus spotify. Once those defaults flow through, `executeGroup` in `agent-pipeline.ts:89-90` will route each of them as `tool-using` and run them individually instead of batching. For users who flipped on chat-level tools but never edited their agents, this turns one batched call into N sequential ones — a real cost/latency hit. Consider gating per-agent `toolContext` on an explicit `enabledTools` (i.e. only when `agentSettings.enabledTools` is set), or only for agents whose default list overlaps the chat's `activeToolIds`.

2. **Duplicate-tool rejection throws inside the request handler.** The built-in pre-pass throws `Error("Duplicate tool name…")` synchronously inside the route. If a user happens to name a custom tool with a built-in name (the most likely collision in practice), the entire generation request 500s instead of just skipping the offending custom tool. The custom-tool branch already has a graceful skip via `console.warn`. I'd lean toward making both cases warn-and-skip, or at minimum caught and surfaced as a structured SSE `error` event so the chat doesn't just die. (Side note: this file should be using the Pino logger per `CLAUDE.md § Logging`, not `console.warn`.)

3. **`DEBUG_AGENTS=true` is undocumented.** `isDebugAgentsEnabled()` is the only consumer in `runtime-config.ts:126`, and the var doesn't show up in `docs/CONFIGURATION.md` or the README. If it's intended for end-users, please add it to the Logging Levels reference; if it's strictly internal, a one-line comment near the export saying so would help future maintainers.

4. **`needsSpotify` short-circuit is correct but fragile.** `chatAllowsSpotify` is computed by intersecting `resolvedToolNames` with the `spotify` slot in `DEFAULT_AGENT_TOOLS`. If anyone ever adds a new built-in Spotify tool without updating that map, the early-refresh branch will silently no-op for it. Worth a comment at the `DEFAULT_AGENT_TOOLS["spotify"]` declaration noting it's a server-side allowlist as well as a UI default.

## Smaller suggestions

- `enabledCustomTools.filter((ct: any) => …)`, `customFiltered.forEach((ct: any) …)`, `e: any` in `searchLorebookForTools` — these `any`s undo the new `ToolExecutionContext` typing work in `tool-executor.ts`. Worth a follow-up to type `customToolsStore.listEnabled()` and the lorebook entry shape properly.
- The `resolvedAgents.find(... "spotify") ?? enabledConfigs.find(...) ?? (await agentsStore.list()).find(...)` chain is doing a DB read inside a hot path. The fallback only fires when no enabled spotify config exists, so it's bounded, but a short comment explaining why we ever look beyond `resolvedAgents` would help — the intent isn't obvious.
- The `agentsStore.update(...).catch(() => {})` swallow on token refresh is a known pattern, but with the new redacted logger available you could `.catch((err) => logger.warn(err, "[spotify] failed to persist refreshed tokens"))` cheaply.

## Manual tests I'd like to see in the PR

Since there's no automated coverage for this path, could you confirm:

- A chat with `enableTools=true`, `activeToolIds=[]` (i.e. all tools), and only built-in agents — does the agent pipeline log show batched groups, or does every agent now run individually? This is the regression I'm most worried about.
- A custom tool whose `name` collides with a built-in (e.g. `roll_dice`) — does the request 500, or does it skip cleanly?
- Two custom tools where one has an invalid `parametersSchema` and one is valid — does only the invalid one get skipped, and does the valid one still execute?
- Main-generation tool-call denial — model invokes a tool not in the chat's `activeToolIds` allowlist; verify the synthetic error result is fed back and the model recovers gracefully rather than looping.

Happy to re-review once those land or once we've decided on the per-agent default-tool gating question.

---

adunato:
Hey, thanks for the review. I've pushed a follow-up addressing the main points.

## Changes made

- **Per-agent default-tool gating**
Good catch. I updated this so the runtime behavior now follows the persisted per-agent tool configuration instead of applying built-in defaults server-side.

`DEFAULT_AGENT_TOOLS` is still used to initialize the UI selection for built-in agents. When an agent is added or saved, that selection is persisted as `agent.settings.enabledTools`.

The server-side resolver now uses that persisted configuration only:

- `chat.enableTools` remains the master gate for tool use in the chat.
- `chat.activeToolIds` determines which tools are available in that chat.
- `agent.settings.enabledTools` determines which of those available tools the agent may use.
- If `settings.enabledTools` is absent, the server does not fall back to `DEFAULT_AGENT_TOOLS` and does not attach `toolContext`.

That prevents existing chats with chat-level tools enabled but no persisted per-agent tool config from implicitly moving built-in agents into the tool-using path.

I also confirmed the expected behavior manually with chat-level tools enabled and an agent with no per-agent tools enabled:

```text
[agent-pipeline] executeGroup: 1 batchable, 0 tool-using {"batch":["prose-guardian"],"tools":[]}
```

- **Custom tool collision handling**
  - Custom tools whose names collide with an already-registered built-in/custom tool are now skipped with a warning instead of aborting generation.
  - Built-in duplicate detection remains a hard invariant.

- **Logging alignment**
  - Agent tool argument/result debug logs now use the existing Pino debug standard via `LOG_LEVEL=debug`.
  - Removed CR002's use of the undocumented `DEBUG_AGENTS` gate.
  - Replaced the remaining custom-tool `console.warn` with `logger.warn`.

- **Spotify cleanup**
  - Agent-side Spotify refresh detection now only considers explicit per-agent `enabledTools`.
  - Added comments explaining the Spotify tool allowlist and credential fallback lookup.
  - Token refresh persistence failures now log a warning instead of being silently swallowed.

## Validation

- Confirmed batching behavior with chat-level tools enabled and no per-agent tools enabled:

```text
[agent-pipeline] executeGroup: 1 batchable, 0 tool-using {"batch":["prose-guardian"],"tools":[]}
```

Agents only move to the `tool-using` path when they have explicit per-agent tools enabled.

- Confirmed duplicate custom tool collision handling by creating a custom tool named `roll_dice`:

```text
[tools] Skipping custom tool "roll_dice" because it collides with existing built-in tool
```

Generation continues and the built-in tool remains registered.

- I could not manually reproduce an invalid `parametersSchema` through the current UI. The editor builds schemas from constrained controls, and Static Result is just returned as tool output rather than treated as an output schema. This guard is therefore defensive coverage for legacy/import/API-corrupted rows; it now skips only the offending invalid tool instead of failing the whole resolver.

- I also do not think the denied main-generation tool-call path is a reliable manual UI test, because it depends on the model choosing to call a disabled tool. The deterministic server path is in place: denied tool calls are converted into synthetic error tool results, merged by `toolCallId`, and fed back so the model receives a tool message for every tool call it made. This is better suited to an automated test with a mocked provider.

Server validation also passed:

```text
pnpm --filter @marinara-engine/server lint
server test suite
```

SpicyMarinara:
Thank you for your hard work!

PR #213, opened 2026-04-24
fix(client): Author's Notes edits lost on outside-click

Trade-Mottoes:
## The bug

Open Author's Notes, type content, click outside the popover. The typed content is lost — sometimes. Whether the save takes is, in the current implementation, in the lap of the gods. :)

**Repro:** Open Author's Notes on any chat. Type a paragraph. Click anywhere outside the popover. Re-open. Roughly half the time the content is gone.

## Root cause

The outside-click handler fires on `mousedown` and unmounts the textarea before its `onBlur` (which triggered the save) can run. No reliable fix is possible within the popover-plus-blur-save architecture — the container is trying to be both a transient peek surface (dismiss on mousedown) and a live editor (persist on blur), and those contracts race by construction.

## Fix

Split viewing and editing into their proper containers.

- **`AuthorNotesPeek`** — read-only popover. Transient dismissal is safe because no mutation is in flight. Uses `click` rather than `mousedown` for predictable close semantics.
- **`AuthorNotesPanel`** — committed editing in the shared `Modal` primitive (same one used by `EditAgentModal`, `LorebookMakerModal`, etc.) with explicit Save / Cancel. `isDirty` gates prop-sync so external refetches don't clobber pending edits.
- **`AuthorNotesButton`** — three-state machine (closed / peek / edit) owning all transitions; the two child components are unaware of each other.

## Why peek/edit rather than just "use Modal for everything"

Viewing and editing have different optimal containers. Peeking is by far the more common action ("what did I put in here again?") and is cheap, frequent, and safe. Editing wants a committed feel — backdrop, focus trap, Escape. The one-container-fits-both compromise is precisely what caused the save race: the popover was trying to serve both contracts at once.

## User-visible result

All four close paths (X / backdrop / Escape / Cancel) are race-proof by construction. Desktop and mobile now render as a single consistent dialog (previously desktop was a popover, mobile a full-screen portal). Also picks up `role=dialog`, `aria-modal`, Escape handling and focus trap from the shared `Modal`.

## Testing

- `pnpm lint && pnpm build` pass.
- Manually verified all four close paths on desktop and mobile viewport.
- Manually verified the save race is gone: rapid type-then-outside-click no longer loses content.

## Screenshots
Peek with discrete edit button:
image

Edit:
image

## Release Notes

* **New Features**
  * Author's Notes now use explicit Save and Cancel buttons for draft editing instead of auto-save.
  * Preview mode for Author's Notes is now read-only with an Edit button to transition to edit mode.
  * Enhanced mobile UI for Author's Notes with improved preview and edit workflow.

### DIFF
--- a/packages/client/src/components/chat/ChatRoleplayPanels.tsx
+++ b/packages/client/src/components/chat/ChatRoleplayPanels.tsx
@@ -1,5 +1,5 @@
 import { useEffect, useState } from "react";
-import { Globe, Loader2, PenLine, X } from "lucide-react";
+import { Globe, Loader2, PenLine, Save, X } from "lucide-react";
 import { useUpdateChatMetadata } from "../../hooks/use-chats";
 import { useActiveLorebookEntries } from "../../hooks/use-lorebooks";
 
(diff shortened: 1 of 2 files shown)


Trade-Mottoes:
Both flagged items are out of scope for this PR:

The docstring-coverage check is mis-targeted for a TypeScript codebase.
isMobile on line 454 is pre-existing upstream code (blame: 566aaba0, Apr 3) and matches the repo's pattern across ~10 files. Introducing a useIsMobile hook would be a useful separate refactor but inconsistent if applied here alone.

LukaTheHero:
looking into it.

LukaTheHero:
Hey @Trade-Mottoes — first off, thank you for catching this and writing up such a clear diagnosis. The mousedown-unmounts-before-blur analysis is exactly right, and you saved me a lot of time figuring out why edits were vanishing.

I went a different direction in #239 and want to be upfront about why, so this isn't just a silent close.

We prefer to keep CSS/HTML styling consistent with the existing infrastructure — the popover-with-implicit-save is the established pattern across the chat surface, and swapping in the shared Modal primitive (with explicit Save/Cancel, backdrop, focus trap) changes the UX contract for this control even though it's a clean win architecturally elsewhere. I'd rather fix the race in place than reshape the surface.

The other thing is the no-explicit-save behavior. Author's Notes are a quick, mid-flow edit ("tweak the tone, keep going") — adding a Save button means one extra click on every edit, forever, to fix a bug that only bites when the save races the unmount. So #239 keeps the implicit save and just closes the race: a useEffect cleanup flushes the latest values on unmount (cleanups run synchronously, React Query's mutate is fire-and-forget, so the save survives the textarea being destroyed). A baseline snapshot gates it so peek-and-close doesn't trigger spurious writes. Zero markup or styling changes — one file, +29/-1.

Going to close this in favor of #239, but seriously — thanks for the bug report and the writeup. Please keep them coming.

PR #222, opened 2026-04-24
feat(chat): add "Mari is thinking…" indicator for command execution

cha1latte:
## Summary

Adds a small "Mari is thinking…" pill with pulsing dots above the composer whenever Professor Mari is executing one of her embedded commands (create/update character, fetch, create chat, navigate). Shows only in chats where Mari is a participant.

## Why

From [the Discord feature-request thread](https://discord.com/) — users can't tell whether Mari is still working on a task or stalled, because her commands run *after* her reply finishes streaming and there's no visible signal during that window. Quote: *"I don't know if asking the assistant to work on my cards is happening, or if it's frozen, or what."*

Toasts fire only after each command completes, so they're confirmation after-the-fact, not reassurance during-the-wait.

## How it works

| Layer | Change |
|---|---|
| Server (`generate.routes.ts`) | Emits new SSE events `assistant_commands_start` (before the command loop) and `assistant_commands_end` (after), with no filtering — any conversation-mode character's commands emit them. |
| Client store (`chat.store.ts`) | New `commandsExecutingChatId: string \| null` mirroring the server-side window, set by SSE handlers in `use-generate.ts`. |
| Client hook (`use-generate.ts`) | Handles the two new events; also clears on `done` / `error` as a safety net. |
| Component (`MariThinkingIndicator.tsx`) | Self-computes `isMariChat` from the active chat's `characterIds` (so the pill is Mari-scoped on the client, keeping the server event neutral). Mounted in both `ChatInput` (roleplay) and `ConversationInput` (conversation). |

Two notable implementation details:

- **Direct Zustand subscribe, not a selector hook.** Simple commands (single DB write) finish faster than a React tick — the start/end state updates would land in the same render batch and the component would never observe the active state. Subscribing to the store directly catches both transitions regardless of batching.
- **600ms minimum display duration.** Without it, fast commands would mount+unmount the pill below perceptual threshold. This way even a 1ms command gives a visible "blip."

## Known limitations

- The event is **not** Mari-specific on the server. It emits for any conversation-mode character that runs commands (e.g. `schedule_update`, `cross_post`). The Mari gate lives on the client, so those cases don't show a pill — but if a future feature wants a generic "assistant is working" indicator for non-Mari commands, the plumbing is already there.
- The pill is not customizable yet. Keeping this PR focused on the core signal; a user-swappable visual (e.g. the Mari-themed GIF referenced in the Discord thread) can layer on top later with no changes to the event plumbing.
- If the stream aborts mid-command-loop, the safety clears on `done`/`error` handle cleanup, but the `assistant_commands_end` event won't be received. Behavior: pill disappears when the stream errors, which is correct but means the user doesn't see a natural "finished" signal if the server crashes mid-loop. Acceptable.

## Test plan

Manually verified in dev against a Mari conversation chat:

- [x] Mari chat + action request (create/update/fetch character) → pill appears after stream, disappears after toast
- [x] Mari chat + pure conversation (no commands) → no pill
- [x] Non-Mari chat → no pill ever, including when that character's own commands run
- [x] Multiple commands in one reply → pill stays for the whole window
- [x] Abort mid-stream → no pill
- [x] Page reload while pill is showing → no stuck state after reload
- [x] Chat switch during command execution → pill disappears with the chat context
- [x] `pnpm check` clean (lint + production build, no new warnings)
- [x] `pnpm version:check` passes
- [x] `pnpm guard:installer-artifacts` passes

Screenshot of the pill mid-execution attached.

image

## Summary by CodeRabbit

* **New Features**
  * Added a "Mari is thinking…" indicator above the chat/composer input that appears while embedded background commands run (character/chat actions and navigation). It provides ARIA-live feedback, enforces a minimum visible duration to avoid flicker, hides immediately when switching chats, and reliably shows/hides during rapid or batched command sequences so the UI no longer appears frozen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 This file is the release-notes source of truth for Marinara Engine. Reuse these entries when publishing GitHub Releases for tags in the `vX.Y.Z` format.
 
+## [Unreleased]
+
+### Added
+
+- "Mari is thinking…" indicator appears above the composer while Professor Mari executes her embedded commands (create/update character, fetch, create chat, navigate). Makes it clear her background work is running and not frozen.
+
 ## [1.5.5]
 
 ### Added
--- a/packages/client/src/components/chat/ChatInput.tsx
+++ b/packages/client/src/components/chat/ChatInput.tsx
@@ -25,6 +25,7 @@ import { EmojiPicker } from "../ui/EmojiPicker";
 import { QuickConnectionSwitcher } from "./QuickConnectionSwitcher";
 import { QuickPersonaSwitcher } from "./QuickPersonaSwitcher";
 import { QuickSwitcherMobile } from "./QuickSwitcherMobile";
+import { MariThinkingIndicator } from "./MariThinkingIndicator";
 
 interface Attachment {
   type: string; // MIME type
@@ -614,6 +615,9 @@ export const ChatInput = memo(function ChatInput({
           ))}
         </div>
       )}
+
+      {/* Mari command-execution indicator */}
+      <MariThinkingIndicator />
 
       {/* Main input container */}
       <div
--- a/packages/client/src/components/chat/ConversationInput.tsx
+++ b/packages/client/src/components/chat/ConversationInput.tsx
@@ -24,6 +24,7 @@ import { QuickPersonaSwitcher } from "./QuickPersonaSwitcher";
 import { QuickSwitcherMobile } from "./QuickSwitcherMobile";
 import { EmojiPicker } from "../ui/EmojiPicker";
 import { GifPicker } from "../ui/GifPicker";
+import { MariThinkingIndicator } from "./MariThinkingIndicator";
 
 interface Attachment {
   type: string;
@@ -625,6 +626,9 @@ export function ConversationInput({ characterNames = [] }: ConversationInputProp
           ))}
         </div>
       )}
+
+      {/* Mari command-execution indicator */}
+      <MariThinkingIndicator />
 
       {/* Input bar */}
       <div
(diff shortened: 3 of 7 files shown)


LukaTheHero:
Hey @cha1latte — first of all, thank you for this. The diagnosis was sharp, the wiring (SSE start/end events, per-chat Set, min-visible duration, isMariChat gate, chat-switch handling) was all the right shape, and reviewing it gave me a clean foundation to work from.

I'm closing this in favor of #240, which preserves your four commits with original authorship and adds my changes as a single commit on top. Wanted to be transparent about what I changed and why:

**Bug — indicator never appeared in practice.** The component used `useChatStore.subscribe(selector, evaluate)`. Turns out this codebase has documented timing issues with `subscribeWithSelector` + React 19 batching — `GameSurface.tsx:2079-2081` has a comment about exactly this, with the workaround already in use there. SSE events arrived correctly (verified via Network → EventStream) but the subscription dropped them in flight, so the pill never rendered. Switched the transport to window `CustomEvent`s (`marinara:mari-phase`), matching the established pattern (`marinara:generation-complete`, `marinara:generation-error`) already used elsewhere in `use-generate.ts`.

**UX extension — covers the streaming window too.** While I was in there I extended the pill to cover the token-streaming phase, not just the post-stream command window. Two phases now:
- "Mari is thinking…" — fires on the first token (same edge that clears the typing indicator, so they never overlap)
- "Mari is updating your stuff…" — replaces it when commands start

Same transport, same min-visible-duration logic, same per-chat ownership tracking — just a wider window of "Mari is doing work for you."

**Cleanup** — fixed the `setCommandsExecuting` missing-dep lint warning, extracted `isMariParticipant` as a pure function, split the chat-switch hide into its own focused effect.

Your store changes (the `commandsExecutingChatIds` Set and `setCommandsExecuting` action), your SSE event names, your safety-net pattern, your min-duration approach, the dual-mount layout — all preserved. The fix is really just "swap the transport and widen the trigger."

Closing this in favor of #240.

PR #230, opened 2026-04-25
Feature: Add scene fork roleplay actions

OnlyJimmy:
## Why this change

Active scenes could be concluded or abandoned, but there was no way to preserve a scene as an independent roleplay thread. This adds a focused fork flow so users can keep narrative continuity without carrying over scene lifecycle behavior.

## What changed

- Added `POST /api/scene/fork` with `clone` and `convert` modes.
- Added shared scene fork request/response types.
- Added `Clone from here` to active scene message actions, copying scene continuity and messages up to the selected message into a standalone roleplay chat.
- Added `Convert` beside the scene discard controls, with confirmation, to detach the active scene into a standalone roleplay.
- Added hidden narrator continuity context so forked chats preserve useful pre-scene context without showing scene setup in the transcript.
- Preserved copied message metadata, timestamps, active swipe state, and all message swipes during fork creation.
- Preserved normal roleplay metadata such as characters, persona, group, preset, connection, lorebook metadata, settings, and background where applicable.
- Ensured forked roleplay chats strip scene lifecycle metadata and do not write scene summaries or character memory during convert cleanup.
- Added fork in-flight guards and disabled clone/convert actions while scene generation is streaming.
- Added best-effort cleanup for partially-created fork chats if message copying or convert cleanup fails.
- Updated forked chat names to distinguish cloned and converted chats.

## Validation

- [x] `pnpm check`

### Manual verification notes

- Manually tested functionality with Prof. Mari
- Scenes remain disconnected from their clones
- Context is available to the AI, but hidden from the user
- Conversations are not impacted by cloned or converted roleplays
- Lorebooks travel with cloned and converted scenes
- Swipes and Attachments now transfer to cloned and converted roleplays
- Convert and Clone disable during streaming to prevent errors
- Cloning or Converting from the very first message of a scene functions as expected

## Docs and release impact

- [x] No docs changes needed

## UI evidence

Clone Button
shot1

Clone Result
shot2

Convert Button
shot3

Confirmation Prompt
shot4

## Summary by CodeRabbit

* **New Features**
  * "Clone from here" to create a new scene starting at a selected message (supports cloning up to a point and including scene context).
  * "Convert" to turn an active scene into a standalone roleplay chat.
  * Messages marked hidden are omitted from the visible chat transcript.

* **UX Improvements**
  * Action buttons show contextually in scene UIs, disable during fork operations, and surface success/error feedback.

### DIFF
--- a/packages/client/src/components/chat/ChatArea.tsx
+++ b/packages/client/src/components/chat/ChatArea.tsx
@@ -725,10 +725,10 @@ export function ChatArea() {
 
   const handleCloneSceneFromHere = useCallback(
     (messageId: string) => {
-      if (!activeChatId || isForking) return;
+      if (!activeChatId || isForking || isStreaming) return;
       forkScene(activeChatId, "clone", { upToMessageId: messageId });
     },
-    [activeChatId, forkScene, isForking],
+    [activeChatId, forkScene, isForking, isStreaming],
   );
 
   // Peek prompt state
@@ -1361,7 +1361,8 @@ export function ChatArea() {
           onToggleConversationStart={handleToggleConversationStart}
           onPeekPrompt={handlePeekPrompt}
           onBranch={isSceneChat ? undefined : handleBranch}
-          onCloneSceneFromHere={isSceneChat && !isForking ? handleCloneSceneFromHere : undefined}
+          onCloneSceneFromHere={isSceneChat ? handleCloneSceneFromHere : undefined}
+          isCloneSceneFromHereDisabled={isForking || isStreaming}
           onToggleSelectMessage={handleToggleSelectMessage}
           onSummaryContextSizeChange={handleSummaryContextSizeChange}
           onRerunTrackers={handleRerunTrackers}
@@ -1369,7 +1370,7 @@ export function ChatArea() {
           onConcludeScene={() => concludeScene(activeChatId)}
           onAbandonScene={() => abandonScene(activeChatId)}
           onForkScene={forkScene}
-          isForkingScene={isForking}
+          isForkingScene={isForking || isStreaming}
           onOpenSettings={() => setSettingsOpen(true)}
           onOpenFiles={() => setFilesOpen(true)}
           onOpenGallery={() => setGalleryOpen(true)}
--- a/packages/client/src/components/chat/ChatArea.tsx
+++ b/packages/client/src/components/chat/ChatArea.tsx
@@ -290,7 +290,7 @@ export function ChatArea() {
   const chatMode = rawMode ?? lastModeRef.current;
   const isRoleplay = chatMode === "roleplay" || chatMode === "visual_novel";
   const { startEncounter } = useEncounter();
-  const { concludeScene, abandonScene } = useScene();
+  const { concludeScene, abandonScene, forkScene } = useScene();
   const encounterActive = useEncounterStore((s) => s.active || s.showConfigModal);
 
   // Sprite sidebar settings from chat metadata
@@ -721,6 +721,14 @@ export function ChatArea() {
       );
     },
     [activeChatId, branchChat],
+  );
+
+  const handleCloneSceneFromHere = useCallback(
+    (messageId: string) => {
+      if (!activeChatId) return;
+      forkScene(activeChatId, "clone", { upToMessageId: messageId });
+    },
+    [activeChatId, forkScene],
   );
 
   // Peek prompt state
@@ -1353,12 +1361,14 @@ export function ChatArea() {
           onToggleConversationStart={handleToggleConversationStart}
           onPeekPrompt={handlePeekPrompt}
           onBranch={isSceneChat ? undefined : handleBranch}
+          onCloneSceneFromHere={isSceneChat ? handleCloneSceneFromHere : undefined}
           onToggleSelectMessage={handleToggleSelectMessage}
           onSummaryContextSizeChange={handleSummaryContextSizeChange}
           onRerunTrackers={handleRerunTrackers}
           onStartEncounter={() => startEncounter()}
           onConcludeScene={() => concludeScene(activeChatId)}
           onAbandonScene={() => abandonScene(activeChatId)}
+          onForkScene={forkScene}
           onOpenSettings={() => setSettingsOpen(true)}
           onOpenFiles={() => setFilesOpen(true)}
           onOpenGallery={() => setGalleryOpen(true)}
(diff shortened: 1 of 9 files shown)


LukaTheHero:
## Review of #230 — Scene Fork Roleplay Actions

Solid feature work. The PR description is unusually clear (motivation, screenshots for both flows, manual-test notes), and the architecture lands cleanly: most of the complexity stays in one route (`packages/server/src/routes/scene.routes.ts`), the shared types are defined once and reused on both sides, and `createMessagesBatch` getting a per-message `extra`/`swipeExtra` passthrough is a nice low-risk generalization.

### Things that look good

- **Metadata stripping is genuinely thorough.** `SCENE_FORK_METADATA_EXCLUDE` plus the `key.startsWith("scene")` catch-all in `buildRoleplayForkMetadata` covers every scene field I could find in the codebase (`sceneOriginChatId`, `sceneInitiatorCharId`, `sceneDescription`, `sceneScenario`, `sceneSystemPrompt`, `sceneRating`, `sceneStatus`, `sceneConversationContext`, `sceneRelationshipHistory`, `sceneBackground`, `activeSceneChatId`, `sceneBusyCharIds`). Cross-checked against the prompt assembly path in `generate.routes.ts` (the `chatMeta.sceneStatus === "active"` branch and friends around lines 2002 and 2459-2517) — none of those keys leak through. The fork really does drop the chat back into normal roleplay codepaths.
- **Hidden continuity injection** as a `narrator` role with `hiddenFromUser: true` is the right call — `messageRoleSchema` already accepts `narrator`, and `ChatRoleplaySurface`'s new `isHiddenFromUser` filter keeps it out of the transcript while still feeding prompt assembly.
- **Re-fork guard.** The `isForkingRef` ref + `isForking` state pair in `use-scene.ts` is the correct pattern (state for re-render, ref for the synchronous guard).
- **Convert respects the no-summary contract.** Verified the route does *not* call into the summary/memory path used by `/scene/conclude` — character `extensions.characterMemories` is untouched, and no narrator summary message is injected into the origin.

### Questions / clarifications

1. **`includeParticipationGuide=false` skips only the first narrator.** The loop checks `!skippedParticipationGuide && msg.role === "narrator"` and bails after the first hit. Today `/scene/create` only ever injects one narrator participation guide, so this works — but if a user later adds OOC narrator messages mid-scene, the flag becomes a no-op for them. Worth a one-line comment, or rename the option to `skipFirstNarrator` so the contract matches the implementation. Either way, the hook always sends `true`, so this is dead code at the moment — is the option meant to be user-configurable later?
2. **Permitted to clone a *concluded* scene?** The validator is `if (!isActiveScene && !originChatId) return 400`. A concluded scene still has `sceneOriginChatId`, so it passes through and gets cloned. Probably desirable, but worth confirming — the UI only exposes the button on active scenes, but the API endpoint accepts it. Document the intent.
3. **No origin existence check for `convert`.** The route reads the origin via `chats.getById(originChatId)` and silently no-ops if it's missing. If the origin was deleted out from under the scene, convert will still succeed and remove the scene. That's probably the right outcome, but logging an info line would help debugging.

### Suggestions

- **`Clone from here` button doesn't visually disable during a fork.** In `ChatArea.tsx` the prop is gated with `isSceneChat && !isForking ? handleCloneSceneFromHere : undefined`, which makes the button vanish from the DOM mid-click rather than dimming. `ActionBtn` already supports a `disabled` prop (see `ChatMessage.tsx:1841`) — passing `onCloneSceneFromHere` unconditionally and adding `disabled={isForkingScene}` to the two `ActionBtn` instances around lines 1419 and 1755 would match how the Convert button handles it (`disabled={isForking}` in `SceneBanner.tsx`).
- **Clone has no confirmation.** Convert prompts via `showConfirmDialog`, but a misclick on `Clone from here` silently spawns a new chat and navigates away. A toast "Scene cloned — undo" pattern, or a lightweight confirm if the scene has >N messages, would soften that.
- **Chat naming: scenes that don't start with `Scene: `.** `name.startsWith("Scene: ") ? replace : `${name} (roleplay)`. User-renamed scenes hit the second branch; a user who reads the suffix may not realize it's a fork. Minor, but consider `${name} — clone` vs `${name} — converted` to disambiguate the two modes.

### Risks / edge cases worth confirming

- **Atomicity.** Fork is not transactional: `chats.create` → `updateMetadata` → `createMessagesBatch` → (convert) `disconnectChat` → `remove`. If `createMessagesBatch` throws halfway, you're left with an empty roleplay chat in the sidebar. Convert is the dangerous case — the new chat is created *before* the original is removed, so worst-case is a duplicate, not data loss, but a try/catch that deletes the half-built fork on failure would be cleaner.
- **CI didn't run.** The status rollup only shows the (paused) CodeRabbit context — `pnpm-validate` and `container-build-test` haven't fired against `3f52540`. The branch may need a rebase or empty commit to trigger workflows before merging.
- **Manual tests I'd want reported.** (1) Clone from the very first message of a scene (zero prior messages besides the participation guide + first message). (2) Clone a scene with active swipes that have their own attachments — verify `swipeExtra` round-trips. (3) Convert while a generation is streaming on the scene — does the active stream race with the `chats.remove`? (4) Verify lorebooks bound to the scene via `activeLorebookIds` survive into the fork (they should — `buildRoleplayForkMetadata` doesn't strip them).

Overall the design is sound and the surface area is well-scoped. The dead `includeParticipationGuide` option, the disabled-state inconsistency, and the lack of fork-cleanup on partial failure are the main things I'd want addressed; everything else is a question or a polish suggestion.

---

OnlyJimmy:
Thanks for the thorough review. I went through the points you raised and pushed follow-up fixes.

Addressed:

- Added comments clarifying that includeParticipationGuide=false skips the generated first narrator participation guide, not arbitrary later narrator messages.
- Documented that cloning inactive scene-like chats with an origin is intentional, so concluded/older scene transcripts can still be recovered into standalone roleplay.
- Added an info log when convert proceeds but the origin chat no longer exists.
- Changed Clone from here so it remains visible but disables during fork activity, matching the Convert button behavior.
- Added best-effort cleanup for partially-created fork chats if message copy or convert cleanup fails after the new chat is created.
- Updated fork chat naming to use - clone / - converted.
- Disabled Clone/Convert while the scene is actively streaming after manual testing showed convert mid-generation could hit an SQLite race.
- Fixed swipe preservation: forked messages now preserve all message swipes, active swipe index, swipe metadata, and timestamps.

I intentionally left clone confirmation out for now. Clone from here feels closer to the existing branch/fork interaction, and the in-flight/streaming disabled states cover the accidental double-submit/race cases without adding extra friction.

Additional manual validation:

- Clone creates the expected disambiguated name.
- Lorebook metadata travels with the clone.
- Clone from the first scene message works.
- Clone and Convert both disable appropriately during streaming.
- Message swipes are preserved during cloning.


I'm not able to restart CodeRabbit reviews myself

LukaTheHero:
Awesome, just reviewed and tested on my install, great work! pulling in now!

PR #231, opened 2026-04-25
fix(game): strip JSON-content tags fully in GM narration parser

Romuromylus:
## Why this change

In Game Mode, GM dialogue sometimes ends in a trailing `}]` for users on models that invent their own structured tags. The catch-all stripper in `parseGmTags`, `stripGmTags`, and `stripGmTagsKeepReadables` uses `/\[\w+:[^\]]*\]/g`, which matches up to the **first** `]`. When the GM emits an unrecognized tag whose body contains a JSON array (e.g. `[some_tag: {"x":[1,2]}]`), the regex matches `[some_tag: {"x":[1,2]` (stopping at the inner `]`) and leaves `}]` trailing in user-visible content.

## What changed

- Added `stripUnknownBracketTags` in `packages/client/src/lib/game-tag-parser.ts`: a quote-aware walker that tracks `"`/`'` strings (with `\` escapes) and `[`/`]` depth, so JSON content inside a model-invented tag is removed along with its outer brackets.
- Replaced the three `/\[\w+:[^\]]*\]/g` callsites — in `parseGmTags`, `stripGmTags`, and `stripGmTagsKeepReadables` — with `stripUnknownBracketTags`.
- Optional `keep` predicate so `stripGmTagsKeepReadables` can skip `[Note:]` and `[Book:]` (preserving previous inline-readables behavior).
- No changes to recognized-tag parsing and no public API changes — only the catch-all behavior is touched.

## Validation

- [x] `pnpm check` (TypeScript + ESLint clean; client build succeeds)

### Manual verification notes

- Deployed the patched build to a private instance and reproduced the original bug on the unpatched baseline (Game Mode, model that emits structured tags). After the patch, the trailing `}]` no longer appears.
- Spot-checked recognized-tag parsing in Game Mode (`[music:]`, `[choices:]`, `[combat:]`, `[direction:]`, `[widget:]`, `[Note:]`, `[Book:]`) — all still parsed and stripped as before.
- `[Note:]` / `[Book:]` content still rendered inline via `stripGmTagsKeepReadables`.

## Docs and release impact

- [x] No docs changes needed

N/A — backend tag-parser fix, output is plain text rendered in the existing GM dialogue surface. The user-visible difference is the absence of trailing `}]` in narration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tag parsing reliability when tags contain nested brackets or quoted strings, ensuring complete tag removal in all cases.

### DIFF
--- a/packages/client/src/lib/game-tag-parser.ts
+++ b/packages/client/src/lib/game-tag-parser.ts
@@ -877,8 +877,13 @@ export function stripGmTagsKeepReadables(content: string): string {
     .replace(/\[party-turn\]/gi, "")
     .replace(/\[party-chat\]/gi, "")
     .replace(/\[dice:\s*[^\]]+\]/gi, "");
-  // Quote-aware catch-all for unknown tags, keeping Note/Book inline
-  text = stripUnknownBracketTags(text, (name) => name === "Note" || name === "Book");
+  // Quote-aware catch-all for unknown tags, keeping Note/Book inline.
+  // Case-insensitive to match extractBalancedTags (which lowercases the prefix);
+  // otherwise `[note:]` / `[book:]` would slip past extraction and get stripped.
+  text = stripUnknownBracketTags(text, (name) => {
+    const lower = name.toLowerCase();
+    return lower === "note" || lower === "book";
+  });
   // Balanced bracket stripping for non-readable tags
   text = stripBalancedTag(text, "[map_update:");
   text = stripBalancedTag(text, "[choices:");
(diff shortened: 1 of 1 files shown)


Romuromylus:
Thanks good catch. `extractBalancedTags` lowercases its prefix when matching `[Note:]`/`[Book:]`, so a lowercase tag would slip past extraction and get stripped here. Fixed in [26832b7](https://github.com/Romuromylus/Marinara-Engine/commit/26832b7) by lowercasing the tag name in the keep predicate.

LukaTheHero:
Thank you! I checked it, nice job! merging!

PR #233, opened 2026-04-25
feat: Oracle agent — web search with lorebook persistence

### LINKED ISSUE #232: [Feature]: Oracle agent — let characters reference live web facts via <search> triggers

Right now, characters are effectively stuck at the point where their card/lore was written.
If a user wants to roleplay (or just CONVO - as i'm on it =^^=) around something current “what’s on at the cinema this weekend?”, “is there a rugby match this Sunday?”, “what are the latest patch notes for X?”, the character has no good way to answer. It either invents something plausible, or breaks immersion by saying it does not know.
The existing Knowledge Retrieval agent only searches the local lorebook, so there is currently no path for live information.

The workaround is to manually edit a lorebook entry every time the user wants the character to know something current. That works, but it is tedious and reverses the normal storytelling flow: you have to write the answer before asking the question.

## Proposed solution

A new built-in **Oracle** agent, opt-in, that fires when the user types `<search>topic</search>` inside a chat message. 

The flow is:
  1. **Trigger extraction**: server strips `<search>` tags from the user message before saving / before the main LLM
  sees it.
  2. **Web fetch**: Tavily search (MVP; provider abstraction so we could add Brave / Exa / SerpAPI later).
  3. **Strict summary**: Oracle summarizes the results with an anti-hallucination prompt: no first person, no character voice, no inferred facts
  4. **Injection** : the summary is wrapped in `<web_search>` and appended to the user message before the main generation,
  so the character answers from real material.
  5. **Memory persistence (optional)** : after the character replies, a second LLM pass distils what the character
  "committed to" such as choices, preferrenced or dismissed options and saves that as a lorebook entry tagged `web-research`. The
  takeaway is saved, not the raw web facts and urls, so the lorebook stays "relevant".

Configuration would lives under Settings → Connections → Oracle with options, ex: 
* provider, 
* encrypted API key, 
* max results (1–10), 
* summary token cap (100–2000),
* timeout, 
* auto-persist toggle.

### Slash command
`/web_search query --save`   
This probably fits best with the existing slash-command system in packages/client/src/lib/slash-commands.ts, alongside /roll, /sys, /narrator, /scene, etc.

advantages:  
* it is discoverable through /help and autocomplete
* flags like --save, --no-save, or --max=5 are cleaner than XML-style attributes
* typos fail loudly, for example /web_serch, instead of leaking a malformed tag into the conversation

I still think the inline tag fits this specific use case better, because Oracle is not just a lookup utility. It is meant to augment a specific turn.

With a slash command, the flow tends to become:  

    User:  /web_search films this week  
    System: [results displayed]  
    User:  Which one do you want to see?  
    Char:  ???  (No context)  

At that point, the results still need to get into the character’s context somehow. Either the app re-injects them into the next user message, generates some kind of filler response, or asks the user to copy-paste the findings.  

So i prefered to "search" using the message that *prompts* the character to use it:  

    User:  Which one do you want to see? <search>films this week</search>  
        ↓ server strip + inline injection + main LLM call   
    Char:  Masters of the Universe looks amazing, want to try the 8pm showing?  

That said, both can coexist. A later /web_search command could call the same executeOracle() backend for users who want a plain lookup outside an in-character message

### Generic tool calling
Letting the model decide when to search would be more automatic, but I do not think it is the right MVP.

It makes behavior harder to debug: did the model search or not? It also makes every message a possible search candidate, which has cost and latency implications. It is also less reliable for small local models that may not follow tool schemas consistently.
   
### Extending Knowledge Retrieval
KR currently read-only against the local lorebook. Adding web access there would mix two very different pipelines: local retrieval and live external search. They have different cost, latency, failure modes, and privacy expectations.  

### Manual lorebook entry  
This is the current workaround. It works, but it does not scale and it breaks the flow of roleplay
   
## Additional context
* MVP scope: one query per turn, Tavily-only.
* Non-goals for v1: streaming search results, multi-provider fan-out, scheduled background searches.
* I already have an implementation ready in a draft PR, but I am happy to rework or drop the approach.
* Documentation: `docs/ORACLE.md` covers the full flow + design decisions.
* Currently only using it sparingly in CONVO 
* This issue and the related PR were written with LLM assistance, the design and implementation are based on my own use case and testing 
* Non breaking change
* Implementation draft #233 

## Template check

Beherit79:
## Why this change

Closes #232

Characters in Marinara are currently stuck with whatever was written in their lore/card. If a user wants to roleplay around something current, e.g a real-world event, recent news, an actual schedule, there is currently no clean way to get that information into the character's context. The existing Knowledge Retrieval agent only reads from the local lorebook.

This PR adds **Oracle**, an opt-in pre-generation agent triggered by `<search>topic</search>` in a user message.

For the MVP, Oracle fetches web results through Tavily, summarises them with a strict anti-hallucination prompt, injects the summary into the current turn, and can optionally persist a post-reply takeaway as a lorebook entry tagged `web-research`.

The persistence step intentionally saves what the character actually committed to, what they picked, preferred, dismissed, or remembered, instead of saving raw web results. That keeps future turns useful for RP without filling the lorebook with URLs, provider snippets, or stale search output.

### Server
  - Added `services/agents/oracle.ts` for the Oracle flow: search → summarise → takeaway extraction.
  - Added a web-search provider abstraction under `services/agents/web-search/`, with Tavily as the first implementation.
  - Added `/api/oracle/{config,test}` routes.
    - API keys are encrypted at rest with the existing `crypto` util.
    - GET responses return a masked key.
    - PUT preserves the stored key when the mask is sent back unchanged.
  - Updated `generate.routes.ts`:
    - extracts `<search>` triggers from the user message
    - bypasses the chat-level `enableAgents` gate when Oracle is explicitly user-triggered
    - runs Oracle alongside the other pre-generation agents
    - persists the post-generation takeaway when enabled
  - Updated `agent-executor.ts` so agents can explicitly provide `settings.maxTokens`.
    - Oracle uses a smaller output budget, around 600 tokens, so more input budget is available for source material.
    - The existing 16384 default floor is still preserved when no explicit value is set, so other agents should be unaffected.
  - Updated `lorebook-keeper-utils.ts` with a `source?: { agentId, description }` parameter so non-LBK agents like Oracle can attribute auto-created lorebook entries correctly.

### Shared
  - Added `types/oracle.ts` for the Zod config schema, masked-key constant, and lorebook tag constant.
  - Updated `types/agent.ts`:
    - Oracle entry in `BUILT_IN_AGENTS`
    - Oracle entry in `DEFAULT_AGENT_TOOLS`
    - `web_search` in `BUILT_IN_TOOLS`
  - Added Oracle's default summariser prompt in `constants/agent-prompts.ts`.

### Client
  - Added `OracleConfigCard.tsx` in the Connections panel.
    - provider
    - API key, with masked/debounced save
    - max results
    - summary token cap
    - timeout
    - auto-persist toggle
    - test button
  - Added `useOracle` React Query hook for config get/save/test.

### Docs
  - Added `docs/ORACLE.md` covering the end-to-end flow, design decisions, failure modes, and known limits.
  - Updated `docs/CONFIGURATION.md` with an Oracle section under Agent-Specific Configuration.
  - Updated `README.md` to mention Oracle in the AI Agent System paragraph and Documentation table.
  - Updated `CHANGELOG.md` under `[Unreleased] / Added`.

## Validation

  - [x] `pnpm check`
  - [x] Manual verification completed (described below).

### Manual verification notes

- Configured Oracle with a Tavily key. The Test button returns `ok: true` with a non-zero result count.
- Sent a chat message containing `<search>European rugby match this weekend</search>`.
  - Server logs show:
    - `[oracle] Detected 1 <search> trigger(s)`
    - `[oracle] Search returned 6 result(s)`
    - `[oracle] Summariser result — success=true`
  - The character used the actual fixtures in its in-character reply.
- After the reply, the takeaway pass either:
  - captured a single-sentence commitment, for example “X picked the Bordeaux match over Leinster…”
  - or correctly emitted `NO_MEMORY (skipped)` when the character only acknowledged the info without making a choice.
- With `autoPersist: true`, a valid takeaway creates or reuses a lorebook entry tagged `web-research`, visible in the lorebook panel.
- Disabling Oracle in the agent panel makes `<search>` tags pass through as plain text. No API call is made and no Oracle log line is emitted.
- Empty or whitespace-only `<search>` blocks are ignored without errors.
- A misconfigured key, for example a provider 401, logs `[oracle] Search failed` and the character still replies normally without injected findings.

## Docs and release impact

  - [ ] No docs changes needed
  - [x] Updated docs (README / CONFIGURATION / ORACLE / CHANGELOG) as needed

This is a feature PR, not a release PR. I did not touch version-bearing files. The changelog entry is under `[Unreleased]` for maintainers to include in the next release.

### Oracle agent entry

image
  -
image
  -
image
  -
image

### Settings → Connections → Oracle config card

Masked key and Test button visible.
  
image

### Chat with `<search>` trigger and in-character reply

mari_search_oracle
  -
image

_No, it is not biased lol._

### Lorebook entries (Uncategorized)

image
  -
image

### Lorebook entry details

Different test run from the previous screenshot.

image

### Retrieval

image
 
## Status

This PR is intended as a concrete proposal for Oracle rather than a final “this is the only shape” implementation.

The code is ready for review and manual testing passes, but I am open to changing the trigger syntax, persistence behavior, config surface, or cutting scope if maintainers prefer a smaller first version.

The issue and this PR description were written with LLM assistance. The feature itself comes from my own use case, and the implementation was manually tested in CONVO.

## Summary by CodeRabbit

* **New Features**
  * Oracle agent for live web search functionality
  * Trigger searches via `<search>topic</search>` tags in messages
  * Automatic summarization and injection into character context
  * Optional persistence of character insights to lorebook
  * Configure in Settings → Connections → Oracle

* **Documentation**
  * Added Oracle agent setup and usage documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 This file is the release-notes source of truth for Marinara Engine. Reuse these entries when publishing GitHub Releases for tags in the `vX.Y.Z` format.
 
+## [Unreleased]
+
+### Added
+
+- New agent: **Oracle** — fetches live web results when the user types `<search>topic</search>` in a chat message, summarises them with anti-hallucination rules, injects the summary into the character's context, and persists the character's takeaway (preferences, choices, opinions) to the lorebook with the `web-research` tag. Tavily is the MVP provider; configure the API key and behaviour from Settings → Connections → Oracle. See `docs/ORACLE.md` for the end-to-end flow.
+
 ## [1.5.5]
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ Character expression sprites with automatic emotion switching, custom scene back
 
 ### AI Agent System
 
-25+ built-in agents that run alongside your chat — world state tracking, quest management, combat, expression detection, background selection, narrative direction, prose analysis, Spotify DJ, CYOA choices, and more. All disabled by default; enable only what you want, or create custom agents.
+25+ built-in agents that run alongside your chat — world state tracking, quest management, combat, expression detection, background selection, narrative direction, prose analysis, Spotify DJ, CYOA choices, an Oracle that answers `<search>topic</search>` triggers with live web results (via Tavily) and persists character takeaways to the lorebook, and more. All disabled by default; enable only what you want, or create custom agents.
 
 ### Prompt Engineering
 
@@ -167,6 +167,7 @@ Export chats as JSONL or plain text. Fully local SQLite database — all data st
 | [docs/TROUBLESHOOTING.md](docs/TROUBLESHOOTING.md) | Common issues and fixes                                         |
 | [docs/FAQ.md](docs/FAQ.md)                         | Frequently asked questions (LAN access, etc.)                   |
 | [docs/FRONTEND.md](docs/FRONTEND.md)               | Frontend architecture, components, hooks, and API reference     |
+| [docs/ORACLE.md](docs/ORACLE.md)                   | Oracle agent — web search trigger, summary, and persistence     |
 | [android/README.md](android/README.md)             | Android WebView wrapper (APK) guide                             |
 | [CONTRIBUTING.md](CONTRIBUTING.md)                 | Contributor workflow, validation, versioning, and release steps |
 | [CHANGELOG.md](CHANGELOG.md)                       | Release notes                                                   |
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -57,6 +57,26 @@ LOG_LEVEL=debug pnpm start
 
 > **Note:** Client-side (browser) logging uses the standard `console.*` API and is not controlled by `LOG_LEVEL`. Production client builds automatically strip `console.log` calls; only `console.warn` and `console.error` survive in the browser.
 
+## Agent-Specific Configuration
+
+Some agents have configuration that lives outside `.env` — typically because it includes encrypted credentials managed through the UI rather than environment variables.
+
+### Oracle (Web Search)
+
+The Oracle agent fetches live web results when the user types `<search>topic</search>` inside a chat message, summarises them with anti-hallucination rules, injects the summary into the character's context, and (optionally) persists character takeaways to the lorebook. It is configured from **Settings → Connections → Oracle**, and the encrypted API key is stored alongside other application settings.
+
+| Setting            | Default   | Range / Values    | Description                                                                                                                                              |
+| ------------------ | --------- | ----------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `enabled`          | `false`   | boolean           | Master switch for the Oracle agent. The agent must also be enabled in the agent panel for it to fire.                                                    |
+| `provider`         | `tavily`  | `tavily`          | Web-search provider. Only Tavily is implemented in the MVP.                                                                                              |
+| `apiKey`           | _(empty)_ | string            | Provider API key. Encrypted at rest using `ENCRYPTION_KEY`. The UI displays a masked placeholder once a key is saved.                                    |
+| `maxResults`       | `3`       | `1`–`10`          | Number of raw search results requested from the provider. The summariser still caps its own output independently.                                        |
+| `summaryTokenCap`  | `400`     | `100`–`2000`      | Hard cap on the summary length handed back to the character. Smaller values are faster and cheaper; larger values keep more facts.                       |
+| `timeoutMs`        | `8000`    | `2000`–`30000`    | Fetch timeout (ms) before Oracle gives up and returns an empty injection.                                                                                |
+| `autoPersist`      | `true`    | boolean           | When the character makes a durable commitment after a search (a choice, a preference, an opinion), persist it as a lorebook entry tagged `web-research`. |
+
+See [`docs/ORACLE.md`](./ORACLE.md) for the end-to-end flow, the trigger syntax, and how summaries are injected and persisted.
+
 ## Notes
 
 - The shell launchers (`start.bat`, `start.sh`, `start-termux.sh`) source `.env` automatically. If you run `pnpm start` directly, make sure the variables are set in your environment or `.env` file.
--- a/docs/ORACLE.md
+++ b/docs/ORACLE.md
@@ -274,6 +274,27 @@ chat-scoped lorebook loses everything when the user opens a fresh chat. So:
 Future Oracle runs with the same character reuse that one lorebook, so entries
 accumulate instead of spawning a new book per turn.
 
+### Why Oracle entries are persisted as `constant: true`
+
+Lorebook entries normally activate by matching keywords against the recent
+chat. That model breaks for conversational recall: when the user asks
+*"which film were we going to see again?"* a week after the search, the
+message contains none of the original entity names (`Wicked`, `Avatar 3`)
+and the entry never fires — the character forgets.
+
+Oracle entries are written with `constant: true`, which makes them
+always-on: whenever the character's lorebook is in scope, every takeaway
+the character has accumulated is injected, no keyword matching required.
+This is safe because the entries are already character-scoped — they only
+ever appear in chats with that specific character, never bleeding into
+other roleplays.
+
+The trade-off is context size: each takeaway is a single short sentence
+(~30-40 tokens), so 30 entries adds roughly 1k tokens of permanent context.
+Acceptable on modern model windows; if it becomes a problem in practice,
+the user can prune entries from the lorebook panel, and a future iteration
+could cap the count to the N most recent automatically.
+
 ### Why Oracle bypasses the chat's `enableAgents` gate
 
 In conversation mode, Marinara's settings drawer hides the "Enable Agents"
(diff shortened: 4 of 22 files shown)


LukaTheHero:
## Overall impressions

This is a substantial, well-thought-through proposal — the description is excellent, the screenshots cover the happy path and lorebook side effects, the docs in `docs/ORACLE.md` are unusually thorough for a feature PR, and the design has clearly been iterated on (the takeaway-vs-raw-facts distinction is the kind of detail that only comes from actually using the thing). I went through the diff and the new files in `packages/server/src/services/agents/oracle.ts`, the routes, the web-search abstraction, the shared types, and the changes in `generate.routes.ts` and `agent-executor.ts`. I did not run the code; what follows is from reading.

I have one blocking concern (a regression in `generate.routes.ts`), a few security/robustness items, and several smaller suggestions. Treating you as a serious contributor — none of this is "no, never" feedback, more "this is what I'd want to see before merge".

## Things that look good

- **Crypto handling.** `packages/server/src/services/agents/web-search/config.ts` keeps the encrypted blob in storage, decrypts only on `loadOracleConfig`, and the route layer (`oracle.routes.ts`) returns the masked constant on GET and short-circuits to preserve the existing key when the mask is sent back unchanged on PUT. The mask-preservation path correctly avoids re-encrypting the mask string.
- **No key in logs.** `oracle.ts` never logs `apiKey`. It does log queries, source-material length, and a 200-char preview of the summary — fine for a self-hosted tool, worth being aware of.
- **Pino logger usage.** All new server modules import `../../lib/logger.js` and use `logger.info/warn/debug` with correct format specifiers. No `console.*` slipped in.
- **Sterile summariser context.** The deliberate emptying of `recentMessages`, `persona`, `characters` etc. in `executeOracle` (oracle.ts ~L240) is the right call — small models would absolutely continue the roleplay otherwise.
- **Provider abstraction.** `WebSearchProvider` interface + `web-search-provider-factory.ts` with the exhaustive `never` switch is clean. Adding Brave/Serper later is genuinely a one-file change.
- **Graceful degradation.** Tavily 401/429/5xx is caught in `executeOracle`, logged at `warn`, and returns `success: false` with empty injection. The pre-gen orchestrator excludes Oracle from the critical-failure gate (`generate.routes.ts` ~L3724), so a bad Tavily key cannot break generation.
- **Defaults are conservative.** `enabledByDefault: false` in `BUILT_IN_AGENTS`, `enabled: false` in the Zod schema, and the UI ships in the off state.

## Blocking concerns

**1. Regression: the preset-less roleplay/visual-novel lorebook injection block was removed.**

The diff for `packages/server/src/routes/generate.routes.ts` deletes the entire `// ── Lorebook injection for preset-less roleplay / visual_novel ──` block (44 lines, around the original L1951). That block was added in PR #225 (commit `3662c9e`) which is upstream of this branch's merge base — so this is not a "before my time" deletion, it's an active regression that will reintroduce #225.

This needs to be restored during the rebase. Given that the PR is currently `mergeStateStatus: DIRTY / CONFLICTING`, this is a good moment to do both at once.

**2. Stray edits to character-command SSE bookkeeping.**

Same file, near the bottom of the diff — the `try { ... } finally { trySendSseEvent(... assistant_commands_end ...) }` wrapper around `collectedCommands` processing is removed, and so is the matching `assistant_commands_start` event. This looks like an accidental revert, not an intentional change for Oracle. Please put it back.

## Security and robustness

**3. `<search>` parser: code blocks bypass.** `extractSearchQueries` (`character-commands.ts` ~L505) runs `/<search>([\s\S]*?)<\/search>/gi` over the full message. If a user pastes a code fence containing the literal string `<search>foo</search>`, it will be treated as a real trigger and stripped from the message. Easy fix: skip text inside fenced code blocks (you already do similar with `parseCharacterCommands` on the assistant side).

**4. `<search>` parser: no length cap on the query.** `raw.trim()` then dedup — a user can put an arbitrarily large blob inside the tag. Worth a soft cap (256-512 chars) before it goes to Tavily.

**5. `enableAgents` bypass scope.** `generate.routes.ts` ~L3375 — when `oracleSearchQueries.length > 0` the code looks up the Oracle row directly via `agentsStore.getByType("oracle")` and constructs a `ResolvedAgent` outside the normal resolution path. That's correctly scoped to *just* Oracle (no other agent types are affected) and gated on `oracleRow.enabled === "true"`, so the bypass is fine. Worth a one-line comment in `docs/ORACLE.md` clarifying that the bypass does not lift the **agent-level** enabled flag — only the chat-level one — because the doc currently reads as if Oracle "always works once configured".

**6. No rate-limit / per-turn cap.** A user can put `<search>...</search>` on every message and you'll hit Tavily + run two extra LLM calls (summariser + takeaway) on every turn. The MVP only processes the first query (`oracleSearchQueries[0]`), which bounds the per-turn cost — good — but there's nothing stopping per-message abuse over time. Probably out of scope for v1, but worth a `TODO` and a mention in the doc that the user is responsible for their Tavily quota.

**7. `oracle.routes.ts` `/test` swallows errors silently.** The try/catch returns `{ ok: false, error: msg }` to the client but never logs server-side. If a deployer is debugging why Test is failing, they'll see nothing in the server logs. Add `logger.warn(err, "[oracle] /test failed")` in the catch.

## Behavioral concerns

**8. Lorebook entry overwrite, not accumulation.** `persistLorebookKeeperUpdates` dedupes existing entries by `entryName.toLowerCase()` and **overwrites `content`**. Oracle picks the entry name as `payload.keys[0] ?? payload.query` truncated to 120 chars. So if the character makes commitment A about "Bordeaux" in turn 5 and commitment B about "Bordeaux" in turn 12, B silently replaces A — the older memory is lost. Two options: (a) append/merge content for Oracle entries, or (b) make the entry name more disambiguated (e.g. include a short hash of the query or the turn id). Worth at least documenting the current behavior in `docs/ORACLE.md` under "known limits".

**9. `NO_MEMORY` sentinel detection is fragile.** `oracle.ts` ~L100 trims, strips wrapping quotes, then checks `cleaned === "NO_MEMORY" || /^no[_ ]memory$/i.test(cleaned)`. A model that emits `"NO_MEMORY (skipped)"` or `"NO_MEMORY: nothing committed"` will be treated as a real takeaway and persisted. Tighten the test to allow the sentinel as a prefix, or use a more distinctive token.

**10. Input budget concern.** `summaryTokenCap` is for output. The input is `buildSourceMaterial(response)` which concatenates Tavily's `answer` plus up to 10 results × `search_depth: "advanced"` (~2-4× the basic content). I didn't see any truncation before it hits the summariser prompt. On a small-context chat connection that could blow past the model window. Worth either capping per-result content length or computing total chars and trimming.

**11. `agent-executor.ts` floor change — confirmed safe.** The new `typeof explicitMaxTokens === "number" && explicitMaxTokens > 0 ? explicitMaxTokens : 16384` preserves the 16384 floor when no `maxTokens` is set, so existing agents (none of which set it) are unaffected. Good.

## Suggestions / scope

- **Doc-vs-code drift in `docs/ORACLE.md`.** The flow diagram says "synthesises a hook if message is empty" — `extractSearchQueries` does *not* synthesise; the synthesis happens later in `generate.routes.ts` (the `lastUserIdx < 0` branch). Minor wording fix.
- **Consider shipping without `autoPersist` defaulted to `true`.** Users opting in to Oracle are also opting their lorebook into automatic writes. Defaulting `autoPersist: false` (Zod schema currently `.default(true)`) and letting users flip it on once they have seen what Oracle produces would be friendlier. Not blocking, just worth thinking about.
- **CodeRabbit flagged docstring coverage at 46%.** Most of the missing ones are small helpers (e.g. `normalizeResult`, `buildSourceMaterial`, the React `FieldRow`/`ToggleRow` components). Easy sweep before merge.

## Risks worth confirming

- Rebase the branch. `mergeStateStatus: DIRTY` — the conflict is almost certainly the deleted lorebook-injection block from PR #225. Resolving it correctly (i.e. *keeping* the block) also fixes blocking concern #1.
- Re-run `pnpm check` after the rebase, especially because `agent-executor.ts` changed shape.
- Sanity-check that an Oracle lorebook created in a single-character chat shows up correctly in the **lorebook panel filter UI** when filtered by `sourceAgentId === "oracle"` — your screenshots show the entries exist, but I cannot see from the diff whether the UI exposes that filter today.

Thanks for the level of care that went into this — the takeaway/raw-facts split is genuinely the right call, and the strict summariser prompt is well-shaped. With the regression restored and the parser hardening, this is in good shape to merge.

---

SpicyMarinara:
Holding off until issues are addressed.

SpicyMarinara:
Closing since the author never got back.

PR #241, opened 2026-04-26
feat(game): jump back to a segment by clicking it in Session Logs

Romuromylus:
## Why this change

In Game Mode, the Session Logs modal is a read-only history. If you scroll past a beat and want to re-experience it (replay the typewriter for that exact narration segment, or reground yourself in a moment of dialogue), there's no way to get back — `setActiveIndex` is internal to `GameNarration` and segments past the active one have no affordance. This adds the natural interaction: click a logged row from the current scene to make it the active narration segment again.

## What changed

- `packages/client/src/components/game/GameNarration.tsx`: each row in the Session Logs modal that belongs to the latest assistant message (the "current scene") now accepts a click. Clicking sets `activeIndex` to that segment's index in the live `segments` list, resets `visibleChars` via `getSegmentStartVisibleChars` (so the typewriter replays — or reveals instantly if `gameInstantTextReveal` is on), closes the modal, and clears any in-progress log edit state.
- Past-scene rows remain non-interactive: `latestAssistant` is the only message the narration stage can play, so older messages would require a destructive "delete everything after" rewind. Out of scope here — this PR is scoped strictly to the cheap, non-destructive case.
- Click handler short-circuits on `target.closest("button, input, textarea, a")` so the existing edit/delete pencil/trash buttons (and inline edit textareas) keep working without bubbling into a jump.
- Hover affordance: jumpable rows get `cursor-pointer` and a faint `hover:ring-1 hover:ring-white/15`. The currently-active segment keeps its existing `ring-1 ring-[var(--primary)]/40` and is excluded from the jump (it would be a no-op).

Wired across all four segment renderers in the modal — dialogue, system, readable (Note/Book), and the default narration row.

## Validation

- [x] `pnpm check`

### Manual verification notes

Built and deployed to the author's EasyPanel instance off fork main; verified in-browser:

- Hovering a current-scene row shows the white hover ring and pointer cursor; tooltip reads "Jump back to this segment".
- Clicking a current-scene row closes the modal and rewinds the narration stage to that segment, with the typewriter replaying from the start of that segment.
- The currently-active segment shows the existing primary-color ring and does not pick up the hover affordance.
- Past-scene rows (older AI messages) are not interactive — no hover ring, default cursor, no click effect.
- Edit (pencil) and delete (trash) icons on hover continue to work normally; clicking them does not also trigger a jump.
- Persistence: the rewound `activeIndex` flows through the existing `onSegmentChange` path and is written to both `localStorage` (`narration-idx:<chatId>`) and the server's `gameNarrationIndex`, so a refresh restores the rewound position.

## Docs and release impact

- [x] No docs changes needed

## Summary by CodeRabbit

* **New Features**
  * Log rows can be focused and clicked (or activated via Enter/Space) to jump back to the corresponding live narration segment when applicable.
  * Jump restores the conversation view to that point, closes the logs modal, clears log-edit state, resets scroll tracking, and plays a click sound.
  * Eligible rows show hover/focus ring styling; in-row controls are ignored so their interactions aren’t interrupted.

### DIFF
--- a/packages/client/src/components/game/GameNarration.tsx
+++ b/packages/client/src/components/game/GameNarration.tsx
@@ -1,7 +1,16 @@
 // ──────────────────────────────────────────────
 // Game: Narration Area (VN-style segmented box)
 // ──────────────────────────────────────────────
-import { useEffect, useMemo, useState, useCallback, useRef, type ReactNode, type MouseEvent as ReactMouseEvent } from "react";
+import {
+  useEffect,
+  useMemo,
+  useState,
+  useCallback,
+  useRef,
+  type ReactNode,
+  type MouseEvent as ReactMouseEvent,
+  type KeyboardEvent as ReactKeyboardEvent,
+} from "react";
 import DOMPurify from "dompurify";
 import { MessageCircle, RefreshCw, ScrollText, X, Package, Pencil, Check, Play, Pause, Trash2 } from "lucide-react";
 import { cn } from "../../lib/utils";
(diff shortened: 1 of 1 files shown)


Romuromylus:
Both CodeRabbit nits applied in a548148 (and mirrored to the deployment).

- **Click SFX**: now plays via the existing `playClickSfx()` for parity with `nextSegment` and the auto-play tick.
- **Keyboard / a11y**: row gets `role="button"`, `tabIndex={0}`, and an `onKeyDown` that fires on Enter/Space (with the same "ignore clicks inside buttons/inputs/textareas" guard as the click handler). Added a `focus-visible:ring-1 ring-white/30` so keyboard focus is discoverable. The attrs only attach when `canJumpToSeg` is true so non-jumpable rows aren't announced as buttons.

I refactored the per-row attrs into a single `jumpRowProps` spread so all four segment renderers (dialogue, system, readable, narration) stay in sync.

Skipping the docstring-coverage warning — generic 80% threshold isn't a fit for a 31-line UI patch.

Romuromylus:
Good catch — the lucide icons render as SVG, so `e.target` on an icon click is an `SVGElement`/`SVGPathElement`, which fails `instanceof HTMLElement`. The original `as HTMLElement` cast in the first revision was a TS-only assertion (no runtime check), which is why it accidentally worked; tightening it to a real `instanceof` introduced the bypass.

Fixed in 9082dbd by widening the check to `instanceof Element`. `Element.closest()` is defined on the same interface, so the `button, input, textarea, a` selector still resolves the ancestor button regardless of where in the icon tree the click landed.

LukaTheHero:
It's a great addition, I'd love to merge, but I need @SpicyMarinara to confirm if they have any plans to implement something or have a different vision on this.

LukaTheHero:
Honestly, this is a great feature, I'm adding this now, mari can change it if they want to rework it in the future to something else.

PR #257, opened 2026-04-27
refactor(server): drop duplicate key details from inline conversation summaries

JorgeLTE:
## Summary

In conversation mode, every per-day and per-week summary's `keyDetails` were being injected into the prompt twice:

1. Inline, appended to each `<summary date="...">` / `<summary week="...">` block in the chat history.
2. As the canonical `<important_memories>` block at the end of `conversationSystemPrompt`, where they are already labeled by date and chronologically sorted.

This PR keeps `<important_memories>` as the single source of truth for key details. The inline `<summary>` blocks now hold the narrative recap only.

## Test plan

- [x] `pnpm install`
- [x] `pnpm check` (lint + build) passes
- [x] Manual: open a conversation-mode chat that already has at least one daily summary and one consolidated weekly summary; confirm the model still references key details correctly and that summaries no longer carry an inline `Key details: …` line in the request payload

## Release Notes

* **Refactor**
  * Optimized the internal processing of conversation summaries to improve the organization and accessibility of key details within the system.

### DIFF
--- a/packages/server/src/routes/generate.routes.ts
+++ b/packages/server/src/routes/generate.routes.ts
@@ -1315,14 +1315,11 @@ export async function generateRoutes(app: FastifyInstance) {
               const wEntry = weekSummaries[weekKey]!;
               const monday = parseDateKey(weekKey);
               const sunday = new Date(monday.getFullYear(), monday.getMonth(), monday.getDate() + 6);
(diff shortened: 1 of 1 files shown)


LukaTheHero:
Thank you! great catch!

PR #270, opened 2026-04-27
Fix sidecar CUDA runtime setup

munimunigamer:
Now installs cuda .dll files rather than relying on user to have cuda installed into their system already

## Summary by CodeRabbit

* **New Features**
  * Added Windows CUDA runtime dependency management.

* **Bug Fixes**
  * Improved local sidecar availability detection for tracker agents.
  * Enhanced model file validation to prevent unsupported formats.
  * Fixed fallback behavior when local sidecar is unavailable.

* **Tests**
  * Added model file validation tests.
  * Added Windows environment setup tests.


SpicyMarinara:
Thanks, Muni!

Issue #226, opened 2026-04-25
[Feature]: Add knowledge-router agent: lower-cost alternative to knowledge-retrieval


cha1latte:
## Background

In Discord, a user proposed a leaner alternative to the existing `knowledge-retrieval` agent. SpicyMarinara replied:

> "This is a potential for a separate agent or an extension, but as of now, I am not going to bother. If anyone makes this, I will add to the main branch."

*[(link to Discord thread)](https://discord.com/channels/1417099416812392641/1491088609950957669)*

I'd like to take this on. Opening this issue as a tracking hub for the design and the PRs that will land it.

## Problem

The current `knowledge-retrieval` agent has two costs that grow with lorebook size:

1. **Redundant context.** Entries marked `constant: true` are already injected by standard lorebook activation, but they also get re-summarized by the retrieval agent. Same content, two copies, two token bills.
2. **Full extraction pass.** All enabled entries from the configured source lorebooks are fed into an LLM call that summarizes them. For large lorebooks (One Piece, Wheel of Time, sprawling worldbuilding), this gets chunked into multiple passes — multiplying tokens and latency before main generation has even started.

## Proposed solution

A new agent type `knowledge-router` that *routes to* entries instead of *extracting from* them.

Pipeline:

1. Build a compact catalog from the configured source lorebooks: one line per entry containing `name + description`. Skip `constant: true` entries (already in context).
2. Send one LLM call: chat history + catalog + system prompt asking for a JSON array of relevant entry IDs.
3. Load those entries' full content from the DB and inject verbatim as a context block — same shape as standard lorebook activation, no LLM-generated summary in between.
4. Return an `AgentResult` of type `context_injection`.

Both agents coexist. Users who prefer the current behavior keep it. The agent settings UI will warn if both are enabled in the same pipeline.

## Schema change

Lorebook entries don't currently have a description field. Adding one:

- New column `description: text` on `lorebookEntries`, default empty string.
- Migration handles existing databases (no data migration needed; empty descriptions fall back to entry name + first key when building the catalog).
- Entry editor in the client gets a small textarea with a placeholder explaining the field's purpose.

## Implementation plan

Splitting into three reviewable PRs that land in order:

| # | Scope | Depends on |
|---|---|---|
| 1 | Schema + migration + entry editor textarea for `description` | — |
| 2 | `knowledge-router` agent: prompt, type registration, server wiring, tests, benchmark | PR 1 |
[...]

PR #295, opened 2026-04-28
feat(agents): add Knowledge Router — lower-cost lorebook routing agent

cha1latte:
Part 2 of #226 — the Knowledge Router agent itself.

## Summary
Adds a new built-in pre-generation agent, **Knowledge Router**, as a lower-cost alternative to the existing Knowledge Retrieval agent for large lorebooks.

Where Knowledge Retrieval reads the *full content* of every enabled entry from the source lorebooks and summarizes it (chunking when needed), the router reads a *short catalog* (one row per entry: id, name, keys, summary) and returns the IDs of the entries it thinks are relevant. Selected entries are then injected verbatim — no per-entry summarization pass.

## Why
For large lorebooks (One Piece, Wheel of Time, etc.) the existing Knowledge Retrieval agent gets expensive — it has to summarize every enabled entry on every turn, often across multiple chunked LLM calls. The router avoids that by routing *between* entries instead of summarizing them.

The two agents complement each other rather than replacing one another. Power users with smaller, well-curated lorebooks may still prefer Retrieval for its summarization quality; users with large lorebooks get a cheaper option.

## How it works (short version)
1. Route layer loads enabled entries from the agent's source lorebooks. `constant: true` entries are skipped (they already inject unconditionally via the standard activation pipeline — routing them would duplicate work).
2. The router builds a catalog row per candidate entry: `<entry id="..." name="..." keys="...">summary</entry>`.
3. The catalog `summary` uses the entry's `description` field (added in #227) when non-empty, otherwise falls back to the first ~60 tokens of the entry's content. This keeps the router useful out of the box on existing lorebooks while letting power users tune precision via descriptions.
4. The agent LLM returns `{ "entryIds": [...] }`.
5. Selected entries are looked up and injected verbatim into the prompt as `<knowledge_router>` (or `## Knowledge Router` under markdown wrap), parallel to how `<knowledge_retrieval>` is injected.

The router runs in parallel with Knowledge Retrieval and the standard pre-gen agent pipeline via the same `Promise.all` gate — so enabling both adds at most the slower of the two latencies, not their sum.

## Architecture (commit-by-commit)

**Feature commits:**
| Commit | Layer |
|---|---|
| `33eebee` | Shared — register agent type + default prompt |
| `2fa39d6` | Server agents — `executeKnowledgeRouter` + `<entry_catalog>` injection in agent-executor |
| `17dd7fc` | Server routes — wire into generate pipeline (loader, runner promise, prompt injection, regen filter) |
| `496f70a` | Tests — unit tests for the pure helpers + empty-entries early-return |
| `1d19c51` | Cost-model script for the PR description (see below) |
| `074b9f0` | Client — extend the existing Knowledge Sources picker to also show for Router (file uploads stay Retrieval-only) |

**Review-feedback commits (CodeRabbit passes 1 & 2):**
| Commit | What |
|---|---|
| `d4d7c83` | knowledge-router: escape all catalog fields, dedupe IDs, logger.error |
| `8c6a869` | generate: router try/catch, regen-cache rehydration fix, entryStateOverrides |
| `d46c6fe` | client: don't persist sourceFileIds for knowledge-router agent |
| `4f4b990` | benchmark script: process.stdout.write instead of console.log |
| `372b82b` | knowledge-router: trim whitespace from parsed entry IDs |
| `cea6e1f` | generate: KR try/catch parity + ephemeral projection + correct regen replay |

## Cost model
Modeled (not benchmarked) input/output tokens for both agents at three lorebook sizes. The script (`packages/server/scripts/benchmark-knowledge-router.ts`) uses the real `buildCatalog` helper to measure actual catalog sizes, then estimates LLM costs analytically. All assumptions documented at the top of the file so reviewers can challenge or re-run with their own numbers.

Default-assumption results (~50% description coverage, 800-char avg entry content):

| Entries | Knowledge Retrieval | Knowledge Router | Router vs KR (input) |
|---|---|---|---|
| 20 | 4,750 in / 300 out (1 call) | 1,465 in / 40 out (1 call) | **30.8%** |
| 100 | 23,150 in / 1,200 out (4 calls) | 5,775 in / 200 out (1 call) | **24.9%** |
| 500 | 114,550 in / 5,400 out (18 calls) | 27,763 in / 1,000 out (1 call) | **24.2%** |

The 500-entry case is the killer — KR makes 18 sequential LLM calls, Router makes 1.

## Known limitations
- **Router quality scales with description quality.** With 0% description coverage, every catalog row is a content snippet — accuracy drops on entries whose first ~60 tokens aren't representative. The hybrid fallback keeps the agent functional out of the box, but the encouraged path is for users to write tight descriptions on their important entries.
- **Catalog capped at 400 entries.** Larger lorebooks are truncated (order-preserving) with a warning log to prevent context-window blowups. The router still works on the truncated set; a future enhancement could rank or paginate before truncating, but real-world lorebooks rarely approach this scale.
- **No "both agents enabled" warning UI yet.** If a user enables both Knowledge Retrieval and Knowledge Router on the same lorebook, both will fire and inject — the model handles it fine but it's redundant work. Warning + a description-coverage badge are planned for the next PR in this set.
- **Inconsistency: Knowledge Retrieval's loader does not honor `chatMeta.entryStateOverrides`.** The router's loader does (added in this PR after CodeRabbit feedback) — KR's existing loader has the same gap and should be brought in line in a follow-up. Not addressed here to keep this PR's scope tight. (Note: KR's parallel-promise rejection bug *was* incidentally fixed in `cea6e1f` since the try/catch pattern was trivial to apply to both.)

## Test plan
Manual smoke tests performed on a fresh dev install (`pnpm dev`), Roleplay chat, 5-entry test lorebook ("Bingus Bongus studios") with mixed description coverage (3/5 with description, 2/5 with blank description to exercise the content fallback):

- [x] Router fires on a fresh send. Console: `[Agent] ✓ Knowledge Router (knowledge-router) — 10.1s` with verbatim entry text.
- [x] Router selected the right 2 of 5 entries for the test prompt ("tell me about the studio's mascot and where the team works") — both with blank descriptions, picked via content fallback.
- [x] Selected entries appear in the chat response (mascot/Beans + Greenhouse details verbatim).
- [x] Router does NOT re-fire on regen — only one `[Agent] ✓ Knowledge Router` line across original + regen.
- [x] `constant: true` entries are filtered before reaching the router. Flipped Beans to Constant; router's text payload no longer contained `### Beans the Salamander`. Beans content still appeared in the response (via the standard activation pipeline).
- [x] Both Knowledge Router and Knowledge Retrieval enabled simultaneously — both fire in parallel, total wall-clock bounded by the slower one (~12s, not ~22s). Both outputs reach the prompt without collision.
- [x] `pnpm check` passes locally.
- [x] 19/19 unit tests pass (`packages/server/test/knowledge-router.test.ts`).

## Summary by CodeRabbit

* **New Features**
  * Added Knowledge Router agent to select relevant lorebook entries during generation
  * Agent Editor: configure Knowledge Router and choose knowledge sources (router shows router-specific help; uploaded files remain retrieval-only)
  * Generation pipeline: Knowledge Router runs alongside retrieval to enrich prompts and improves regeneration so cached injections are replayed in the correct position

* **Bug Fixes**
  * Saving agent settings no longer preserves stale uploaded-file IDs when switching to Router

### DIFF
--- a/packages/client/src/components/agents/AgentEditor.tsx
+++ b/packages/client/src/components/agents/AgentEditor.tsx
@@ -222,6 +222,8 @@ export function AgentEditor() {
 
   // Knowledge Retrieval agent — lorebook source selector
   const isKnowledgeRetrievalAgent = agentDetailId === "knowledge-retrieval" || dbConfig?.type === "knowledge-retrieval";
+  // Knowledge Router agent — also uses the lorebook source selector (file picker stays Retrieval-only)
+  const isKnowledgeRouterAgent = agentDetailId === "knowledge-router" || dbConfig?.type === "knowledge-router";
   const { data: allLorebooks } = useLorebooks();
   const { data: allKnowledgeSources } = useKnowledgeSources();
   const uploadSource = useUploadKnowledgeSource();
@@ -993,12 +995,16 @@ export function AgentEditor() {
             </FieldGroup>
           )}
 
-          {/* ── Knowledge Source Lorebooks (only for Knowledge Retrieval agent) ── */}
-          {isKnowledgeRetrievalAgent && (
+          {/* ── Knowledge Source Lorebooks (Knowledge Retrieval + Knowledge Router) ── */}
+          {(isKnowledgeRetrievalAgent || isKnowledgeRouterAgent) && (
             <FieldGroup
               label="Knowledge Sources"
               icon={<BookOpen size="0.875rem" className="text-amber-400" />}
-              help="Select lorebooks and/or upload files for this agent to scan. Supported file types: .txt, .md, .csv, .json, .xml, .html, .pdf"
+              help={
+                isKnowledgeRouterAgent
+                  ? "Select lorebooks for this agent to route over. The router picks relevant entries by id and they're injected verbatim."
+                  : "Select lorebooks and/or upload files for this agent to scan. Supported file types: .txt, .md, .csv, .json, .xml, .html, .pdf"
+              }
             >
               <div className="space-y-4">
                 {/* ── Lorebooks ── */}
@@ -1048,7 +1054,8 @@ export function AgentEditor() {
                   )}
                 </div>
 
-                {/* ── Uploaded Files ── */}
+                {/* ── Uploaded Files (Knowledge Retrieval only) ── */}
+                {isKnowledgeRetrievalAgent && (
                 <div className="space-y-1.5">
                   <p className="text-[0.6875rem] font-medium text-white/60">Files</p>
                   {/* File list */}
@@ -1153,6 +1160,7 @@ export function AgentEditor() {
                     )}
                   </button>
                 </div>
+                )}
 
                 {/* Summary */}
                 {(localSourceLorebookIds.length > 0 || localSourceFileIds.length > 0) && (
--- a/packages/client/src/components/agents/AgentEditor.tsx
+++ b/packages/client/src/components/agents/AgentEditor.tsx
@@ -305,7 +305,13 @@ export function AgentEditor() {
         enabledTools: localEnabledTools,
         ...(localSpotifyClientId ? { spotifyClientId: localSpotifyClientId } : {}),
         ...(localSourceLorebookIds.length > 0 ? { sourceLorebookIds: localSourceLorebookIds } : {}),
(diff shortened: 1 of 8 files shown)


cha1latte:
Blessings upon you, let me know if theres anything i could do better if you see anything (I'm pretty new at Github)

PR #314, opened 2026-04-29
fix(sidecar): preserve Windows PATH key casing to prevent CUDA backend failure

### LINKED ISSUE #315: Sidecar child process loses system PATH on Windows due to env key casing
On Windows, `buildLlamaProcessEnv` spreads `process.env` into a plain object, then writes to `env.PATH`. 
Windows natively stores the key as `"Path"` (title-case). While `process.env` is a case-insensitive Proxy, the spread result is a regular case-sensitive object. Writing to `env.PATH` creates a duplicate key, causing the child process to inherit only the runtime directory instead of the full system PATH.
This strips NVIDIA driver paths and `C:\Windows\System32`, preventing `ggml-cuda.dll` from resolving dependencies like `nvcuda.dll` and `cublas64_*.dll`, causing silent CUDA failure and automatic CPU fallback.
Affected file: `packages/server/src/services/sidecar/sidecar-runtime-env.ts` line 32.

Image

Lochalan:
Closes #315
## Why this change

On Windows, `buildLlamaProcessEnv` in `sidecar-runtime-env.ts` spreads `process.env` into a plain object. While `process.env` itself is a case-insensitive Proxy, the spread result is a plain object that preserves the native key casing — typically `"Path"` (title-case), not `"PATH"`.

The original code writes to `env.PATH`, which creates a **duplicate key** when the existing key is `"Path"`. The short uppercase `PATH` (containing only the runtime directory) takes precedence, stripping the child process of all system directories — including `C:\Windows\System32` and NVIDIA driver paths.

This prevents `ggml-cuda.dll` from resolving its implicit dependencies (`nvcuda.dll`, `cublas64_*.dll`), causing silent CUDA load failure and automatic CPU fallback on every startup.

### Before fix (old runtime b8851 without bundled CUDA DLLs)

image

### Controlled A/B test — same machine, same session

```
=== OLD CODE (env.PATH = ...) ===
PATH-like keys in env: [ 'Path', 'PATH' ]   ← duplicate key
CUDA loaded: false

=== NEW CODE (env[pathKey] = ...) ===
PATH-like keys in env: [ 'Path' ]            ← single key
CUDA loaded: true
ggml_cuda_init: found 1 CUDA devices (Total VRAM: 16375 MiB)
```

## What changed

- **`sidecar-runtime-env.ts`**: Use `Object.keys(env).find(k => k.toLowerCase() === "path")` to detect the actual key casing before prepending. Falls back to `"PATH"` if no key exists.
- **`sidecar-runtime-env.test.ts`**: Added regression test verifying that a `"Path"` key is modified in-place without creating a duplicate `"PATH"` key.

## Compatibility

This fix is complementary to #270 (bundling CUDA DLLs). With bundled DLLs, the CUDA symptom is masked because DLLs are found via same-directory resolution. However, the child process still loses its entire system PATH, which could affect other functionality. This fix addresses the root cause.

## Validation

- [x] `pnpm check`
- [x] Controlled test spawning `llama-server --version` with old vs new env construction
- [x] Verified GPU offload works with fix, breaks without it (on runtime b8851)
- [x] Regression test passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved Windows environment handling that could create duplicate PATH entries. Child processes now inherit the correct PATH, preventing issues with system libraries and GPU/DLL detection.

### DIFF
--- a/packages/server/src/services/sidecar/sidecar-runtime-env.ts
+++ b/packages/server/src/services/sidecar/sidecar-runtime-env.ts
@@ -29,8 +29,14 @@ export function buildLlamaProcessEnv(
 
   if (platform === "win32") {
     const runtimeDir = win32.dirname(runtime.serverPath);
-    env.PATH = prependPathEntry(
-      prependPathEntry(env.PATH, runtimeDir, win32.delimiter),
+    // Windows exposes process.env as a case-insensitive Proxy, but spreading
+    // it into a plain object preserves the native key casing (typically "Path").
+    // Writing to env.PATH when the real key is "Path" creates a duplicate,
+    // causing the child process to lose its system PATH and breaking DLL
+    // resolution for CUDA and other GPU backends.
+    const pathKey = Object.keys(env).find((key) => key.toLowerCase() === "path") ?? "PATH";
+    env[pathKey] = prependPathEntry(
+      prependPathEntry(env[pathKey], runtimeDir, win32.delimiter),
       runtime.directoryPath ?? runtimeDir,
       win32.delimiter,
     );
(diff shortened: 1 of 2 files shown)


LukaTheHero:
@munimunigamer

PR #379, opened 2026-05-03
Fix Windows installer clone staging

TheLonelyDevil9:
## Summary
- stage fresh Windows installer clones under `%TEMP%` instead of inside `$INSTDIR`
- handle robocopy failure return codes and clean external staging on success/failure
- add a static guard to prevent reintroducing unsafe `$INSTDIR\repo-temp` parent/child copy layouts

## Validation
- `node ./scripts/check-windows-installer-layout.mjs`
- `pnpm test`

Note: `makensis` was not available locally, so full NSIS build validation still needs a Windows machine/VM with NSIS installed.

## Summary by CodeRabbit

* **Tests**
  * Added a pre-test validation that detects unsafe installer staging/cloning patterns to prevent accidental in-place operations.

* **Bug Fixes**
  * Installer now performs clones and staging in per-run temporary locations, with retry logic, explicit error detection, reliable cleanup on failure, and clearer failure messaging to avoid partial or corrupted installs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "start": "node packages/server/dist/index.js",
     "format": "prettier --write \"packages/**/*.{ts,tsx}\"",
     "lint": "pnpm -r run lint",
-    "test": "pnpm -r run test",
+    "test": "node ./scripts/check-windows-installer-layout.mjs && pnpm -r run test",
     "version:sync": "node ./scripts/sync-version.mjs",
     "version:check": "node ./scripts/check-version-drift.mjs",
     "guard:installer-artifacts": "node ./scripts/check-tracked-installers.mjs",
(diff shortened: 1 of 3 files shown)


TheLonelyDevil9:
Addressed both CodeRabbit findings in 635dce5.

Changes:
- Added a sibling `$INSTDIR.__stage` publish step so robocopy failures only leave temp/stage directories, not a half-populated `$INSTDIR` with `.git`. The installer removes an empty `$INSTDIR`, renames the completed stage into place, and aborts safely if the target is not empty.
- Hardened `scripts/check-windows-installer-layout.mjs` to ignore comments, catch unquoted direct unsafe paths, detect variables assigned to `$INSTDIR\repo-temp`, and flag their later use in `git clone` / `robocopy`.

Validation:
- `node ./scripts/check-windows-installer-layout.mjs`
- negative guard check against a temporary bad installer assignment
- `pnpm build:shared && pnpm test`
- `git diff --check`

Still not run: `makensis` because NSIS is not installed/found locally.

PR #415, opened 2026-05-04
fix(game): tighten bg fuzzy matcher so new scenes don't pin to first library bg

### LINKED ISSUE #414: [Issue]: Game Mode keeps reusing first library bg instead of generating a new one
## Summary

When starting a new scene in Game Mode (with image generation enabled), the client never asks the server to generate a fresh background — it pins the scene to whichever library bg comes first in iteration order. On installs that include the legacy SillyTavern bg pack, this is `backgrounds:user:ancient_library`, so unrelated scenes appear to "stick" to the ancient library image.

## Expected behavior

A novel scene location emitted by the scene model (e.g. `backgrounds:fantasy:tavern-interior`) should trigger generation through `/game/generate-assets` and the freshly generated image should render. A scene whose location genuinely overlaps an existing library bg (e.g. one containing both "dark" and "forest" → `backgrounds:user:dark_forest`) should still reuse the library bg.

## Actual behavior

The client's `resolveAssetTag` always returns a library tag with score 1 — because `tagScore` counts the universal category prefix word (`backgrounds`) on both sides as a +1 match. Then `bestMatch` accepts any score > 0 and returns the first tag it walked, which on most installs is `backgrounds:user:ancient_library` (alphabetically first in the legacy bg folder).

The downstream effect inside `applySceneResult`:

1. `setCurrentBackground(resolved)` is called with the wrong library tag immediately — the user sees the wrong bg.
2. The `unresolvedBg` find then concludes the bg is already resolved (it lives in the manifest) and `backgroundTag` is **omitted** from the `/game/generate-assets` payload — so no generation is ever requested.

Server logs `[game/scene-wrap] bg "<chosen>" not in manifest; generation will be deferred`, but the deferred follow-up call from the client never names the bg, so nothing is generated.

## Steps to reproduce

1. Have the legacy `data/backgrounds/` SillyTavern bg pack populated (any install where the user dropped in or kept the default 13 jpgs — `ancient_library.jpg`, `dark_forest.jpg`, etc.).
2. Have empty `data/game-assets/backgrounds/{fantasy,modern,scifi}/`.
3. Start a fresh Game Mode chat with sprite/image generation enabled and a configured image connection.
4. Send the first user input that should land in a brand-new location (e.g. a bustling tavern).
5. Observe: the displayed background is `ancient_library.jpg` (or whichever bg is first in the manifest), and `/game/generate-assets` is called without `backgroundTag`, so nothing is generated.

## Environment

- Marinara Engine version: 1.5.7
- Install type: docker (EasyPanel-deployed Dockerfile build of fork main)
- OS + version: Linux x86_64 container on Linode (host Ubuntu 22.04)
- Browser or app shell: Chromium-based browser

Server stdout for the affected turns shows `[game/scene-wrap] bg "X" not in manifest; generation will be deferred` followed by no `[game/generate-assets]` background generation entry. The client console emits `[asset-resolve] "<scene-tag>" → "backgrounds:user:ancient_library"` (or whichever stock bg won the score=1 tie).

## Additional context

A previous attempt to address a similar-sounding "wrong/sticking-bg" symptom on the multi-segment server path (commit `00e97fe`) was reverted (`4afed2d`) when the deployed instance still showed the bug, because the actual root cause is on the client-side fuzzy matcher rather than in scene-wrap's multi-tag generation. PR coming.

## Template check

Romuromylus:
## Linked issue

Closes #414

## Why this change

In Game Mode with image generation enabled, starting a new scene renders the wrong background and never triggers a fresh generation. On installs that include the legacy SillyTavern `data/backgrounds/` pack (which gets manifest-tagged as `backgrounds:user:<name>`), unrelated scenes pin to whichever bg comes first in iteration order — typically `backgrounds:user:ancient_library`.

The root cause is a fuzzy-matcher that's too lax:

- `tagScore` in `packages/client/src/lib/asset-fuzzy-match.ts` counts the universal category prefix word (`backgrounds`) on both the model-emitted tag and every candidate manifest tag, giving every comparison a free +1.
- `bestMatch` accepts any `score > 0`, so that single free point declares a match and returns the first tag walked.

Downstream in `applySceneResult`:

1. `setCurrentBackground(resolved)` immediately sets the wrong bg.
2. The `unresolvedBg` find concludes the bg is already in the manifest, so `backgroundTag` is omitted from the `/game/generate-assets` payload — no generation request is ever made.

Server-side logs `[game/scene-wrap] bg "<chosen>" not in manifest; generation will be deferred`, but the deferred call from the client never names the bg, so nothing is generated and the scene stays pinned to the wrongly-matched library image.

## What changed

`packages/client/src/lib/asset-fuzzy-match.ts`:

- `tagScore` now takes a `category` argument and excludes the category word from both prose words and tag parts. The category prefix is universal across every candidate, so counting it adds noise without signal.
- `bestMatch` now requires at least 2 specific-word overlaps (initial `bestScore = 1`, accept on `s > bestScore`). A brand-new location with only the (already stripped) category prefix in common no longer gets pinned to an arbitrary library bg.
- The single call site in `resolveAssetTag` was updated to pass the category through.

Backwards-compatible behavior preserved:

- Tags that genuinely overlap a library bg (≥ 2 specific-word matches) still resolve to the library entry — e.g. a scene tag containing both "dark" and "forest" still reuses `backgrounds:user:dark_forest`.
- Tags that already exist exactly in the manifest take the early return at line 73 and never hit the fuzzy matcher.
- Backgrounds with no fuzzy match still fall through to the `backgrounds:generated:<slug>` branch, which is what triggers the `/game/generate-assets` request the bug was suppressing.

No public type changes (`tagScore` and `bestMatch` are file-local).

## Validation

- [x]  Checked edge cases (light + dark mode, mobile viewport, empty states, error paths)

### Manual verification notes

Tested on a Dockerfile build of this branch deployed to a private EasyPanel instance:

1. Started a fresh Game Mode chat with an image-generation connection configured and `enableSpriteGeneration` on. Confirmed legacy `data/backgrounds/` pack present (13 jpgs incl. `ancient_library.jpg`) and `data/game-assets/backgrounds/{fantasy,modern,scifi}/` empty.
2. Sent the first user message, expected the scene to land in a brand-new location.
3. Before fix: the bg snapped to `ancient_library.jpg` and stayed there; server logs showed `bg "X" not in manifest; generation will be deferred` followed by no `/game/generate-assets` background generation entry.
4. After fix: the scene model's tag fell through to `backgrounds:generated:<slug>`, the `/game/generate-assets` call carried `backgroundTag`, and the freshly generated image rendered.
5. Verified that a scene tag containing words that genuinely overlap a library bg (≥ 2 specific-word matches) still reuses the library entry rather than re-generating.

Edge cases box left unchecked: this is a non-UI client lib change with no light/dark, mobile, or empty-state visual surface — the matcher is invoked from `applySceneResult` only. Server-rendered visuals were verified against an existing Game Mode chat.

## Docs and release impact

- [x] No docs changes needed

Not a UI change — affects the client-side asset-tag resolver only. The visible effect is "the right bg renders" (or generation begins) instead of "the wrong stock bg sticks". A demonstration would require generating a scene against a configured image provider, which I can attach as a follow-up if helpful.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved asset tag matching accuracy by excluding the category word from tag/prose comparisons.
  * Made background asset matching stricter to reduce false positives (requires greater overlap).
  * Avoided re-processing already-generated background tags to prevent incorrect rematches.

### DIFF
--- a/packages/client/src/lib/asset-fuzzy-match.ts
+++ b/packages/client/src/lib/asset-fuzzy-match.ts
@@ -38,15 +38,20 @@ function tagScore(prose: string, tag: string, category: string): number {
 }
 
 /**
- * Find the best-matching tag for a prose description. Requires at least two
- * specific-word overlaps so a brand-new location with only the (already
- * stripped) category prefix in common does not get pinned to whichever
- * library tag happens to come first in iteration order.
+ * Find the best-matching tag for a prose description. For backgrounds we
+ * require at least two specific-word overlaps so a brand-new location with
+ * only the (already stripped) category prefix in common does not get pinned
+ * to whichever library tag happens to come first in iteration order. Audio
+ * categories keep the original ≥1 threshold — single-word matches like
+ * "tense" or "combat" are meaningful in the deeper audio hierarchy and the
+ * "no fuzzy match" fall-through for those categories is silent (returns the
+ * original tag, no playback) rather than visually disruptive.
  */
 function bestMatch(prose: string, tags: string[], category: string): string | null {
   if (!tags.length) return null;
+  const minOverlap = category === "backgrounds" ? 2 : 1;
   let best: string | null = null;
-  let bestScore = 1;
+  let bestScore = minOverlap - 1;
   for (const tag of tags) {
     const s = tagScore(prose, tag, category);
     if (s > bestScore) {
--- a/packages/client/src/lib/asset-fuzzy-match.ts
+++ b/packages/client/src/lib/asset-fuzzy-match.ts
@@ -6,17 +6,24 @@
 // tags in the manifest using keyword overlap scoring.
 // ──────────────────────────────────────────────
 
-/** Score how well a prose description matches an asset tag by keyword overlap. */
-function tagScore(prose: string, tag: string): number {
+/**
+ * Score how well a prose description matches an asset tag by keyword overlap.
+ * The category word is excluded from scoring on both sides because it is the
+ * universal prefix of every tag in the manifest — counting it would give
+ * every candidate a free point and make the "first tag wins" tie-breaker
+ * pick an arbitrary library bg for any novel scene.
+ */
+function tagScore(prose: string, tag: string, category: string): number {
+  const categoryLower = category.toLowerCase();
   const words = prose
     .toLowerCase()
     .replace(/[^a-z0-9\s]/g, " ")
     .split(/\s+/)
-    .filter((w) => w.length > 2);
+    .filter((w) => w.length > 2 && w !== categoryLower);
   const parts = tag
     .toLowerCase()
     .split(/[:\-_]+/)
-    .filter((p) => p.length > 1);
+    .filter((p) => p.length > 1 && p !== categoryLower);
 
   let score = 0;
   for (const part of parts) {
@@ -30,13 +37,18 @@ function tagScore(prose: string, tag: string): number {
   return score;
 }
 
-/** Find the best-matching tag for a prose description. Returns null if no reasonable match. */
-function bestMatch(prose: string, tags: string[]): string | null {
+/**
+ * Find the best-matching tag for a prose description. Requires at least two
+ * specific-word overlaps so a brand-new location with only the (already
+ * stripped) category prefix in common does not get pinned to whichever
+ * library tag happens to come first in iteration order.
+ */
+function bestMatch(prose: string, tags: string[], category: string): string | null {
   if (!tags.length) return null;
   let best: string | null = null;
-  let bestScore = 0;
+  let bestScore = 1;
   for (const tag of tags) {
-    const s = tagScore(prose, tag);
+    const s = tagScore(prose, tag, category);
     if (s > bestScore) {
       bestScore = s;
       best = tag;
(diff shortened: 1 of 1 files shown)


Romuromylus:
Good catch — I had raised the threshold globally without thinking through the implications for music/sfx/ambient. The disruptive symptom is bg-specific (a visible wrong image sticks), while the audio categories fall through silently if nothing matches and have a deeper hierarchy where single-word matches like "tense" or "combat" still carry meaning.

Scoped the ≥2 threshold to `category === "backgrounds"` only; audio categories keep their original ≥1 threshold. The category-prefix exclusion in `tagScore` is unchanged and still applies across all categories.

Fixed in bbaf11f.

Romuromylus:
Applied. The current callers don't loop generated tags back through `resolveAssetTag` (the model emits original-format tags like `backgrounds:fantasy:moonlit-grove`, and `applyGeneratedAssets` sets the bg directly without re-resolving), but the guard is cheap, removes a class of "wrong bg sticks" regressions if a future refactor changes that, and also prevents the malformed nested fall-through tag (`backgrounds:generated:backgrounds-generated-X`) you'd otherwise get for a generated tag that isn't yet in the manifest.

Fixed in e6756bd.

SpicyMarinara:
Thank you so much! Lmk on Discord if you'd like ot join our team, you're very active!

PR #485, opened 2026-05-06
fix(game-map): preserve explicit move past metadata hydration

### LINKED ISSUE #484: [Issue]: Setting a destination on the map doesn't move the party — marker stays at the old location
## Summary

In Game Mode, clicking a node and confirming **Set destination** writes the move to the server but the party marker on the map stays at the previous location. The GM narrates the move correctly from the user's `*moves to X*` message text, but the map state never advances, so subsequent turns build their prompts from the old `partyPosition`.

## Expected behavior

After confirming a destination, `/api/game/map/move` should persist the new `partyPosition`. The map widget marker should jump to the chosen node, and the next `/generate` call should see the new position in `chat.metadata.gameMap.partyPosition`.

## Actual behavior

The marker stays at the previous node. The GM still narrates the player walking to the destination (because the user message contains `*moves to <label>*`), but the map widget never updates, and the next round of prompt building reads the stale `partyPosition` again.

## Steps to reproduce

1. Open or resume a Game Mode chat with a node-style map and an active scene.
2. Click an adjacent (or any reachable) node on the map widget.
3. Click **Set destination**, then send any message.
4. Watch the map widget after the GM responds — the marker stays at the old node.

## Environment

- Marinara Engine version: 1.5.8
- Install type: self-hosted Docker (EasyPanel)
- OS + version: server Debian/Linux; client Windows 11
- Browser or app shell: Chromium-based browser

No client-side errors. The DevTools console shows `/api/game/map/move` succeeding (200) and `/api/generate` running normally.

Root cause is server-side in `packages/server/src/routes/game.routes.ts`: the `/map/move` handler writes the new `partyPosition`, then runs `buildHydratedGameMeta`, which calls `syncGameMapMetaPartyPosition(meta, currentLocation)` where `currentLocation` is read from the latest `gameStateSnapshot.location` — and that snapshot still records the previous location until the next `/generate` run produces a new one. The sync silently snaps `partyPosition` back to the old node before persistence.

## Additional context

- Bug shipped together with the `/map/move` handler in #333; not a recent regression in `buildHydratedGameMeta` itself.
- A fix PR is on the way that re-applies the user's chosen position on top of the hydrated meta so the explicit move wins.

## Template check

Romuromylus:
## Linked issue

Closes #484

## Why this change

- After a player clicked **Set destination** on the Game Mode map, the marker stayed stuck at the previous node. The GM still narrated the move (because the user message contains `*moves to <label>*`), but `chat.metadata.gameMap.partyPosition` never advanced — so subsequent prompts kept building from the stale location and the map widget never updated.

## What changed

- In `packages/server/src/routes/game.routes.ts`, the `/map/move` handler now re-applies the user's chosen `partyPosition` on top of the hydrated meta after `buildHydratedGameMeta`. That helper calls `syncGameMapMetaPartyPosition(meta, currentLocation)` where `currentLocation` is read from the latest `gameStateSnapshot.location`. Until the next `/generate` run persists a new snapshot, that location still records the previous node, so the sync was silently snapping `partyPosition` back to the old node before persistence.
- A `withActiveGameMapMeta` call rebuilds `gameMaps[]` and `activeGameMapId` consistently with the corrected map.

## Validation

### Manual verification notes

- Built and deployed the patched server image to a self-hosted Docker (EasyPanel) instance running Marinara Engine 1.5.8.
- Opened an in-progress Game Mode chat with a node-style map and an active scene.
- Clicked an adjacent node, confirmed **Set destination**, and sent a turn.
- Before the fix: map marker remained on the previous node; on the next round the GM still framed the player as being at the old location even though it had narrated the move.
- After the fix: the map marker jumps to the chosen node immediately on send, and the next GM round narrates from the new location. The destination indicator clears as before. No client-side errors in DevTools.

## Docs and release impact

- [x] No docs changes needed

No UI code changed. Behaviour change is observed in the existing map widget — the marker now follows the user's destination click. Happy to attach a short screen recording if reviewers want one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Player moves now reliably take precedence over background updates so chosen positions are not reverted.
  * Deliberate location overrides are immediately reflected in saved game state and activity logs.
  * After a move, the displayed map is anchored to the persisted map so the final view matches the stored game state.

### DIFF
--- a/packages/server/src/routes/game.routes.ts
+++ b/packages/server/src/routes/game.routes.ts
@@ -2502,6 +2502,7 @@ export async function gameRoutes(app: FastifyInstance) {
   const buildHydratedGameMeta = async (
     chatId: string,
     baseMeta: Record<string, unknown>,
+    options: { explicitLocation?: string | null } = {},
   ): Promise<Record<string, unknown>> => {
     const gameStateStore = createGameStateStorage(app.db);
     const latestState = await gameStateStore.getLatest(chatId);
@@ -2513,7 +2514,12 @@ export async function gameRoutes(app: FastifyInstance) {
     }
 
     const activeQuests = extractActiveQuests(latestState?.playerStats);
-    const currentLocation = typeof latestState?.location === "string" ? latestState.location : null;
+    // Prefer a caller-supplied explicit location over the most recent snapshot. The snapshot's
+    // location field only refreshes after /generate persists a new game state, so callers that
+    // have just committed a deliberate move (e.g. /map/move) need to override it — otherwise the
+    // sync and journal reconciliation below run against the previous location.
+    const snapshotLocation = typeof latestState?.location === "string" ? latestState.location : null;
+    const currentLocation = options.explicitLocation ?? snapshotLocation;
     hydratedMeta = syncGameMapMetaPartyPosition(hydratedMeta, currentLocation);
     const currentJournal = (hydratedMeta.gameJournal as Journal) ?? createJournal();
     return {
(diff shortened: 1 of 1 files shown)


Romuromylus:
Valid point on internal consistency — agreed that re-applying `partyPosition` after `buildHydratedGameMeta` left both `syncGameMapMetaPartyPosition` and `reconcileJournal` running against the stale snapshot location.

Threaded the explicit move through hydration in 8f3bbea: `buildHydratedGameMeta` now takes an optional `explicitLocation` override that wins over the snapshot, so the location-derived reconciliation runs against the destination instead. The post-hydration `partyPosition` re-apply is kept as defence-in-depth against `syncGameMapMetaPartyPosition`'s label-name match landing on a different node when labels collide.

For the journal in particular: `addLocationEntry` early-returns when the location is already in `journal.locations`, so under the previous fix the practical drift was bounded (the new node was already added via the `discovered`-flag iteration earlier in `reconcileJournal`). But the principle is right — better to have hydration know the explicit move than to override after.

Romuromylus:
Valid edge case — `syncGameMapMetaPartyPosition` matches the explicit location string against every map's nodes/cells, so a label collision across maps could leave `hydratedMeta.gameMap` pointing at a different map than the one the click came from. Applied the suggested anchor in 03b4c51: `finalMap` is now resolved from the hydrated maps by `targetMapId` (with `updatedMap` as the fallback) before re-applying `partyPosition`.

Romuromylus:
Tested with mobile, should be safe to merge

Issue #260, opened 2026-04-27
[Issue]: Can't remove background.


Dark-Thoughts:
## Summary

When removing the background from a chat and you switch back & forth, the previous background will just return.

## Expected behavior

Background should stay removed.

## Actual behavior

Background switches back.

## Steps to reproduce

1. Set a background in a chat.
2. Remove the background.
3. Switch to another chat.
4. Switch back.

## Environment

- Marinara Engine version: v1.5.5
- Install type: Git
- OS + version: Bazzite:unstable
- Browser or app shell: Firefox

## Logs, screenshots, or video

-

## Additional context

-

## Template check

Please **uncheck (untick)** the box below before submitting so we know you read the template. It is intentionally pre-checked:

- [  ] I DID NOT read this template and provide the requested details.

LukaTheHero:
Fix is up in #275.

Three compounding issues were causing this. The most stubborn one was a stale React Query cache: \`useUpdateChatMetadata\` only invalidated the chat detail without writing the server response into the cache, so if the mutation completed after navigating away (or under certain switch-orderings), the cache held the pre-removal snapshot and the restore effect re-applied the supposedly-removed background. The Settings Remove handler also relied on a debounced effect chain that occasionally dropped the write, and the chat-switch restore effect only ever *set* the background, never *cleared* it — so a stale URL from a previous chat could leak through.

PR fixes all three:
- The metadata mutation now writes the server response into the cache directly.
- The Remove handler fires the persist mutation from the click, no debounce.
- The chat-switch restore effect mirrors metadata exactly, including null.

Verified locally: removed background now sticks across refreshes and chat switches in both directions.

PR #574, opened 2026-05-08
Fix chat background switch race

### LINKED ISSUE #547: [Issue]: Background image changes incorrectly when swapping between chats
## Summary

Moving between roleplay chats too quickly can fail to update the background image to the correct file, leaving the background image for the previous chat showing. This can also overwrite the background image URL associated with the swapped-to chat.

## Expected behavior

Every roleplay chat should have an associated background, which only changes when the user changes it in settings or when the "Background" agent changes it. The background image associated with the currently selected roleplay chat should be displayed. 

## Actual behavior

Swapping between two roleplay chats sometimes fails to update the background image, leaving the background for the former chat displayed behind the latter chat. This also rewrites the saved background for the latter roleplay chat.

## Steps to reproduce

1. Set up two roleplay chats, A and B, with two distinct backgrounds.
2. Swap from chat A to chat B, then swiftly back.
3. Chat A will now the the background originally assigned to chat B.

This may not happen every time, due to the race condition.

## Environment

- Marinara Engine version: v1.5.8
- Install type: Git
- OS + version: Windows 11
- Browser or app shell: Firefox, Chrome

## Additional context

This is related but distinct to #260.

My understanding is that this is the result of the two background-related `useEffect`s in `ChatArea.tsx` both having `chat?.id` as a dependency, meaning they can race each other and happen in the wrong order. The timeout that delays persisting the filename also contributes inconsistency.

I have a working fix locally that I can make into a PR.

## Template check

cha1latte:
## Linked issue

Closes #547

## Why this change

- Rapidly switching between chats could let the background metadata persist effect write the previous chat's background into the newly active chat before the metadata restore finished.

## What changed

- Track when `ChatArea` is restoring a chat background from metadata during a chat switch.
- Skip persistence while the UI background still reflects the previous chat, then resume normal persistence once the restored background value has landed.

## Validation

### Manual verification notes

- Codex verification: `node scratch/issue-547-background-race-repro.mjs old` reproduced the old race by overwriting chat B metadata from `b.png` to `a.png`.
- Codex verification: `node scratch/issue-547-background-race-repro.mjs fixed` preserved chat B as `b.png`, still persisted a manual background change after restore, and preserved a null saved background.
- Worker verification: `pnpm check` passed in the issue worktree.
- Coordinator review: re-ran both scratch repro modes after reviewing the diff.

## Docs and release impact

- [ ] No docs changes needed

No docs or release metadata changes included.

N/A - behavior is covered by a focused state-transition repro script.

### DIFF
--- a/packages/client/src/components/chat/ChatArea.tsx
+++ b/packages/client/src/components/chat/ChatArea.tsx
@@ -434,10 +434,17 @@ export function ChatArea() {
   // only set on truthy values, leaving the global chatBackground stale when
   // switching to a chat whose metadata has been cleared, which made a removed
   // background re-appear after a chat switch round-trip.
+  const restoredChatBackgroundRef = useRef<{ chatId: string | null; url: string | null; isSyncing: boolean }>({
+    chatId: null,
+    url: null,
+    isSyncing: false,
+  });
   useEffect(() => {
     if (!chat?.id) return;
     const bg = chatMeta.background as string | null | undefined;
-    useUIStore.getState().setChatBackground(bg ? `/api/backgrounds/file/${encodeURIComponent(bg)}` : null);
+    const restoredUrl = bg ? `/api/backgrounds/file/${encodeURIComponent(bg)}` : null;
+    restoredChatBackgroundRef.current = { chatId: chat.id, url: restoredUrl, isSyncing: true };
+    useUIStore.getState().setChatBackground(restoredUrl);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [chat?.id]);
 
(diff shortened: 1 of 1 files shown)


cha1latte:
@coderabbitai review

aidanbrocklebank:
Wow, thanks for the speedy PR that fixed the bug I raised!

It's great to have that fixed, but I do worry that that section of code is becoming a bit complicated and that more bugs like #547 and #260 are liable to arise again. I think it could be a bit simpler, and I'd be happy to write a follow-up PR cleaning it up a little, but I don't know if that's how the team likes to do things in this repo. Would that be welcomed, or is it better to leave it and move ahead onto other issues & features?

PR #703, opened 2026-05-11
fix(chat): make Up-Arrow recall always available and edit any role

LukaTheHero:
## Why

PR #639 introduced an Up-Arrow shortcut to recall the last message for inline editing, but two problems made it look broken after merge:

1. The shortcut was gated behind the existing **Intuitive swipe navigation** toggle, which defaults to `false`. Anyone who hadn't already opted into swipe nav pressed Up Arrow and saw nothing happen.
2. The help-text update describing the new behavior was attached to that same disabled toggle, so the discoverability hint was hidden behind the off switch too.

It also only ever recalled the most recent **user** message, which surprised users who expected "edit the last thing in the chat" to mean the last bubble period — including the AI's response.

## What this changes

- Adds a dedicated **"Up Arrow edits last message"** toggle in Settings, defaulted **on**, with its own help text. UI store version bumped `27 -> 28` so existing users pick up the new default.
- Splits the keydown handler in `ChatArea` so the Up-Arrow recall runs in its own effect, independent of the swipe-nav toggle and effect.
- Targets the most recent visible message regardless of role — user or assistant — skipping `hiddenFromUser` system messages and bailing while a generation is streaming or an agent is processing.
- Drops the user-only guard from the per-message event listeners in `ChatMessage` and `ConversationMessage` so any role responds to the recall event, matching the existing Pencil button behavior.
- Reverts the **Intuitive swipe navigation** help text to swipe-only wording now that Up Arrow lives in its own setting.

## Linked issue

No GitHub issue was opened for this — it's a follow-up to the merged PR #639. One should be filed if we want to track the regression formally.

## Summary by CodeRabbit

* **New Features**
  * Added "Up Arrow edits last message" setting toggle in Settings to quickly edit the most recent message with keyboard shortcut
  * Up Arrow now supports editing both user and assistant messages in Conversation and Roleplay modes
  * Setting is configurable and persists across sessions

[Review Change Stack](https://app.coderabbit.ai/change-stack/Pasta-Devs/Marinara-Engine/pull/703)

### DIFF
--- a/packages/client/src/components/chat/ChatArea.tsx
+++ b/packages/client/src/components/chat/ChatArea.tsx
@@ -134,6 +134,7 @@ export function ChatArea() {
   const guideGenerations = useUIStore((s) => s.guideGenerations);
   const intuitiveSwipeNavigation = useUIStore((s) => s.intuitiveSwipeNavigation);
   const intuitiveSwipeRerollLatest = useUIStore((s) => s.intuitiveSwipeRerollLatest);
+  const editLastMessageOnArrowUp = useUIStore((s) => s.editLastMessageOnArrowUp);
   const messagesEndRef = useRef<HTMLDivElement>(null);
   const scrollRef = useRef<HTMLDivElement>(null);
   const prevScrollHeightRef = useRef(0);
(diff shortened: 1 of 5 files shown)


LukaTheHero:
@SpicyMarinara  don't mention this in the update notes, I rushed the last one without proper testing when you were updating, I'm fixing my whoopsie

PR #749, opened 2026-05-12
refactor(folders): null-guard route handlers and atomicity for storage loops

### LINKED ISSUE #747: [Feature]: add explicit null-guard to storage.create/update in both folder route modules

The folder route modules (`packages/server/src/routes/connection-folders.routes.ts` and the parallel `packages/server/src/routes/chat-folders.routes.ts`) use a non-null assertion (`folder!.collapsed`) after `storage.create` / `storage.update` in their POST and PATCH handlers.

This is currently safe — both storage methods resolve via `getById(id)` immediately after the insert/update, so a `null` return would only happen if the underlying SQL silently failed (which would have thrown first). The PATCH path additionally has a `getById` precheck that guarantees the row exists.

That said, the non-null assertion is a fragile coupling: if the storage layer ever evolves to handle creation failures differently (e.g. returning `null` on a constraint violation rather than throwing), every handler using `folder!` would explode at runtime with a cryptic `Cannot read property 'collapsed' of null`.

## Proposed solution

Replace the non-null assertion with an explicit guard in both folder route modules, in lockstep:

```ts
// POST / (create)
const folder = await storage.create({ name: name.trim(), color });
if (!folder) return reply.status(500).send({ error: "Failed to create folder" });
return reply.send({ ...folder, collapsed: folder.collapsed === "true" });
```

```ts
// PATCH /:id (update)
const existing = await storage.getById(req.params.id);
if (!existing) return reply.status(404).send({ error: "Folder not found" });
const folder = await storage.update(req.params.id, req.body);
if (!folder) return reply.status(500).send({ error: "Failed to update folder" });
return reply.send({ ...folder, collapsed: folder.collapsed === "true" });
```

Both `connection-folders.routes.ts` and `chat-folders.routes.ts` should be updated in the same PR — diverging patterns between two analogous modules invites confusion for future contributors.

## Alternatives considered

- **Leave the non-null assertion** — works under current semantics but rots the moment storage error-handling evolves.
- **Wrap the entire handler in a try/catch and 500 on any throw** — broader change, would mask legitimate validation errors (400s) from being distinguished from internal failures.
- **Change `storage.create/update` to throw on missing post-write fetch** — semantically equivalent to the current behaviour; doesn't remove the type-level `T | null` problem at the route boundary.

The explicit `if (!folder)` guard is the smallest change that satisfies the type system without altering observable behaviour for any current code path.

## Additional context

Raised during CodeRabbit review of PR #745 (specifically [this thread](https://github.com/Pasta-Devs/Marinara-Engine/pull/745#discussion_r3226725893)). Filed as a lockstep follow-up so the connection-folders implementation in #745 stays consistent with the existing chat-folders pattern rather than diverging mid-feature.

## Template check

### LINKED ISSUE #748: [Feature]: wrap folder storage per-row UPDATE loops in db.transaction() across both folder modules

The folder storage modules (`packages/server/src/services/storage/chat-folders.storage.ts` and `packages/server/src/services/storage/connection-folders.storage.ts`) and the chat-folders reorder route (`packages/server/src/routes/chat-folders.routes.ts` — `POST /reorder-chats`) all contain per-row UPDATE loops that lack atomicity:

- `create()` — shifts every existing folder's `sortOrder` by +1, then inserts the new folder. If the insert fails partway through the shift, the DB is left with shifted rows but no new folder.
- `reorder(orderedIds)` — issues one UPDATE per ID to set the new `sortOrder`. If the loop crashes halfway, the folder list is left in a mixed state.
- `reorderConnections(orderedIds, folderId)` (connection-folders only) — same shape as `reorder`, but on `api_connections`. Already exposed by the route, not yet wired to UI (drag-within-folder is a deferred follow-up).
- `chat-folders.routes.ts` `POST /reorder-chats` — runs `chatsStorage.update(id, { sortOrder })` + `chatsStorage.setFolderForChat(id, folderId)` in a `for` loop without transaction boundaries.

Same pattern lives in both folder systems, so this is the natural moment to fix them in lockstep.

## Proposed solution

Wrap each of the above operations in `db.transaction(async (tx) => { ... })` so the per-row UPDATEs (or the shift + insert in `create()`) commit atomically. This is the canonical pattern in this codebase — see `lorebooks.storage.ts:231,285,295` and `agents.storage.ts:388`.

Applied to both folder modules in the same PR so the chat-folder and connection-folder storage stay consistent.

## Alternatives considered

- **Single-query bulk UPDATE with a SQL expression** (e.g. `tx.update(folders).set({ sortOrder: sql\`${folders.sortOrder} + 1\` })`). **Rejected — incompatible with the file-native storage backend.** Marinara v1.5.7+ defaults to `STORAGE_BACKEND=files` (see `packages/server/src/config/runtime-config.ts:245`), which routes all DB calls through `packages/server/src/db/file-backed-store.ts`. That backend treats `.set()` patches as plain row literals and `.run()` is an intentional no-op (file-backed-store.ts:932 — "Raw SQL is intentionally unsupported by the file-native runtime"). A Drizzle `sql` template would silently break for any user not on opt-in SQLite mode.

- **CASE-based bulk UPDATE via `db.run(sql.raw(...))`** — same blocker as above. The file-native runtime's `run()` is a no-op.

- **Leave the loops as-is** — status quo. Acceptable because folder counts in a single-user Marinara install are O(dozens), so the practical impact of the missing atomicity is small. But the consistency win (matching `lorebooks.storage.ts`'s transaction pattern) justifies the small change.

- **Chunked transactions to avoid the libSQL Windows crash** (see `chats.storage.ts:450-453`, issue #73). Not needed at folder/connection scale — the entire reorder fits in a single small transaction.

## Additional context

Raised during CodeRabbit review of PR #745 across three threads — [storage.ts:41 (create)](https://github.com/Pasta-Devs/Marinara-Engine/pull/745#discussion_r3226725943), [storage.ts:70 (reorder)](https://github.com/Pasta-Devs/Marinara-Engine/pull/745#discussion_r3226725959), [storage.ts:86 (reorderConnections)](https://github.com/Pasta-Devs/Marinara-Engine/pull/745#discussion_r3226725970).

Filed as a lockstep follow-up so both folder storage modules pick up the transaction wrap at the same time, rather than letting them diverge.

To be implemented together with the null-guard refactor in #747, since both touch the same files and target the same two folder modules.

## Template check

Romuromylus:
## Linked issue

Closes #747
Closes #748

## Why this change

Two follow-ups from CodeRabbit's review of [#745](https://github.com/Pasta-Devs/Marinara-Engine/pull/745), both of which I deferred at the time because applying them only to the new `connection-folders` files would have diverged from the existing `chat-folders` pattern. With #745 merged, this PR addresses both findings simultaneously in lockstep across both folder modules:

1. **#747** — POST/PATCH handlers in both folder route modules used `folder!.collapsed` non-null assertions. Safe under current semantics, but brittle to future error-handling changes.
2. **#748** — `create()`, `reorder()`, and `reorderConnections()` in the folder storage modules — plus `POST /api/chat-folders/reorder-chats` — issued per-row UPDATE loops without transaction boundaries. A partial failure mid-loop could leave `sort_order` / `folder_id` columns in a mixed state.

Folder counts in a single-user Marinara install are O(dozens), so the *performance* impact of the missing atomicity is negligible — but a transaction wrap gives us actual atomicity, matches the canonical pattern already used in `lorebooks.storage.ts` / `agents.storage.ts`, and costs almost nothing.

## What changed

**Routes — null-guard the create/update responses:**
- `packages/server/src/routes/chat-folders.routes.ts` — POST + PATCH: replace `folder!.collapsed` with an explicit `if (!folder) return reply.status(500)...` guard.
- `packages/server/src/routes/connection-folders.routes.ts` — same change.

**Storage — wrap per-row loops in `db.transaction()`:**
- `packages/server/src/services/storage/chat-folders.storage.ts` — `create()` (shift + insert) and `reorder()` (per-row sort_order update).
- `packages/server/src/services/storage/connection-folders.storage.ts` — `create()`, `reorder()`, and `reorderConnections()` (the route is wired but UI for drag-within-folder is still deferred — wrapping it now means the eventual UI flip is pure-frontend).
- `packages/server/src/services/storage/chats.storage.ts` — `update()` and `setFolderForChat()` now accept an optional `tx` parameter (typed structurally as `Pick<DB, "select" | "update">`, the same shape used by `syncLorebookLinks` in `lorebooks.storage.ts:154`). When `tx` is omitted the behaviour is identical to before.
- `packages/server/src/routes/chat-folders.routes.ts` — `POST /reorder-chats` now wraps the loop in `app.db.transaction()` and threads `tx` through both `chatsStorage.update` and `chatsStorage.setFolderForChat`, so the sortOrder write and the group-sibling folder propagation commit atomically.

Each transaction uses a single shared `now()` timestamp so all rows within one reorder share an `updated_at` value (intentional — the previous "later-timestamp-per-row" behaviour was an accidental side-effect of the per-row loop, not a deliberate semantic).

## Why not a single-query bulk UPDATE?

CodeRabbit's original suggestion was `tx.update(folders).set({ sortOrder: sql\`${folders.sortOrder} + 1\` })` to collapse the shift into one statement. That's a no-go on this codebase's default storage backend:

- Marinara v1.5.7+ defaults to `STORAGE_BACKEND=files` (see `packages/server/src/config/runtime-config.ts:245`).
- The file-native runtime in `packages/server/src/db/file-backed-store.ts:932` intentionally treats `run()` as a no-op and `.set()` patches as plain row literals — a Drizzle `sql` template silently breaks for any user not on opt-in `STORAGE_BACKEND=sqlite`.

The transaction wrap is portable across both backends (file-backed-store has its own `transaction` implementation at file-backed-store.ts:829 — a deep table snapshot with rollback-on-throw — and this PR's changes flow through it correctly).

## Validation

### Manual verification notes

Verified live against `me.romyromulus.com` after fork-deploy:

- **Atomic create (connection-folders)** — 5 existing folders pre-shift, created a new folder, confirmed via `GET /api/connection-folders` that the new folder is at `sortOrder=0` and the 5 originals shifted to `1..5` exactly.
- **Atomic create (chat-folders)** — same flow against `/api/chat-folders`, no existing folders, new folder inserted at `sortOrder=0`.
- **Atomic reorder (connection-folders)** — POSTed `{orderedIds: [...].reverse()}` to `/api/connection-folders/reorder`, confirmed full reverse applied (e.g. `LinkAPI LLM` from `5 → 0`, `Tx Test Folder` from `0 → 5`).
- **No console errors** under any of the above flows.
- **Cleanup** — test folders deleted, original 5 connection folders + their connections intact.

Things that would benefit from human verification before merge:

- Manually verify reorder-chats in the UI — drag chats between folders in CONVO/RP/GM tabs, confirm both the per-chat sortOrder and the group-sibling folder propagation still work end-to-end.
- Manually verify drag-to-reorder folders in the connections panel (the `Reorder.Group` integration).
- Manually verify the 500 path on the POST/PATCH null-guard — hard to trigger without inducing a real DB failure; safe to skip if the lower-risk happy paths are confirmed.

## Docs and release impact

- [x] No docs changes needed

No UI changes — pure backend/storage refactor. The connections panel and chat sidebar behaviour are identical to before from the user's point of view, just with stronger consistency guarantees under partial-failure scenarios.

## Release Notes

* **Bug Fixes**
  * Improved error handling for folder creation and update operations
  * Enhanced reliability of folder reordering to prevent inconsistent states during concurrent updates

[Review Change Stack](https://app.coderabbit.ai/change-stack/Pasta-Devs/Marinara-Engine/pull/749)

### DIFF
--- a/packages/server/src/routes/chat-folders.routes.ts
+++ b/packages/server/src/routes/chat-folders.routes.ts
@@ -31,7 +31,8 @@ export async function chatFoldersRoutes(app: FastifyInstance) {
       return reply.status(400).send({ error: "Invalid mode" });
     }
     const folder = await storage.create({ name: name.trim(), mode, color });
-    return reply.send({ ...folder, collapsed: folder!.collapsed === "true" });
+    if (!folder) return reply.status(500).send({ error: "Failed to create folder" });
+    return reply.send({ ...folder, collapsed: folder.collapsed === "true" });
   });
 
   // ── Update a folder ──
@@ -42,7 +43,8 @@ export async function chatFoldersRoutes(app: FastifyInstance) {
     const existing = await storage.getById(req.params.id);
     if (!existing) return reply.status(404).send({ error: "Folder not found" });
     const folder = await storage.update(req.params.id, req.body);
-    return reply.send({ ...folder, collapsed: folder!.collapsed === "true" });
+    if (!folder) return reply.status(500).send({ error: "Failed to update folder" });
+    return reply.send({ ...folder, collapsed: folder.collapsed === "true" });
   });
 
   // ── Delete a folder (chats are moved to root) ──
--- a/packages/server/src/routes/chat-folders.routes.ts
+++ b/packages/server/src/routes/chat-folders.routes.ts
@@ -90,13 +90,20 @@ export async function chatFoldersRoutes(app: FastifyInstance) {
   }>("/reorder-chats", async (req, reply) => {
     const { orderedChatIds, folderId } = req.body;
     if (!Array.isArray(orderedChatIds)) return reply.status(400).send({ error: "orderedChatIds must be an array" });
-    for (let i = 0; i < orderedChatIds.length; i++) {
-      const id = orderedChatIds[i]!;
-      // Update sortOrder on the visible representative chat, then propagate
-      // the folder assignment to its sibling branches.
-      await chatsStorage.update(id, { sortOrder: i + 1 });
-      await chatsStorage.setFolderForChat(id, folderId);
-    }
+    // Atomic: a partial failure mid-loop would leave chats with a mix of
+    // old and new sort_order / folder_id values across siblings of the
+    // same group. Chats-per-folder counts are O(dozens), well under the
+    // libSQL Windows transaction use-after-free threshold noted in
+    // chats.storage.ts:449.
+    await app.db.transaction(async (tx) => {
+      for (let i = 0; i < orderedChatIds.length; i++) {
+        const id = orderedChatIds[i]!;
+        // Update sortOrder on the visible representative chat, then propagate
+        // the folder assignment to its sibling branches.
+        await chatsStorage.update(id, { sortOrder: i + 1 }, { tx });
+        await chatsStorage.setFolderForChat(id, folderId, { tx });
+      }
+    });
     return reply.send({ ok: true });
   });
 }
--- a/packages/server/src/routes/connection-folders.routes.ts
+++ b/packages/server/src/routes/connection-folders.routes.ts
@@ -25,7 +25,8 @@ export async function connectionFoldersRoutes(app: FastifyInstance) {
     const { name, color } = req.body;
     if (!name?.trim()) return reply.status(400).send({ error: "Name is required" });
     const folder = await storage.create({ name: name.trim(), color });
-    return reply.send({ ...folder, collapsed: folder!.collapsed === "true" });
+    if (!folder) return reply.status(500).send({ error: "Failed to create folder" });
+    return reply.send({ ...folder, collapsed: folder.collapsed === "true" });
   });
 
   // ── Update a folder ──
@@ -36,7 +37,8 @@ export async function connectionFoldersRoutes(app: FastifyInstance) {
     const existing = await storage.getById(req.params.id);
     if (!existing) return reply.status(404).send({ error: "Folder not found" });
     const folder = await storage.update(req.params.id, req.body);
-    return reply.send({ ...folder, collapsed: folder!.collapsed === "true" });
+    if (!folder) return reply.status(500).send({ error: "Failed to update folder" });
+    return reply.send({ ...folder, collapsed: folder.collapsed === "true" });
   });
 
   // ── Delete a folder (connections are moved to root) ──
--- a/packages/server/src/services/storage/chat-folders.storage.ts
+++ b/packages/server/src/services/storage/chat-folders.storage.ts
@@ -20,23 +20,27 @@ export function createChatFoldersStorage(db: DB) {
     async create(input: { name: string; mode: string; color?: string }) {
       const id = newId();
       const timestamp = now();
-      // Shift existing folders down and place new folder at the top
-      const existing = await db.select().from(chatFolders);
-      for (const f of existing) {
-        await db
-          .update(chatFolders)
-          .set({ sortOrder: f.sortOrder + 1 })
-          .where(eq(chatFolders.id, f.id));
-      }
-      await db.insert(chatFolders).values({
-        id,
-        name: input.name,
-        mode: input.mode as "conversation" | "roleplay" | "visual_novel",
-        color: input.color ?? "",
-        sortOrder: 0,
-        collapsed: "false",
-        createdAt: timestamp,
-        updatedAt: timestamp,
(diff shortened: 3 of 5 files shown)


Romuromylus:
@coderabbitai full review

Romuromylus:
@coderabbitai review

PR #817, opened 2026-05-14
feat: add Fish Audio TTS provider option

ionstorm66:
## Linked issue

No issue to link.

-

## Why this change

Adds Fish Audio (fish.audio) as a new TTS provider option alongside the existing OpenAI, ElevenLabs, and PocketTTS sources.

-

## What changed

Three-layer change (shared schema → server → client):

| Layer | File | Change |
|---|---|---|
| Shared types | `packages/shared/src/types/tts.ts` | Added `"fishaudio"` to `ttsSourceSchema` enum |
| Server routes | `packages/server/src/routes/tts.routes.ts` | Added defaults, voice listing (`GET /v1/models`), `/speak` handler with Bearer auth + `model` header + JSON body |
| Client UI | `packages/client/src/components/panels/s

-

## Validation

### Manual verification notes

Works, but I have it set to only list your own created voices. Fish has like a few 100k voices, so a dropdown is pretty terrible for selecting. If you're using your own server, all voices will show as owned by you anyway.

-

## Docs and release impact

- [x] No docs changes needed


Xelvanis:
Would you mind redoing the PR form with the proper template? :>

ionstorm66 (commented):
<sub>Reviewed in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/Pasta-Devs/Marinara-Engine/pull/817)</sub>

Romuromylus:
Took a pass through this, nice add overall, fish audio fills an obvious gap. A few things worth tightening before it lands:

**PR template**
Like Xelvanis mentioned above, the description got truncated mid-sentence in the Client UI row, which is why `Pull Request template check` keeps failing on every triage run. Refilling the template should clear that.

**Voice-fetch policy is stricter than the rest of the file**
In `packages/server/src/routes/tts.routes.ts`, the new fishaudio voice-fetch hardcodes:
```ts
policy: {
  allowLocal: false,
  allowedProtocols: ["https:"],
  flagName: "TTS_LOCAL_URLS_ENABLED",
},
```
Every other provider's voices fetch (and the `/speak` path) uses `allowLocal: allowLocalTtsUrl(cfg)` and `["https:", "http:"]`. This contradicts your own validation note about self-hosted servers. `/speak` works fine against a self-hosted fish-compatible server, but voice listing silently falls back to empty because the policy blocks it. Aligning it with the openai voices fetch should fix that.

**Voice filter is too permissive**
```ts
.filter((item) => item.type === "tts" || !item.type)
```
The `|| !item.type` branch will surface voice clones / VC models that don't carry a `type` field, and they'll then 400 at synthesis time. Probably safer to require `item.type === "tts"` strictly.

**Minor**
- `{ ...openAiHeaders(cfg.apiKey), model: model }` → shorthand `model` (pretty sure `pnpm-validate` will flag this).
- With `self=true` on the voice listing and the `<select disabled={voiceOptions.length === 0}>` UI, a fish cloud user with no self-created voices ends up with a disabled empty picker and no escape hatch. A pockettts-style `<input list="...">` + `<datalist>` would let people type a voice id directly and still get autocomplete from whatever the API returns. Could be a follow-up.

Happy to re-review once those are in 🙏

SpicyMarinara:
Will close if OP doesn't get back to us.

PR #882, opened 2026-05-15
feat(lorebooks): ReDoS hardening for keyword regex activation

### LINKED ISSUE #846: [Feature]: ReDoS hardening for the lorebook activation/regex layer

Lorebook entries support author-supplied regex keywords (`entry.useRegex` + `entry.keys[]`). When `useRegex` is on, the matcher constructs `new RegExp(keyword, flags)` and runs `.test(text)` against chat context (or, after #845, against keyword-test sample text in the editor). A pathological pattern like `(a+)+$` against `"a".repeat(N) + "!"` triggers catastrophic backtracking and stalls the thread that runs it.

The trust model is self-contained today: the author writes the regex, the author runs it against their own chat in their own browser/server. So this is fundamentally a self-inflicted-DoS / footgun rather than a security vulnerability — but it's still worth tracking, because:

1. The footgun is invisible until activation time. The keyword-test panel added in #845 (and #816) makes the failure mode surface earlier, but only if the user happens to trigger the pattern. A near-bad pattern can stay in a lorebook for months and only stall the chat under the right input.
2. Importing lorebooks from third parties — SillyTavern World Info, shared lorebook bundles — does cross the trust boundary. An imported entry whose author isn't the current user can carry a malicious regex, and the importing user has no easy way to spot it.
3. The flagged code path runs in both the server (`packages/server/src/services/lorebook/keyword-scanner.ts`) and now the client (`packages/client/src/components/lorebooks/LorebookEditor.tsx` preview, via the shared helper at `packages/shared/src/utils/lorebook-keyword-matching.ts`). A server-side stall affects all live chats on that instance; a client-side stall jams the editor tab.

## Proposed solution

Add a regex-safety layer at the activation boundary — not at the leaf `testKeyword` helper — so it applies uniformly across server scanning, client preview, and any future caller. Options (in roughly increasing weight):

1. **Pre-validate with a complexity heuristic** (e.g. `safe-regex` or a hand-rolled depth check). Reject or fall back to literal substring matching for patterns that fail the check. Cheap to add, false-positives possible.
2. **Per-call execution timeout** wrapper around `regex.test`. Implementable in the browser via `setTimeout` + a worker; on the server via `worker_threads` with a timer. Catches patterns that pass static checks but explode on specific inputs.
3. **Non-backtracking engine.** RE2 via `re2-wasm` on both client and server. Strongest, but adds ~300KB WASM to the client bundle and changes semantics (no backreferences, no lookaround). Probably too aggressive for this codebase's user expectations.

A combination of #1 (cheap pre-filter) + #2 (timeout safety net) is probably the right place to land — keeps the JS regex semantics users already know, but breaks the worst-case stall.

## Alternatives considered

- **Do nothing.** Match the existing posture of `packages/shared/src/utils/regex-replacement.ts`, which has the same "author runs their own regex" trust model and no hardening today. Defensible given the trust model, but leaves the import-from-third-party case unaddressed.
- **Add hardening at the leaf `testKeyword` helper.** Rejected during the #845 review — the leaf doesn't know the call context, and applying it there would force every caller to absorb the cost / behavior change. The activation layer (`scanForActivatedEntries` and the editor preview's `previewMatches` memo) is the better seam.
- **Warn on suspicious patterns in the editor instead of enforcing.** Would address #1 (early warning) without changing matcher semantics. Could ship alongside enforcement — they're not exclusive.

## Additional context

- Surfaced during CodeRabbit's review of #845 ([thread](https://github.com/Pasta-Devs/Marinara-Engine/pull/845#discussion_r3244734888)). The leaf-helper hardening was declined for that PR; the discussion concluded a dedicated activation-layer effort was the right scope.
- Existing call sites today: `packages/server/src/services/lorebook/keyword-scanner.ts:testPrimaryKeys` (via the shared helper after #845 merges), and `packages/client/src/components/lorebooks/LorebookEditor.tsx`'s `previewMatches` memo.
- Worth checking whether `packages/shared/src/utils/regex-replacement.ts` should be folded into the same hardening pass — same posture, separate code path.
- Not urgent. Filing as a low-priority hardening item.

## Template check

Romuromylus:
## Linked issue

Closes #846

## Why this change

Lorebook entries support author-supplied regex keys (`entry.useRegex` + `entry.keys[]`). When `useRegex` is on, the matcher does `new RegExp(keyword, flags).test(text)` against chat context — and after #845, against keyword-test sample text in the editor preview. A pathological pattern like `(a+)+` against `"a".repeat(N) + "!"` triggers catastrophic backtracking and stalls the thread that runs it: server-side that affects every live chat on the instance, client-side that jams the editor tab.

Trust model is mostly self-contained today (you write the regex, you run it on your own chat), so it's footgun more than vulnerability — but the SillyTavern lorebook import path crosses that boundary, and the 2026-04 issue here is that the failure mode is invisible until the right input shows up. I want a guard at the activation seam that applies uniformly across server scanning, client preview, and any future caller, without changing semantics for safe patterns.

## What changed

Two-layer defense, both gated at the activation boundary instead of the leaf `testKeyword` helper (per the #845 review thread that scoped this out of the leaf):

**1. Static safety heuristic in `@marinara-engine/shared` (`isPatternSafe`)** — `packages/shared/src/utils/regex-safety.ts`
- Star-height check rejects nested quantifiers like `(a+)+`, `(a*)*`, `(a+|b)+`, `((ab)+)+`
- Pattern length cap (1000 chars)
- `{n,m}` repetition cap (100, also catches unbounded `{1,}`)
- Walks the source once, tracks character classes / escapes / group prefixes (`(?:`, `(?=`, `(?<name>`) without depending on `safe-regex`/`ret`. Keeps `@marinara-engine/shared` at zero new deps.

Wired into `testKeyword`: when `useRegex` is on and `isPatternSafe` returns false, the helper falls back to literal substring matching — same posture as the existing invalid-regex `catch` fallback, principle of least surprise.

**2. Server-side execution timeout (`vmRegexExecutor`)** — `packages/server/src/services/lorebook/regex-timeout.ts`
- Runs `regex.test` inside `vm.runInNewContext` with a 50 ms timeout. V8's interrupt checks fire during regex execution, so catastrophic backtracking aborts instead of pinning the event loop.
- On timeout: logs a warn with the pattern source + flags + text length and returns `false`. **Doesn't** substitute literal-substring semantics on timeout — the pattern compiled and may have matched on simpler input; switching meanings mid-run would be a silent surprise.
- Injected via the new optional `regexExecutor` on `KeywordMatchOptions` so the shared helper stays runtime-agnostic; the client preview keeps using the default executor.

Client-side has the static check only. A browser-side sync timeout would require re-architecting the helper around Web Workers (sync `regex.test` on the main thread has no clean interrupt seam), which is a heavier change than the symptom warrants — a frozen tab is recoverable, a frozen Node event loop isn't.

`packages/shared/src/utils/regex-replacement.ts` is intentionally out of scope. Same posture, different feature surface (regex scripts on display/output, separate call sites). Filing a follow-up so this one stays focused on the lorebook activation layer.

## Validation

### Manual verification notes

- 8 new tests in `packages/server/test/lorebook-regex-safety.test.ts` cover both layers — `isPatternSafe` for safe / nested-quantifier / oversized / unbalanced patterns, `testKeyword` fallback semantics, and the canonical `(a|aa)+` × `"a".repeat(28) + "!"` ReDoS pair under the vm timeout (verifies the abort actually fires within ~1 s on the runner).
- Manually verify in browser on a real lorebook with `useRegex: true`: paste a pathological pattern like `(a+)+` and confirm the keyword-test panel from #845 stays responsive — without this change, that combo freezes the editor tab.
- Manually verify safe regex keys still match — e.g. `\bword\b`, `foo|bar`, `(?:foo){1,5}`. The static check accepts these; behavior should be unchanged.
- Manually verify the literal-substring fallback by setting an unsafe pattern as a key and pasting text containing that pattern's source verbatim — entry should still activate (matching the existing invalid-regex fallback).
- Manually check server logs for the new warn line on a contrived pathological pattern that survives the static check (e.g. add one via direct DB insert) — should see `Lorebook regex /…/… exceeded 50ms timeout against chat context (length=…) — entry will not match this scan`.

## Docs and release impact

- [x] No docs changes needed

No UI changes — purely runtime hardening behind the existing `useRegex` flag and the keyword-test panel from #845.

## Summary by CodeRabbit

* **New Features**
  * Added regex safety validation and timeout protection (50ms default) for keyword matching to prevent pattern hangs; unsafe patterns automatically fall back to literal substring matching.

* **Tests**
  * Added comprehensive test suite for regex safety validation and timeout behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Pasta-Devs/Marinara-Engine/pull/882)

### DIFF
--- a/packages/server/src/services/lorebook/keyword-scanner.ts
+++ b/packages/server/src/services/lorebook/keyword-scanner.ts
@@ -12,6 +12,7 @@ import type {
   LorebookSchedule,
 } from "@marinara-engine/shared";
 import { testPrimaryKeys, testSecondaryKeys } from "@marinara-engine/shared";
+import { vmRegexExecutor } from "./regex-timeout.js";
 
 /** Compute cosine similarity between two vectors. Returns 0 for empty/mismatched vectors. */
 function cosineSimilarity(a: number[], b: number[]): number {
@@ -463,6 +464,7 @@ export function scanForActivatedEntries(
       useRegex: entry.useRegex,
       matchWholeWords: entry.matchWholeWords,
       caseSensitive: entry.caseSensitive,
+      regexExecutor: vmRegexExecutor,
     };
 
     // Test primary keys
--- a/packages/server/src/services/lorebook/regex-timeout.ts
+++ b/packages/server/src/services/lorebook/regex-timeout.ts
@@ -1,0 +1,55 @@
+// Server-side execution timeout for compiled regex tests against chat context.
+//
+// The shared static check (isPatternSafe) catches the common ReDoS shapes before
+// compilation, but expert-crafted patterns can still pass that check and explode
+// on specific input. This wrapper runs the regex.test call inside a fresh V8 vm
+// context with a hard timeout — the engine inserts interrupt checks during
+// regex execution, so catastrophic backtracking aborts instead of stalling the
+// event loop indefinitely.
+//
+// On timeout: log a warning with the pattern source so the lorebook author can
+// see why an entry stopped activating, and return false. We deliberately do NOT
+// fall back to literal substring on timeout — the pattern compiled and may have
+// matched on simpler input; substituting literal-substring semantics here would
+// cause silent surprise matches.
+
+import * as vm from "node:vm";
+import { logger } from "../../lib/logger.js";
+
+/** Default per-call timeout for a single regex.test against chat context, in ms. */
+export const DEFAULT_REGEX_TIMEOUT_MS = 50;
+
+/** Build a regex executor that runs `regex.test(text)` under a vm timeout. */
+export function createTimeoutRegexExecutor(timeoutMs: number = DEFAULT_REGEX_TIMEOUT_MS) {
+  return function vmRegexExecutor(regex: RegExp, text: string): boolean {
+    // The vm context only needs the regex + text; we recompile inside the vm so
+    // the interrupt check is wired through the new isolate's regex execution.
+    const context = vm.createContext({ __pattern: regex.source, __flags: regex.flags, __text: text });
+    try {
+      const result = vm.runInContext("(new RegExp(__pattern, __flags)).test(__text)", context, {
+        timeout: timeoutMs,
+        displayErrors: false,
+      });
+      return Boolean(result);
+    } catch (err) {
+      const message = err instanceof Error ? err.message : String(err);
+      // V8 surfaces timeouts as "Script execution timed out." — log warn and skip the entry.
+      if (message.includes("timed out")) {
+        logger.warn(
+          "Lorebook regex /%s/%s exceeded %dms timeout against chat context (length=%d) — entry will not match this scan",
+          regex.source,
+          regex.flags,
+          timeoutMs,
+          text.length,
+        );
+        return false;
+      }
+      // Anything else (rare — invalid regex would have thrown earlier) — let testKeyword's
+      // outer catch handle it via its literal-substring fallback by re-throwing.
+      throw err;
+    }
+  };
+}
+
+/** Default executor used by keyword-scanner.ts. */
+export const vmRegexExecutor = createTimeoutRegexExecutor();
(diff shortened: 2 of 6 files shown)


Romuromylus:
@coderabbitai review

PR #970, opened 2026-05-18
feat(image-gen): add Draw Things HTTP API support

etherious1804:
## Linked issue

Closes #969

## Why this change

- Marinara already aliases the `drawthings` provider to the AUTOMATIC1111 dispatch case, but Draw Things' HTTP server diverges from A1111 enough that the path silently fails today: requests sending `cfg_scale` return `HTTP 200 {"images": []}` (full generation cycle, no image), filling the connection editor's Model field causes `HTTP 422` because Draw Things rejects `override_settings`, and the sampler dropdown only overlaps on five names with Draw Things' actual sampler set. Out-of-the-box Draw Things connections look configured but generate nothing.
- Local-network image generation on Apple Silicon Macs is meaningfully faster via Draw Things (Metal + Apple Neural Engine + CoreML) than the ComfyUI Python path. Fixing this unlocks a working backend for that audience without adding a new provider abstraction — it's all behavior changes inside the existing `automatic1111` dispatch.

## What changed

- `packages/server/src/services/image/image-generation.ts`:
  - Passes the explicit `serviceHint` through to `generateAutomatic1111` and branches on `serviceHint === "drawthings"`.
  - Draw Things branch uses `guidance_scale` (not `cfg_scale`) and `batch_count` (not `n_iter`/`batch_size`); drops `override_settings`, `scheduler`, `restore_faces` which Draw Things 422s on.
  - Lazy-loads `sharp` (already a server dep, same load pattern as `sprites.routes.ts`) and resizes img2img init images to match request `width`/`height` exactly, since Draw Things does not auto-resize like A1111/Forge does. Falls back to the original image if `sharp` is unavailable.
  - Adds the missing `import { logger }` referenced at line 1430 — pre-existing bug that prevented the file from compiling.
- `packages/shared/src/constants/image-generation-defaults.ts`:
  - Extends `SD_WEBUI_SAMPLER_OPTIONS` with 16 Draw Things-specific samplers (`UniPC AYS`, `DPM++ SDE Substep/Trailing/AYS`, `TCD`, `TCD Trailing`, `LCM`, `PLMS`, etc.) labeled `(Draw Things)` so AUTOMATIC1111/Forge users can avoid them.
- `README.md`, `docs/FAQ.md`, `docs/CONFIGURATION.md`: adds Draw Things to the supported-image-providers list and to the SSRF-allowed local image backends section.
- Existing AUTOMATIC1111 and Forge connections see zero behavior change — the new branch only activates when the explicit hint is `drawthings`.

[...]

PR #1027, opened 2026-05-20
fix(image-gen): harden OpenRouter image response parsing

### LINKED ISSUE #1026: [Bug]: OpenRouter image generation throws unhelpful errors when response is not the expected chat envelope
## Summary

`generateOpenRouter` in `packages/server/src/services/image/image-generation.ts` assumes every successful response from OpenRouter's image-generation endpoint is a chat-completions JSON envelope with an extractable image URL. In practice there are at least two shapes that break the path today:

## Failure modes

1. **Raw image bytes.** Some OpenRouter image models (raw provider passthrough) return PNG / JPEG / GIF / WEBP bytes directly with an `image/*` content-type instead of a JSON envelope. The current `await resp.json()` call then throws `SyntaxError: Unexpected token ...` and the user sees a JSON parse crash instead of the image. There is no way to recover this in user space — the response is being thrown out before any image data is read.

2. **JSON envelope with no image (model refusal, text-only response, etc.).** The current "no image" error message is just `No image data in OpenRouter response: <first 200 chars>` with no `finish_reason`, no shape hint, and no log line. There is no way to tell from this error whether the issue is a wrong API key, a model that doesn't support image output, a prompt the model refused, or a misconfigured `modalities` setting on the model itself.

3. **Gzipped responses.** OpenRouter sometimes returns `Content-Encoding: gzip` from upstream providers. The shared `imageFetch` helper does not set `decodeCompressedResponse: true`, so the buffer fails the same JSON-parse path.

## Proposed fix

- Read the response into a `Buffer`, sniff the content-type + magic bytes, and short-circuit to a base64 image return when the bytes look like an image.
- Otherwise parse as JSON; if JSON parse fails, surface the content-type and first 80 bytes (hex) in the error message so the user can self-diagnose.
- If the JSON has a top-level `error` field, surface it directly.
- If JSON parses but no image URL is extractable, log the full message shape via `logger.warn` (model, `finish_reason`, first 800 chars of the message body) and include `finish_reason` + message-object keys in the thrown error.
- Enable `decodeCompressedResponse: true` on the shared `imageFetch` helper so gzipped responses are decoded transparently.

## Repro

1. Configure an OpenRouter image-generation connection.
2. Use any model that doesn't return a chat-completions JSON envelope (e.g., a raw-passthrough image model).
3. Run a generation. Observe the crash / unhelpful error.

PR coming.

etherious1804:
## Linked issue

Closes #1026

## Why this change

OpenRouter's image-generation endpoint surfaces several response shapes depending on the underlying provider and the model's \`modalities\` config, and Marinara's current parser only handles the chat-completions JSON envelope cleanly. Two real failure modes today:

1. **Raw image bytes.** Some OpenRouter image models (raw provider passthrough) return PNG / JPEG / GIF / WEBP bytes directly with an \`image/*\` content-type instead of a JSON envelope. The current \`await resp.json()\` throws \`SyntaxError: Unexpected token ... is not valid JSON\` and the user sees a JSON-parse crash instead of an image.
2. **JSON envelope with no image (model refusal, text-only response, etc.).** The existing \"no image\" error message is just \`No image data in OpenRouter response: <first 200 chars>\` with no \`finish_reason\`, no shape hint, and no log line. There is no way to tell from this error whether the issue is a wrong API key, a model that doesn't support image output, a prompt that was refused, or a misconfigured \`modalities\` setting.
3. **Gzipped responses.** OpenRouter sometimes returns \`Content-Encoding: gzip\` from upstream providers. The shared \`imageFetch\` helper does not set \`decodeCompressedResponse: true\`, so those buffers fail the JSON-parse path even when the underlying payload is a valid envelope.

This PR is the split-out half of #970 that the reviewer asked to move into its own change.

## What changed

\`packages/server/src/services/image/image-generation.ts\`:

- \`generateOpenRouter\`: read the response into a \`Buffer\` first, then sniff the content-type and magic bytes. If it looks like an image, base64-encode it and return directly via the existing \`detectImageMimeType\` / \`imageExtensionFromMimeType\` helpers.
- Otherwise parse as JSON. If the parse fails, throw an error that includes the content-type and the first 80 bytes (hex) so the user can immediately tell what shape they actually received.
- If the JSON has a top-level \`error\` field, surface it verbatim (truncated to 400 chars).
- If JSON parses but no image URL is extractable, log the full message shape via \`logger.warn\` (model, \`finish_reason\`, first 800 chars of the message body) and throw an error that includes \`finish_reason\` and the message-object keys.
- Enable \`decodeCompressedResponse: true\` on the shared \`imageFetch\` helper so gzipped responses are decoded transparently. The other image-related callers already work on decoded buffers, so this is a no-op for them.
- Add the missing \`import { logger }\` referenced by the new \`logger.warn\` call.

No behavior change for OpenRouter responses that were already producing images successfully — the new path only branches on the response shape.

## Validation

- [x] \`pnpm check\` passes locally
- [ ] Read and followed \`CONTRIBUTING.md\`

> Only ticking \`pnpm check\` because it is a deterministic command result that this agent ran on the final commit. The manual-verification boxes are intentionally left unchecked per the AI-Generated PR guidance — those are a to-do list for the human contributor and reviewer, not evidence that work is done. The work I actually exercised is described below as prose.

### What I ran locally (prose, for reviewer reference)

- \`pnpm check\` green on the final commit (\`tsc -b\` + ESLint across shared/server/client; server tsc-noemit; client build).
- Triggered the raw-image-bytes path by pointing the OpenRouter connection at a model that returns a direct PNG response. Without the patch the call crashed with a JSON SyntaxError; with the patch the image was returned and displayed in chat.
- Triggered the \"JSON envelope with no extractable image\" path by sending a refusal-likely prompt; the error message now reports \`finish_reason\` and the message-object keys, and a \`logger.warn\` line is written with the full shape.

**Manual verification a reviewer should still do** (these are paths I could not easily fixture locally):

- Verify a gzipped OpenRouter response still parses correctly end-to-end — I enabled \`decodeCompressedResponse: true\` but did not have a gzipped fixture handy. Worth running one OpenRouter call through a debug proxy that forces \`Content-Encoding: gzip\`.
- Sanity-check that no other image-flow caller (Stability AI, Pollinations, custom OpenAI-compatible, NovelAI, ComfyUI, AUTOMATIC1111, etc.) regresses with \`decodeCompressedResponse: true\` on the shared helper. Those callers already work on decoded buffers, but the flag is shared, so worth a smoke test in CI.

## Docs and release impact

- [ ] No docs changes needed

Purely a server-side parser hardening. No docs or release-metadata changes.

No UI surface touched.

## Note on relationship to #970

#970 (Draw Things) was originally a single combined PR with these OpenRouter parsing improvements bundled in. Per the reviewer's call to split unrelated work, those OpenRouter changes have been pulled out into this PR; #970 now only contains Draw Things support.

### DIFF
--- a/packages/server/src/services/image/image-generation.ts
+++ b/packages/server/src/services/image/image-generation.ts
@@ -24,6 +24,7 @@ import {
 } from "@marinara-engine/shared";
 import { isImageLocalUrlsEnabled } from "../../config/runtime-config.js";
 import { generateRunPodComfyUI } from "./runpod-comfyui.service.js";
+import { logger } from "../../lib/logger.js";
 import { normalizeLoopbackUrl, safeFetch, validateOutboundUrl } from "../../utils/security.js";
 
 const GALLERY_DIR = join(DATA_DIR, "gallery");
@@ -221,6 +222,7 @@ function imageFetch(url: string | URL, init?: RequestInit, options: { allowLocal
       flagName: "IMAGE_LOCAL_URLS_ENABLED",
     },
     maxResponseBytes: MAX_IMAGE_RESPONSE_BYTES,
+    decodeCompressedResponse: true,
   });
 }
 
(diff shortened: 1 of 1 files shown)


cha1latte:
Looks good, thank you.

PR #1132, opened 2026-05-24
feat(tts): pass model to provider voices endpoint

### LINKED ISSUE #1131: TTS voice list: pass the configured model to OpenAI-compatible provider voices endpoint
### Problem

When the TTS source is `openai` but the `baseUrl` points at a local or third-party OpenAI-compatible TTS server (e.g. [mlx-audio](https://github.com/Blaizzy/mlx-audio), [openedai-speech](https://github.com/matatonic/openedai-speech), LocalAI), the voice dropdown silently falls back to OpenAI's hardcoded 9 voices instead of showing what the configured server actually supports.

### Root cause

`fetchProviderVoices` in `packages/server/src/routes/tts.routes.ts` already fetches `${base}/audio/voices` for the openai source and falls back gracefully on 404. But it does **not** pass the configured model as a query parameter.

Multi-model OAI-compatible TTS servers cannot know which voice catalog to return without that hint. mlx-audio in particular can host Kokoro (54 voices), Higgs Audio (3 sample voices + reference cloning), Chatterbox, etc. side-by-side — each with very different catalogs. Without `?model=`, those servers either 400 the request or return the wrong list.

The result: even if your local server has the right endpoint, Marinara won't show the voices, so per-character voice mode (`voiceMode: "per-character"`) is effectively unusable against a local TTS server — users can only assign characters to OpenAI's 9 cloud voices.

### Repro

1. Configure TTS with `source: openai`, `baseUrl: http://127.0.0.1:8081/v1`, `model: mlx-community/Kokoro-82M-bf16`.
2. Run an mlx-audio server with that model serving `/v1/audio/voices?model=<id>`.
3. `GET /api/tts/voices` returns `{ fromProvider: false, voices: [alloy, ash, coral, ...] }` — the OpenAI fallback, not the 54 Kokoro voices.

### Proposed fix

Include `cfg.model` as a `?model=` query param when fetching `${base}/audio/voices`. Vanilla `api.openai.com` ignores unknown query params, so the change is safe for the default OpenAI endpoint. Local/multi-model servers can now disambiguate.

The fix is 4 lines, no behavior change for OpenAI-hosted users.

### PR

I'll open a draft PR with the change linked to this issue.

etherious1804:
## Linked issue

Closes #1131

## Why this change

When the TTS source is `openai` but the `baseUrl` points at a local or third-party OpenAI-compatible TTS server (mlx-audio, openedai-speech, LocalAI, etc.), Marinara's voice dropdown silently falls back to OpenAI's hardcoded 9 voices instead of showing what the configured server actually supports.

Root cause: `fetchProviderVoices` already calls `${base}/audio/voices` and falls back gracefully on 404 — but it doesn't pass the configured model as a query param. Multi-model OAI-compatible TTS servers can't return the right voice catalog without that hint, so they 400 or return the wrong list, and Marinara falls back to the hardcoded OpenAI nine.

Concrete impact: per-character voice mode (`voiceMode: "per-character"`) is effectively unusable against a local TTS server today — users can only assign characters to the 9 cloud voices Marinara hardcodes for the openai source.

## What changed

`packages/server/src/routes/tts.routes.ts` — 8 lines added in `fetchProviderVoices`:

- Build the voices URL with `?model=${cfg.model}` when a model is configured.
- Vanilla `api.openai.com` ignores unknown query params, so default OpenAI users see zero behavior change.
- Local/multi-model servers can now disambiguate which model's catalog to return.

No new dependencies. No client changes — the existing voice fetch path picks up the richer list automatically.

## Validation

### Manual verification notes

Verified end-to-end against a local mlx-audio server (Kokoro 82M) at `http://127.0.0.1:8081/v1`:

- `pnpm check` ran cleanly (TS + ESLint + build, exit code 0). The only WARN lines are the pre-existing `Unsupported engine` notes for node 26 vs `>=24 <26` — unrelated to this change.
- Before the patch: `GET /api/tts/voices` returned `{ fromProvider: false, voices: [alloy, ash, coral, echo, fable, nova, onyx, sage, shimmer] }` — the OpenAI hardcoded fallback.
- After the patch: same call returns `{ fromProvider: true, voices: <54 Kokoro voice ids> }` — the 28 English voices (`af_*`, `am_*`, `bf_*`, `bm_*`) plus the 26 non-English ones (`if_*`, `zf_*`, etc.).
- `POST /api/tts/speak` with `voice: "bm_george"` (a voice that wasn't previously visible) returned a valid mp3 and played correctly through afplay.
- Reviewer should: configure TTS against a real OAI-compatible local TTS server (or set `model` to a non-existent value against api.openai.com — the request still goes through, OpenAI ignores the extra param). UI dropdown should populate from the provider rather than fall back. Per-character voice assignment should now show the full provider catalog.

## Docs and release impact

- [ ] No docs changes needed

No docs need touching — the existing behavior is documented as "Marinara queries the provider for its voice list" and that's still what happens. This makes the existing query actually work for multi-model providers.

No UI code changed. The existing voice picker (`packages/client/src/components/panels/settings/TTSConfigCard.tsx`) reads from `/api/tts/voices` and renders whatever it gets — so the dropdown automatically shows the new richer list. No screenshots attached because the visible change is just "more entries in the dropdown."

### DIFF
--- a/packages/server/src/routes/tts.routes.ts
+++ b/packages/server/src/routes/tts.routes.ts
@@ -495,7 +495,14 @@ async function fetchProviderVoices(cfg: TTSConfig): Promise<TTSVoicesResponse> {
     return voices.length > 0 ? responseFromVoiceOptions(cfg.source, voices, true) : fallbackVoices(cfg.source);
   }
 
-  const res = await safeFetch(`${base}/audio/voices`, {
+  // Include the configured model as a query param so OpenAI-compatible servers
+  // that host multiple TTS models (e.g. mlx-audio) can return the right voice
+  // catalog. Vanilla OpenAI ignores extra query params.
+  const voicesUrl = new URL(`${base}/audio/voices`);
+  const modelHint = cfg.model?.trim();
+  if (modelHint) voicesUrl.searchParams.set("model", modelHint);
+
+  const res = await safeFetch(voicesUrl, {
     headers: openAiHeaders(cfg.apiKey),
     signal: AbortSignal.timeout(10_000),
     policy: {


cha1latte:
merged in #1218

PR #1139, opened 2026-05-24
feat(tts): stream audio per sentence as the model generates

### LINKED ISSUE #1138: TTS: stream audio per sentence as the model generates (drop time-to-first-audio)
### Problem

When auto-play is enabled, Marinara dispatches the entire model reply to TTS *only after* `isStreaming` flips to false — i.e. after the full reply has been generated AND any post-processing agents have run. For local TTS providers (mlx-audio, openedai-speech, etc.) where synthesis is much slower than cloud APIs, this means the user waits the *full* reply time before hearing the first word. With a 30-second roleplay reply and a local Chatterbox model, time-to-first-audio is 30+ seconds.

The information needed to start TTS earlier is already available in the chat store: `streamBuffers` is updated token-by-token as the LLM streams. We can split on sentence boundaries and dispatch each completed sentence to `/api/tts/speak` immediately, playing the resulting clips sequentially as they arrive.

### Proposed feature

Opt-in toggle under TTS settings: **"Stream as the model generates"**. When enabled (with auto-play for the chat mode also enabled):

- Each completed sentence in the streaming buffer fires a TTS request in parallel.
- Resulting audio blobs play in order through a chained HTMLAudioElement playback queue.
- Time-to-first-audio drops to ~1-2 seconds (time for the model to emit the first sentence + one TTS roundtrip).
- The end-of-stream "speak the whole reply" path is suppressed when the toggle is on, so the reply isn't spoken twice.
- Default: off (no behavior change for existing users).

### UX edge cases

- **User clicks Stop mid-reply** — audio must cut off immediately, not finish the queued sentences after the LLM aborts.
- **User switches chats mid-stream** — the streaming session for the old chat tears down; the new chat starts fresh.
- **User joins a chat that's already streaming** — prime the chunker cursor to the current buffer length so the backlog isn't spoken as one giant chunk.
- **Inline thinking blocks** (`<thought>`, `<|channel>thought`) — recognized and skipped so the model's reasoning isn't spoken aloud.
- **Server `content_replace` events** — defensive cursor snap so a shortened buffer doesn't desync the chunker.
- **Per-message play buttons** must keep working — streaming TTS uses raw HTMLAudioElement rather than calling `ttsService.stop()`, preserving the global service for other code paths.
- **Speed slider** — `playbackRate` is applied client-side on each audio element so the slider works regardless of whether the local TTS server respects the `speed` API parameter.

### Implementation

Open PR follows.

etherious1804:
> **Heads up — minor conflict with #1141:** Both PRs add an import to `packages/client/src/components/chat/ChatArea.tsx` from `../../lib/tts-dialogue`. This one adds `useStreamingTTS` and leaves the tts-dialogue import alone; #1141 expands the tts-dialogue import to include `clientSidePlaybackRate`. Resolution is one line — keep both imports. Happy to rebase whichever lands second.

## Linked issue

Closes #1138

## Why this change

When auto-play is on, Marinara waits for `isStreaming` to flip to false before dispatching the entire reply to TTS. That flag only flips after the full LLM response AND post-processing agents have completed. For local TTS providers (mlx-audio with Chatterbox / Higgs / etc.), where synthesis isn't as fast as cloud APIs, this means users wait the full 5-30s reply time before hearing any audio. The information to start earlier is already in the chat store — `streamBuffers` updates token-by-token. We can chunk on sentence boundaries and dispatch each completed sentence to `/api/tts/speak` immediately.

## What changed

- `packages/shared/src/types/tts.ts` — new `autoplayStreaming: boolean` field on `ttsConfigSchema`, default `false`. Existing users are unaffected by virtue of the default and Zod's schema-fills-missing-fields behavior.
- `packages/client/src/lib/sentence-chunker.ts` (new) — pure-logic chunker that, given a growing buffer, returns each newly-completed sentence on every call. Handles inline thinking-block patterns (`<think>`, `<thinking>`, `<thought>`, `<|think|>`, `<|channel>thought`) so models that emit reasoning inline aren't spoken. Defensive against `content_replace` SSE events (server can shorten the buffer mid- or end-of-stream).
- `packages/client/src/hooks/use-streaming-tts.ts` (new) — React hook that subscribes to the chat store and runs a streaming session for the active chat. Each completed sentence kicks a TTS fetch immediately; the resulting blobs play sequentially through chained HTMLAudioElement instances. Key design points:
  - **Independent of the global `ttsService` singleton.** The hook tracks its own audios and never calls `ttsService.stop()`, so per-message "speak this message" buttons elsewhere keep working unaltered.
  - **Stop button cancels audio instantly.** Listener attached directly to the chat's `AbortSignal` at session start — teardown fires immediately instead of waiting for the `isStreaming → false` transition (which can lag several hundred ms behind the abort).
  - **No AbortSignal listener leak.** Each audio attaches a single `{once: true}` listener with explicit `removeEventListener` on the natural-end path. No accumulation across long replies.
  - **Joins mid-stream gracefully.** Primes the chunker cursor to current buffer length on session start so switching into a chat that's already streaming doesn't speak the backlog as one chunk.
  - **Speed slider works.** `audio.playbackRate = cfg.speed` is applied client-side, so the slider works uniformly regardless of whether the local TTS server respects the `speed` API parameter.
- `packages/client/src/components/panels/settings/TTSConfigCard.tsx` — exposes the new toggle under the existing Auto-play group.
- `packages/client/src/components/chat/ChatArea.tsx` — invokes the hook gated on `cfg.enabled && cfg.autoplayStreaming && mode-specific autoplay flag`. When streaming TTS is active, the existing end-of-stream autoplay early-returns to avoid double-speaking the reply.

No server-side changes. Existing autoplay path is untouched.

## Validation

### Manual verification notes

Tested against a local mlx-audio server running Chatterbox Turbo with a cloned-voice registry entry (`http://127.0.0.1:8081/v1`):

- **`pnpm check`** ran cleanly end-to-end (TS + ESLint + production build). The only WARN lines are pre-existing `Unsupported engine` notices for node 26 vs `>=24 <26` — unrelated.
- **Time-to-first-audio dropped from ~15-20s to ~1.5-2s** for typical roleplay replies. The first audio chunk plays while the model is still generating the second/third sentence.
- **Stop button**: tested cancelling mid-reply during sentences 3 and 8 of a 12-sentence reply. Audio cuts off immediately on click; subsequent queued sentences do not play. The chat's `AbortSignal` listener path verified by setting `isStreaming` flips to false 200-400ms after Stop on this machine, and audio still stops immediately, confirming the abort listener fires earlier than the isStreaming-based path.
- **Speed slider**: changing the slider (1.0 → 1.5 → 0.8) on the next reply produces audibly faster/slower playback without pitch shift (browser's playbackRate default preserves pitch). Verified with both default Chatterbox voice and a cloned-voice registry entry.
- **Per-message play button**: clicked the "speak this message" button on a previous message while streaming TTS was actively playing. The previous behavior was preserved — that button uses `ttsService.speakSequence` and now coexists with the streaming session. (Two audios overlapping is an existing-tool limitation, not a regression; flagged as a follow-up if Marinara wants to enforce single-active-playback across both code paths.)
- **Inline thinking blocks**: tested with `enable_thinking=true` (model emits `<thought>...</thought>` inline). The chunker waited for the closing tag and never dispatched the thinking content to TTS. Once the close arrived, the actual reply was spoken correctly.
- **End-of-stream autoplay suppression**: with the new toggle on, the legacy end-of-stream autoplay early-returns at the new `if (cfg.autoplayStreaming) return;` guard. With the toggle off, behavior is identical to before this PR (verified by toggling and observing).

Reviewer should: enable the toggle in TTS settings, ensure mode-specific autoplay is on (e.g. `autoplayRP`), then send a roleplay message and listen for the first audio within ~2s. Try Stop mid-reply. Try the speed slider. Try a per-message play button.

## Docs and release impact

- [ ] No docs changes needed

No documentation updated. The new toggle has self-explanatory label + sits next to the existing autoplay toggles. If maintainers want a doc note for the FAQ ("Why does TTS feel slow with local providers? Try the streaming toggle"), I can add it in a follow-up.

No new UI surface beyond a single ToggleRow consistent with the surrounding rows (autoplayRP / autoplayConvo / autoplayGame). No screenshots; the visual diff is one new row labeled "Stream as the model generates" under Auto-play.

### DIFF
--- a/packages/client/src/components/chat/ChatArea.tsx
+++ b/packages/client/src/components/chat/ChatArea.tsx
@@ -62,6 +62,7 @@ import { useEncounterStore } from "../../stores/encounter.store";
 import { useTranslationStore } from "../../stores/translation.store";
 import { ttsService } from "../../lib/tts-service";
 import { useTTSConfig } from "../../hooks/use-tts";
+import { useStreamingTTS } from "../../hooks/use-streaming-tts";
 import { buildTTSVoiceRequests, normalizeTTSCharacterName, withTTSVoiceRequestCacheKeys } from "../../lib/tts-dialogue";
 import { mirrorSpritePlacements, normalizeSpritePlacements } from "./sprite-placement";
 import { normalizeSpriteDisplayModes } from "./sprite-display-modes";
@@ -1412,6 +1413,26 @@ export function ChatArea() {
     },
     [characterMap],
   );
+
+  // Streaming TTS — dispatch per sentence as the model generates so the
+  // first audio arrives in 1-2s instead of waiting for the full reply.
+  // The hook is a no-op unless cfg.enabled && cfg.autoplayStreaming, and
+  // only runs while THIS chat is the actively-streaming one.
+  const streamingTTSEnabled =
+    !!ttsConfig?.enabled &&
+    !!ttsConfig?.autoplayStreaming &&
+    (chatMode === "roleplay" || chatMode === "visual_novel"
+      ? !!ttsConfig.autoplayRP
+      : chatMode === "game"
+        ? false
+        : !!ttsConfig.autoplayConvo);
+  useStreamingTTS({
+    enabled: streamingTTSEnabled,
+    chatId: activeChatId,
+    ttsConfig,
+    resolveCharacterIdForSpeaker: resolveTTSCharacterId,
+  });
+
   useEffect(() => {
     const wasStreaming = prevIsStreamingRef.current;
     prevIsStreamingRef.current = isStreaming;
@@ -1419,6 +1440,11 @@ export function ChatArea() {
 
     const cfg = ttsConfigRef.current;
     if (!cfg?.enabled) return;
+
+    // When the per-sentence streaming-TTS path is active, the audio for the
+    // reply has already played (or is still playing). Skip the end-of-stream
+    // autoplay so we don't speak the same content twice.
+    if (cfg.autoplayStreaming) return;
 
     const mode = chatModeRef.current;
     const shouldAutoplay =
--- a/packages/client/src/components/panels/settings/TTSConfigCard.tsx
+++ b/packages/client/src/components/panels/settings/TTSConfigCard.tsx
@@ -305,6 +305,7 @@ export function TTSConfigCard() {
   const [autoplayRP, setAutoplayRP] = useState(false);
   const [autoplayConvo, setAutoplayConvo] = useState(false);
   const [autoplayGame, setAutoplayGame] = useState(false);
+  const [autoplayStreaming, setAutoplayStreaming] = useState(false);
   const [dialogueOnly, setDialogueOnly] = useState(false);
   const [audioFormat, setAudioFormat] = useState<TTSAudioFormat>("mp3");
 
@@ -349,6 +350,7 @@ export function TTSConfigCard() {
     setAutoplayRP(savedConfig.autoplayRP);
     setAutoplayConvo(savedConfig.autoplayConvo);
     setAutoplayGame(savedConfig.autoplayGame);
+    setAutoplayStreaming(savedConfig.autoplayStreaming ?? false);
     setDialogueOnly(savedConfig.dialogueOnly ?? false);
     setAudioFormat(savedConfig.audioFormat ?? "mp3");
     setSaveStatus("idle");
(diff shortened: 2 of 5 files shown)


etherious1804:
Withdrawing — keeping iteration local. Will reopen once the streaming-TTS path is polished (currently fixing regen-doesn't-stream and narrator-voice-on-demand-mode issues).

PR #1141, opened 2026-05-24
fix(tts): apply playback rate client-side for local OAI-compatible servers

etherious1804:
## Linked issue

Closes #1140

## Why this change

The TTS settings panel exposes a Speed slider that's supposed to control playback rate, but for local OpenAI-compatible TTS servers (mlx-audio, openedai-speech, etc.) the `speed` parameter in `/v1/audio/speech` is an OpenAI extension that most local implementations silently ignore. The user moves the slider and nothing changes — the synthesized audio comes back at exactly the same rate every time. Cloud OpenAI honors the parameter; local servers don't.

We can fix this universally without re-synthesis by applying `playbackRate` on the HTMLAudioElement client-side. Pitch is preserved by default. The slider becomes uniformly effective across every provider, with no extra server round-trip.

## What changed

**Shared types**

- `packages/shared/src/types/tts.ts`: no new type fields; existing `speed` on `TTSConfig` is unchanged. The value continues to be sent to providers that respect it server-side.

**Client — client-side playback rate**

- `packages/client/src/lib/tts-dialogue.ts`: new exported `clientSidePlaybackRate(config)` helper that returns the configured speed (or 1 when undefined/unset). Single source of truth for the "speed-as-playbackRate" interpretation.
- `packages/client/src/lib/tts-service.ts`: `speakSequence` accepts a `playbackRate` option and applies it to each `HTMLAudioElement` (`audio.playbackRate = rate`) immediately after constructing the element, before `play()`.
- `packages/client/src/components/chat/ChatArea.tsx` and `packages/client/src/components/chat/ChatMessage.tsx`: pass `clientSidePlaybackRate(ttsConfig)` into the existing `speakSequence` call sites (autoplay and per-message manual speak).
- `packages/client/src/components/panels/settings/TTSConfigCard.tsx`: existing Speed slider is untouched UI-wise — it simply works now.

The change is intentionally minimal: no surgery in the request builders, no provider-specific branches. The `speed` request param continues to be sent (for providers that honor it — pitch may also be adjusted server-side), and the client-side `playbackRate` layer is purely additive.

## Validation

- [x] `pnpm check` passes locally

- [ ] Container (Docker / Podman) built and ran without issue
- [ ] Ran the app, clicked through the changes manually
- [ ] Checked edge cases (light + dark mode, mobile viewport, empty states, error paths)
- [ ] Above manual verification completed (describe below)
- [ ] Read and followed `CONTRIBUTING.md`

### Manual verification notes

[...]

PR #1142, opened 2026-05-24
feat(tts): stream audio per sentence as the model generates

### LINKED ISSUE #1138: TTS: stream audio per sentence as the model generates (drop time-to-first-audio)

etherious1804:
## Linked issue

Closes #1138

## Why this change

When auto-play is enabled, Marinara today waits for `isStreaming` to flip false before dispatching the entire reply to TTS. For local TTS providers (mlx-audio Chatterbox, openedai-speech, etc.) where synthesis is much slower than cloud APIs, this means the user waits the *full* generate time before hearing any audio. With a 30-second roleplay reply on a local M-series box, time-to-first-audio is 30+ seconds — the assistant has finished writing before it even starts speaking.

The information needed to start TTS much earlier is already available in the chat store: `streamBuffers` updates token-by-token as the LLM streams. We can split on sentence boundaries and dispatch each completed sentence as it arrives, then play the resulting clips sequentially. Time-to-first-audio drops to ~1-2 seconds (the time for the model to produce the first sentence plus one TTS round-trip).

## What changed

**Client — new streaming-TTS hook**

- `packages/client/src/hooks/use-streaming-tts.ts`: new `useStreamingTTS` hook subscribes to the chat store, runs a sentence chunker against the growing stream buffer, and dispatches `ttsService.generateAudio` per completed sentence. Audio fetches run in parallel; playback is serialized through a chained promise so order is preserved. Aborts cleanly on user-cancel (the existing chat-abort signal is forwarded, plus a local AbortController fires on `stopSession`). Per-audio AbortListener is bounded to that element's playback lifetime so long replies don't pile up 50+ listeners on a single signal.
- The hook exports a module-level `stopStreamingTTS()` / `isStreamingTTSActive()` / `subscribeStreamingTTSActive()` trio so any UI surface can interrupt or observe the active session — the per-message speaker icon hooks into this.

**Client — sentence chunker**

- `packages/client/src/lib/sentence-chunker.ts`: new module. Maintains a `cursor` + `highWaterMark` over the growing buffer; `extractNewSentences` emits each completed sentence as soon as a sentence-ending punctuation + whitespace appears. Multi-dot ellipses (`..`, `...`) are explicitly NOT treated as sentence ends (continuations, not terminators). Known abbreviations (`Mr.`, `Dr.`, etc.) are skipped. The high-water mark prevents replay if the server later emits a `content_replace` SSE event that rewinds the buffer to a common prefix (used to strip character commands at end-of-stream).

**Client — stop affordance on speaker icons**

- `packages/client/src/components/chat/ChatMessage.tsx`: every speaker icon subscribes to the streaming-TTS active state. While a session is active, the icon flips to `VolumeX`, shows "Stop streaming narration" as tooltip, and clicking any icon halts playback. The chat's generation-abort signal only fires *while* the LLM is streaming; without this affordance there's no way to interrupt audio after natural completion.

**Wiring**

- `packages/client/src/components/chat/ChatArea.tsx`: invokes `useStreamingTTS` when the chat is in a mode that supports autoplay and `ttsConfig.autoplayStreaming` is on. The legacy "speak the whole reply at end-of-stream" path is gated on `if (cfg.autoplayStreaming) return;` so the reply isn't spoken twice.
- `packages/client/src/components/panels/settings/TTSConfigCard.tsx`: adds the **"Stream as the model generates"** toggle under TTS settings (default off — no behavior change for existing users).
- `packages/shared/src/types/tts.ts`: adds `autoplayStreaming?: boolean` to `TTSConfig`.

## Known limitation

The speaker icon does **not** visually flip back to "play" after natural stream completion — it stays in stop-state until a new stream starts. Click-to-stop continues to function (clicking a stuck stop-icon is a safe no-op since there's nothing to stop). Will track in a follow-up; not blocking.

## Dependency

Includes #1141 (client-side playback rate) as a base commit so the streaming-TTS path can honor the speed slider too. Once #1141 merges, this PR's diff will shrink to just the streaming TTS changes.

## Validation

### Manual verification notes

> Generated PR — unchecked boxes are left for the merger to confirm manually per the repo's AI-PR guidance.

**What was verified on the branch (against a running Marinara dev server, local mlx-audio TTS, local Gemma 4 26B LLM, Roleplay mode with autoplay + streaming-TTS toggle on):**

- Streaming TTS fires per sentence during a long reply — first audio audible within ~1-2 seconds of model start.
- Pressing any speaker icon mid-stream halts audio (the chained playback queue stops at the current element).
- Regenerating a response correctly starts a fresh streaming-TTS session for the new reply (verified after fixing the cursor initialization race against `setStreaming(true) → clearStreamBuffer()`).
- Ellipses like `"But tell me, human..."` don't cause mid-sentence cuts; the chunker waits for a real terminator.
- Local Gemma's end-of-stream `content_replace` (strips character commands) no longer replays the cleaned tail — the high-water mark guards against re-emission.

**To verify in browser:**

- Enable "Stream as the model generates" in TTS settings, autoplay on, Roleplay mode.
- Send a message and confirm audio starts before the model finishes writing.
- Mid-stream, click the speaker icon on the streaming message — audio stops.
- Disable the toggle — behavior reverts to the existing "speak after stream ends" path.

## Docs and release impact

- [x] No docs changes needed

No new UI surface beyond the toggle in TTS settings. The speaker-icon swap (Volume2 ↔ VolumeX + sky-blue tint while streaming) is in-place, no new components.

### DIFF
--- a/packages/client/src/components/chat/ChatArea.tsx
+++ b/packages/client/src/components/chat/ChatArea.tsx
@@ -62,6 +62,7 @@ import { useEncounterStore } from "../../stores/encounter.store";
 import { useTranslationStore } from "../../stores/translation.store";
 import { ttsService } from "../../lib/tts-service";
 import { useTTSConfig } from "../../hooks/use-tts";
+import { useStreamingTTS } from "../../hooks/use-streaming-tts";
 import {
   buildTTSVoiceRequests,
   clientSidePlaybackRate,
@@ -1417,6 +1418,26 @@ export function ChatArea() {
     },
     [characterMap],
   );
+
+  // Streaming TTS — dispatch per sentence as the model generates so the
+  // first audio arrives in 1-2s instead of waiting for the full reply.
+  // The hook is a no-op unless cfg.enabled && cfg.autoplayStreaming, and
+  // only runs while THIS chat is the actively-streaming one.
+  const streamingTTSEnabled =
+    !!ttsConfig?.enabled &&
+    !!ttsConfig?.autoplayStreaming &&
+    (chatMode === "roleplay" || chatMode === "visual_novel"
+      ? !!ttsConfig.autoplayRP
+      : chatMode === "game"
+        ? false
+        : !!ttsConfig.autoplayConvo);
+  useStreamingTTS({
+    enabled: streamingTTSEnabled,
+    chatId: activeChatId,
+    ttsConfig,
+    resolveCharacterIdForSpeaker: resolveTTSCharacterId,
+  });
+
   useEffect(() => {
     const wasStreaming = prevIsStreamingRef.current;
     prevIsStreamingRef.current = isStreaming;
@@ -1424,6 +1445,11 @@ export function ChatArea() {
 
     const cfg = ttsConfigRef.current;
     if (!cfg?.enabled) return;
+
+    // When the per-sentence streaming-TTS path is active, the audio for the
+    // reply has already played (or is still playing). Skip the end-of-stream
+    // autoplay so we don't speak the same content twice.
+    if (cfg.autoplayStreaming) return;
 
     const mode = chatModeRef.current;
     const shouldAutoplay =
--- a/packages/client/src/components/chat/ChatArea.tsx
+++ b/packages/client/src/components/chat/ChatArea.tsx
@@ -62,7 +62,12 @@ import { useEncounterStore } from "../../stores/encounter.store";
 import { useTranslationStore } from "../../stores/translation.store";
 import { ttsService } from "../../lib/tts-service";
 import { useTTSConfig } from "../../hooks/use-tts";
-import { buildTTSVoiceRequests, normalizeTTSCharacterName, withTTSVoiceRequestCacheKeys } from "../../lib/tts-dialogue";
+import {
+  buildTTSVoiceRequests,
+  clientSidePlaybackRate,
+  normalizeTTSCharacterName,
+  withTTSVoiceRequestCacheKeys,
+} from "../../lib/tts-dialogue";
 import { mirrorSpritePlacements, normalizeSpritePlacements } from "./sprite-placement";
 import { normalizeSpriteDisplayModes } from "./sprite-display-modes";
 import type {
@@ -1451,7 +1456,9 @@ export function ChatArea() {
     );
     if (ttsRequests.length === 0) return;
 
-    void ttsService.speakSequence(withTTSVoiceRequestCacheKeys(ttsRequests, cfg, lastMsg.id), lastMsg.id);
+    void ttsService.speakSequence(withTTSVoiceRequestCacheKeys(ttsRequests, cfg, lastMsg.id), lastMsg.id, {
+      playbackRate: clientSidePlaybackRate(cfg),
+    });
   }, [characterMap, isStreaming, resolveTTSCharacterId]);
 
   const newestMsgId = msgData?.pages[0]?.[msgData.pages[0].length - 1]?.id;
--- a/packages/client/src/components/chat/ChatMessage.tsx
+++ b/packages/client/src/components/chat/ChatMessage.tsx
@@ -48,7 +48,12 @@ import { useTranslate } from "../../hooks/use-translate";
 import { api } from "../../lib/api-client";
 import { ttsService } from "../../lib/tts-service";
 import { useTTSConfig } from "../../hooks/use-tts";
-import { buildTTSVoiceRequests, normalizeTTSCharacterName, withTTSVoiceRequestCacheKeys } from "../../lib/tts-dialogue";
+import {
+  buildTTSVoiceRequests,
+  clientSidePlaybackRate,
+  normalizeTTSCharacterName,
+  withTTSVoiceRequestCacheKeys,
+} from "../../lib/tts-dialogue";
 import { DIALOGUE_QUOTE_PATTERN_SOURCE, HTML_SAFE_DIALOGUE_QUOTE_PATTERN_SOURCE } from "../../lib/dialogue-quotes";
 import DOMPurify from "dompurify";
 import type { CharacterMap, ExpressionAvatarResolver, MessageSelectionToggle, PersonaInfo } from "./chat-area.types";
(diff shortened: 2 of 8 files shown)


etherious1804:
## Edge cases preserved in the current commit — checklist for a maintainer port

Cross-referencing if/when this gets ported to the new `src/...` structure (similar to how #1178 ported #1141). These are the non-obvious bugs we hunted during local testing that would silently regress if the port reimplements from the original spec.

### 1. Regenerate fires streaming TTS for the new reply

Trace: `setStreaming(true)` fires **before** `clearStreamBuffer()` for the new generation, so the buffer at rising-edge still holds the previous reply. If the hook anchors `highWaterMark` to `buffer.length` at session start, emission on the next reply is blocked until the new reply exceeds the old reply's length (often: never, for shorter replies).

Fix: at session start, set `cursor = buffer.length` (skip backlog), but **leave `highWaterMark = 0`** — let it grow naturally as emission happens. The defensive `buffer.length < cursor` snap handles the post-`clearStreamBuffer()` case.

### 2. Streaming keeps working after the first stream ends

Trace: chat.store sets `isStreaming = false` AND `streamingChatId = null` in the **same** update. If the hook early-returns on `streamingChatId !== chatId` before updating its `isStreamingRef`, the ref gets stuck at `true`. Next rising edge: `wasStreaming = true (stale), isStreaming = true (new)` → no rising edge detected → session never starts. Symptom in testing: works after refresh, breaks after the 1st-2nd reply, speaker icon flips back to normal but no audio plays on the next send.

Fix: update the streaming-state ref unconditionally **before** any early returns.

### 3. `content_replace` doesn't replay already-spoken content

Trace: Marinara's `content_replace` SSE event rewinds the streamBuffer to a common prefix and re-types the cleaned tail via the typewriter (e.g. when stripping character commands). Without a guard, the chunker re-extracts sentences from the rewind point.

Fix: chunker tracks `highWaterMark` = furthest offset ever emitted. `Math.max(cursor, highWaterMark)` is the scan start. Cursor can snap back after a shrink; hwm can't.

### 4. Ellipses don't trip mid-sentence

`"But tell me, human..."` would otherwise emit at the `.` before the next word arrives. The chunker rejects matches of `^\.{2,}["'""")\]}]?$` — `..`, `...`, `....` and variants. A bare `.` `!` `?` still terminates.

### 5. Thinking blocks anywhere in the reply (not just leading)

The original spec only skipped `<think>/<thought>` blocks at the start. Models emitting `<thought type="cot">...</thought>` (tag attributes), or mid-reply blocks, would leak to TTS before the server's end-of-stream strip.

Fix: `THINKING_BLOCK_RE = /<(think|thinking|thought)\b[^>]*>[\s\S]*?<\/\1>/gi` — `[^>]*` admits attributes, `\1` backreferences the correct close. Plus `findUnclosedThinkingStart()` so emission pauses at any open tag, and `stripClosedThinkingBlocks()` so emitted text never carries them.

### 6. Stop-icon as stop affordance

The chat's generation-abort path only fires WHILE the LLM is streaming — after natural completion, queued audio keeps playing with no way to interrupt. Module-level `stopStreamingTTS()` + pub/sub for active state lets every speaker icon double as a stop button while a session is active.

### 7. Known cosmetic limitation (called out in PR description)

Speaker icon doesn't visually flip back to "play" after **natural** stream completion — it stays in stop-state until a new stream starts. Click-to-stop is still functional (no-op if nothing to stop). Traced to the audio element's `onended` not firing reliably in some scenarios; the chained deactivation `.then()` covers the common case but not all of them. Worth tracking but not blocking.

### Repro patterns useful for verifying any port

- Send → wait for completion → send again → audio should fire on the 2nd reply (regen bug)
- Use a model that emits `<thought>...</thought>` mid-reply → thinking content must not reach TTS
- Reply containing `"He said. She replied. I agreed."` (multi-sentence dialogue in one quote) → must emit as one chunk in the character voice, not three split chunks (depends on #1143's chunker layer, included there)
- Mid-stream click any speaker icon → audio halts cleanly

cha1latte:
taken care of in #1221

PR #1178, opened 2026-05-24
Port TTS speed fallback to refactor

cha1latte:
## What?
Ports the local OpenAI-compatible TTS speed fallback from #1141 onto the refactor branch layout.

This keeps OpenAI cloud and non-OpenAI providers at the provider-owned speed behavior, while local OpenAI-compatible TTS servers apply the configured speed slider through the browser audio element.

## Why?
#1141 targets staging, but the same fix needs to land on refactor. A straight retarget conflicts because the TTS call sites moved in the refactor branch.

## How?
- Added `clientSidePlaybackRate()` in the shared TTS dialogue helpers.
- Applied non-default playback rates inside `ttsService.speak()`.
- Passed the computed playback rate from chat autoplay, per-message speak, and settings preview.

## Testing?
- `pnpm check` passed.
- Scratch Vitest proof passed: 1 file, 4 tests.
- Bunny pass completed for the focused refactor diff.

## Screenshots
N/A, non-visual TTS playback behavior.

## Anything Else?
Source PR: #1141.

Manual verification to consider: test the speed slider against a real local OpenAI-compatible TTS server, since Codex verified the client-side playback path but did not play audio through a real local TTS provider.

### DIFF
--- a/src/features/modes/shared/chat-ui/components/ChatMessage.tsx
+++ b/src/features/modes/shared/chat-ui/components/ChatMessage.tsx
@@ -45,7 +45,11 @@ import { useTranslate } from "../../../../../shared/hooks/use-translate";
 import { storageApi } from "../../../../../shared/api/storage-api";
 import { ttsService } from "../../../../../shared/lib/tts-service";
 import { useTTSConfig } from "../../../../../shared/hooks/use-tts";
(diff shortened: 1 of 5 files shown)


cha1latte:
@coderabbitai review

PR #1187, opened 2026-05-25
Guard NaN and unit-test the legacy text-normalization helpers

Romuromylus:
## Why this change

Follow-up to #1181, which introduced `normalize_legacy_text_array_fields` and `normalize_legacy_text_bool_fields` as the final barrier between malformed legacy TEXT cells and the editor. CodeRabbit flagged in review that the bool helper treats `f64::NAN` as truthy (`NaN != 0.0`) and that neither helper has a direct unit-test matrix - both are unreachable / quick-win nits, addressed together here so the contract is locked in before the next legacy field is added.

## What changed

- `src-tauri/src/commands/storage/shared.rs`: explicit `is_nan()` short-circuit in `normalize_legacy_text_bool_fields`'s number branch so `f64::NAN` returns `false` instead of `true`. Defensive only - `serde_json::Number::from_f64` already refuses to construct `Value::Number(NaN)`, so this is unreachable from any deserialized JSON input. The guard makes the function correct against direct `Value` construction so the next refactor doesn't have to think about it.
- `src-tauri/src/commands/storage/shared.rs`: 18 new unit tests covering the normalization contract for both helpers. Array helper: already-array preserved, parseable JSON-string array parsed, unparseable string -> `[]`, non-array JSON-string -> `[]`, scalar / null / missing-key / non-object-record edge cases. Bool helper: already-bool preserved, `"true"`/`"false"` coerced, the truthy/falsy text aliases (`1`/`0`/`yes`/`no`/`on`/`off`, case-insensitive) coerced, unknown string defaults to `false`, integer and float zero / non-zero coerced, null defaults to `false`, missing-key / non-object-record edge cases. Matches the existing `#[cfg(test)] mod tests` style already in the file.

The NaN test itself is deliberately omitted - `serde_json::Value::from(f64::NAN)` returns `Value::Null` rather than a NaN number, so there's no public API path to exercise the branch. The guard remains as defensive coding for the same reason CodeRabbit flagged it.

## Verification

- `cargo test --manifest-path src-tauri/Cargo.toml` - all 50 existing + 18 new tests pass, including the full `normalize_legacy_text_*` matrix.
- `cargo clippy --manifest-path src-tauri/Cargo.toml --workspace --all-targets -- -D warnings` - clean.
- `pnpm check` (architecture + frontend + docs) - clean.

Manually re-imported the pre-refactor profile snapshot from #1181's test data on `pnpm tauri dev` to confirm the import path still works end-to-end (no behavior change from the helper edits; the test matrix locks down the existing contract).

Out of scope: a separately-investigated post-import bug where imported chats fire agents (confirmed via runtime logs - `resolveAgents` resolves and `runPost` executes the expected 4 trackers for a roleplay imported chat) but the tracker panel and agent activity panel stay empty. Will file as a separate issue with the captured evidence so it can be diagnosed properly without entangling this PR.

## Linked

Follow-up to #1181.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed NaN value handling in legacy field normalization to ensure proper data processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Pasta-Devs/Marinara-Engine/pull/1187?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Romuromylus:
@coderabbitai review

PR #1191, opened 2026-05-25
Strip avatarPath from agent gameState serialization

### LINKED ISSUE #1188: Imported chats: agents run and yield results but tracker panel + agent activity stay empty
## Symptom

After importing a pre-refactor profile zip via Settings -> Import Profile, opening an imported roleplay chat (with `chat.metadata.activeAgentIds: ["world-state", "quest", "character-tracker", "persona-stats"]`) and sending a new message: the model reply arrives normally, but the tracker panel stays on "No tracker data yet" and the agent activity panel stays on "No agent activity yet". Same chat, same provider, same model on a freshly-created refactor chat does populate both panels.

## What I confirmed

I instrumented `src/engine/generation/start-generation.ts` and `src/engine/generation/agent-runner.ts` with `console.log` traces and sent a message in an imported roleplay chat. From the DevTools console:

```
[DBG startGeneration] {chatId: 'iYaXSHsPlEOPYuknQTWhW', chatMode: 'roleplay', chatMetaType: 'string', hasRequestMessages: false, impersonateBlockAgents: undefined}
[DBG resolveAgents] {chatId: 'iYaXSHsPlEOPYuknQTWhW', chatMetaType: 'string', enableAgents: true, activeAgentIds: Array(4), chatAgentsEnabled: true, scopedSize: 4, agentTypesInput: 'undefined'}
[DBG resolveAgents] agent rows 18 samples: [{...}, {...}, {...}]
[DBG resolveAgents] resolved rows: 4 types: ['world-state', 'quest', 'character-tracker', 'persona-stats']
[DBG start-generation] post-gen results {parallel: 0, post: 4, events: 4, sample: Array(3)}
```

So for this imported chat:

- `chatAgentsEnabled` returns `true` (chat metadata parses correctly from its TEXT form).
- `resolveAgents` resolves the expected 4 trackers.
- `runtime.runPost(content)` returns 4 post-gen results.
- 4 `agent_result` events are pushed into `agentEvents` and yielded from `startGeneration`.

Despite that, the agent activity panel (`RoleplayHUDActionsMenu` -> `useAgentStore.thoughtBubbles`) and the tracker panel both remain on their empty-state messages. So the gap is downstream of the agent-result yield - somewhere in `applyAgentResultEffects`, the per-result handlers (`applyTrackerResultToGameState`, `agentStore.addThoughtBubble`, etc.), the tracker snapshot persistence (`persistTrackerSnapshotSafely`), or the panel's React Query subscription - not in agent resolution or execution.

## Likely areas to inspect

Just hypotheses, didn't dig further:

- `src/features/runtime/generation/hooks/use-generate.ts:660` `applyAgentResultEffects` - the per-result handlers may be silently dropping the imported chat's results (e.g. `result.success === false` for an unrelated reason, or the result shape differs).
- `src/engine/generation/tracker-snapshots.ts` `persistTrackerSnapshotSafely` / `commitTrackerSnapshotForTarget` - the imported chat may have a `gameState` shape that the tracker writer can't update against.
- The agent activity panel may filter on `enabledAgentTypes` derived from chat-meta in a way that drops imported-chat thought bubbles before they render, despite the underlying `useAgentStore.thoughtBubbles` being populated.

## Repro

1. Build a pre-refactor profile zip from the test snapshot.
2. `pnpm tauri dev`, fresh data dir.
3. Settings -> Import Profile -> select the zip.
4. Open the imported roleplay chat.
5. Send any message.
6. Observe: model replies, but tracker and agent activity panels remain empty.

## Out of scope

I originally suspected `chat.metadata.activeAgentIds` was empty post-import and that no agents fire at all. After live verification that's incorrect - they fire and yield results. The actual bug is downstream of agent yield. The defensive-improvement PR I already filed (#1187: NaN guard + helper unit tests) is unrelated to this and does not fix this; this bug stays open.

If anyone is starting on this, the debug traces in `agent-runner.ts:578` (`resolveAgents`) and `start-generation.ts:606` (post-gen results) are the fastest way to confirm where in the pipeline the divergence happens. Worth noting that the same 4 agents on a fresh chat populate both panels correctly, so the divergence is data-shape-dependent on the imported chat row, not on the agents themselves.

Romuromylus:
## Why this change

Imported pre-refactor chats can carry full base64 image data URLs in `gameState.presentCharacters[].avatarPath` - the legacy backup format inlines avatars rather than referencing asset paths. Three sites in the agents-runtime layer serialize `gameState` (or a tracker summary built from it) into agent prompts, each of which can pick up a megabyte+ of base64 per present character.

On 1M-token providers (e.g. Gemini) the agent prompt blew past the ceiling and every tracker agent failed with:

```
Provider returned HTTP 400 Bad Request: prompt is too long: 2188040 tokens > 1000000 maximum
```

`applyAgentResultEffects` in `src/features/runtime/generation/hooks/use-generate.ts` short-circuits on `!result.success`, so the failures never reached the bubble / tracker handlers and the user-visible symptom was both panels stuck on their empty states ("No tracker data yet" and "No agent activity yet") with no error indication, even though the main chat reply itself succeeded.

## What changed

New named helper `stripAvatarPathsReplacer` at `src/engine/agents-runtime/strip-avatar-paths.ts` - a `JSON.stringify` replacer that drops any `avatarPath` key from the serialized tree. `avatarPath` is a UI display field (image path or data URL the tracker panel renders); the model never reads it. Every other model-relevant field on each present character (`name`, `mood`, `appearance`, `outfit`, `stats`, `thoughts`, `customFields`, etc.) is preserved as before.

The helper is now applied at all three sites that serialize `gameState` or a tracker summary built from it into a prompt:

1. `buildAgentExtras` `<current_game_state>` block at `src/engine/agents-runtime/executor/agent-executor.ts` - the original site identified in #1188.
2. `buildAgentMessages` `<committed_tracker_state>` block at `src/engine/agents-runtime/executor/agent-executor.ts` - attaches a tracker summary to the last 3 assistant messages and assigns `gs.presentCharacters` whole; structured-output agents would inherit the avatarPath data URLs from history regardless of whether `<current_game_state>` is included.
3. `buildKnowledgeRouterQuery` at `src/engine/agents-runtime/knowledge/knowledge-router.ts` - serializes `context.gameState` verbatim into the router query; latent until the knowledge router is enabled on an imported chat.

Co-locating the rule in a named helper makes it findable by symbol search and keeps future serialization sites a one-import-and-pass-the-replacer change rather than rediscovering the rule per-site.

## Verification

- `pnpm check:frontend`, `pnpm check:architecture`, `pnpm check:docs` all pass locally (595 modules, 2120 dependencies cruised).
- Manually verified in `pnpm tauri dev` on Windows against an imported pre-refactor profile (the exact repro from the issue: roleplay chat with `activeAgentIds: ["world-state", "quest", "character-tracker", "persona-stats"]`). Before the fix: all four agent results returned `success: false` with the prompt-too-long error and both side panels stayed empty. After the fix: all four agents complete normally, the agent activity panel shows the four thought bubbles, and the tracker panel populates with world / character / quest / persona data. Re-verified after the helper extraction; behavior unchanged.

## Out of scope

A second, separate UX bug is visible in passing: when every agent in a turn fails, the user sees only the empty-state panels - the failure surface (`setFailedAgentFailures` / `failedAgentTypes` in `agent.store.ts`) is defined but never called from `applyAgentResultEffects`, so genuine agent failures are silent. That made this bug much harder to spot than it should have been. Filed as #1192 to track separately; intentionally left out of this PR to keep the diff focused on the actual root cause.

Fixes #1188.

## Bug Fixes
- Optimized agent prompt serialization to exclude avatar path data, reducing unnecessary data transmission and improving performance when using legacy imported chats.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Pasta-Devs/Marinara-Engine/pull/1191?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/engine/agents-runtime/executor/agent-executor.ts
+++ b/src/engine/agents-runtime/executor/agent-executor.ts
@@ -1087,7 +1087,14 @@ function buildAgentExtras(context: AgentContext, agentTypes: string[] = []): str
 
   if (context.gameState) {
     parts.push(`<current_game_state>`);
-    parts.push(JSON.stringify(context.gameState));
+    // Strip presentCharacters[].avatarPath before serialization: it is a UI display
+    // field (image path or inline data URL) the model never reads, and a single
+    // inlined data: URL can be megabytes of base64 - imported pre-refactor chats
+    // ship with full data URLs in this field and were blowing past 1M-token
+    // provider ceilings (issue #1188).
+    parts.push(
+      JSON.stringify(context.gameState, (key, value) => (key === "avatarPath" ? undefined : value)),
+    );
     parts.push(`</current_game_state>`);
   }
 
--- a/src/engine/agents-runtime/executor/agent-executor.ts
+++ b/src/engine/agents-runtime/executor/agent-executor.ts
@@ -10,6 +10,7 @@ import type {
 import type { AgentResult, AgentContext, AgentResultType } from "../../contracts/types/agent";
 import { getDefaultAgentPrompt } from "../../contracts/constants/agent-prompts";
 import { DEFAULT_AGENT_CONTEXT_SIZE, DEFAULT_AGENT_MAX_TOKENS, MAX_AGENT_MAX_TOKENS, MIN_AGENT_MAX_TOKENS } from "../../contracts/types/agent";
+import { stripAvatarPathsReplacer } from "../strip-avatar-paths";
 const isDebugAgentsEnabled = () => false;
 const logger = {
   debug: (..._args: unknown[]) => {},
@@ -926,7 +927,7 @@ function buildAgentMessages(
         if (gs.playerStats) trackerSummary.playerStats = gs.playerStats;
         if (gs.personaStats?.length) trackerSummary.personaStats = gs.personaStats;
         if (Object.keys(trackerSummary).length > 0) {
-          content += `\n\n<committed_tracker_state>\n${JSON.stringify(trackerSummary)}\n</committed_tracker_state>`;
+          content += `\n\n<committed_tracker_state>\n${JSON.stringify(trackerSummary, stripAvatarPathsReplacer)}\n</committed_tracker_state>`;
         }
       }
 
@@ -1087,14 +1088,7 @@ function buildAgentExtras(context: AgentContext, agentTypes: string[] = []): str
 
   if (context.gameState) {
     parts.push(`<current_game_state>`);
-    // Strip presentCharacters[].avatarPath before serialization: it is a UI display
-    // field (image path or inline data URL) the model never reads, and a single
-    // inlined data: URL can be megabytes of base64 - imported pre-refactor chats
-    // ship with full data URLs in this field and were blowing past 1M-token
-    // provider ceilings (issue #1188).
-    parts.push(
-      JSON.stringify(context.gameState, (key, value) => (key === "avatarPath" ? undefined : value)),
-    );
+    parts.push(JSON.stringify(context.gameState, stripAvatarPathsReplacer));
     parts.push(`</current_game_state>`);
   }
 
--- a/src/engine/agents-runtime/knowledge/knowledge-router.ts
+++ b/src/engine/agents-runtime/knowledge/knowledge-router.ts
@@ -2,6 +2,7 @@ import type { AgentContext, AgentResult } from "../../contracts/types/agent";
 import type { LorebookEntry } from "../../contracts/types/lorebook";
 import type { BaseLLMProvider } from "../../generation-core/llm/base-provider.js";
 import { executeAgent, type AgentExecConfig } from "../executor/agent-executor.js";
+import { stripAvatarPathsReplacer } from "../strip-avatar-paths.js";
 import {
   scanForActivatedEntries,
   type ScanMessage,
@@ -159,7 +160,7 @@ export function buildKnowledgeRouterQuery(context: AgentContext): string {
     .map((message) => message.content.trim())
     .filter(Boolean);
   if (context.chatSummary?.trim()) parts.unshift(context.chatSummary.trim());
-  if (context.gameState) parts.push(JSON.stringify(context.gameState));
+  if (context.gameState) parts.push(JSON.stringify(context.gameState, stripAvatarPathsReplacer));
   return parts.join("\n\n");
 }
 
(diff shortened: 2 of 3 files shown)


Romuromylus:
Pushed `d9be35c` to address the review.

**Both Medium findings addressed.** Extracted `stripAvatarPathsReplacer` into `src/engine/agents-runtime/strip-avatar-paths.ts` and routed all three serialization sites through it:

1. `buildAgentExtras` `<current_game_state>` block (the original site).
2. `buildAgentMessages` `<committed_tracker_state>` block at agent-executor.ts:929 - the last-3-assistant-messages tracker snapshot still pulled `gs.presentCharacters` whole, so structured-output agents kept inheriting the avatarPath data URLs from history. Now stripped at serialization.
3. `buildKnowledgeRouterQuery` at knowledge-router.ts:162 - same `JSON.stringify(context.gameState)` shape, latent until the router agent is active on an imported chat. Now stripped at serialization.

Co-locating in one named helper makes the rule findable by symbol search and keeps future serialization sites a one-import-and-pass-the-replacer change rather than rediscovering the rule per-site.

**On open question 1** (inline vs named helper): you were right - the original inline replacer was expedient rather than principled. The named helper is the better default; the file's docstring captures the why so the next contributor doesn't have to read this issue thread to understand the rule.

**On open question 2** (`<committed_tracker_state>` carrying full `presentCharacters` including `appearance` / `outfit` / `thoughts` / `customFields` on 3 history messages): genuinely worth exploring, but I'd rather keep that as a separate PR. The history block was deliberately trimmed to last-3 messages and structured-output-only for token budget reasons already, and tightening the per-character shape further has a wider blast radius than this fix (agent prompt templates may rely on specific fields being present in history). Worth a separate analysis pass with its own repro for any agent that reads tracker history.

**Follow-up issue for the silent agent-failure surface**: filing as a separate issue now so it doesn't drift. The 400 storm in this bug was invisible only because `setFailedAgentFailures` / `failedAgentTypes` in `agent.store.ts` is defined but never called from `applyAgentResultEffects`; surfacing genuine agent failures would have shortened this entire debug loop.

Baseline still green (`pnpm check:frontend`, `pnpm check:architecture`, `pnpm check:docs`). Live re-verified in `pnpm tauri dev` against the same imported chat - tracker + agent activity panels still populate after a turn.

PR #1202, opened 2026-05-25
Fix bare speaker tag rendering

### LINKED ISSUE #1126: [Bug]: Bare <speaker> tags (no name attribute) render as literal text
## What's happening

The plain-text rendering path in `ChatMessage.tsx` strips well-formed `<speaker="Name">...</speaker>` tags just fine, but bare `<speaker>...</speaker>` (no `name` attribute) slips through and shows up in the chat as literal text.

## What should happen

Bare `<speaker>` tags should either be stripped like the named version, or filtered out upstream. Either's fine — consistency is what matters.

## What actually happens

A message containing `<speaker>hello</speaker>` renders the literal characters `<speaker>hello</speaker>` in the message body.

## Why it's happening

Two regexes in the file disagree about what a speaker tag looks like:

- The pre-HTML-detection stripper (around line 562): `/<\/?speaker(?:="[^"]*")?>/g` — name attribute optional
- The renderer's `SPEAKER_TAG_RE` (around line 250): `/<speaker="([^"]*)">([\s\S]*?)<\/speaker>/g` — name attribute required

The stripper removes wrappers from a temporary copy so it can decide whether the message contains "real" HTML, but it doesn't touch the original text. The plain-text branch then runs the stricter renderer regex against that unmodified text, fails to match bare tags, and they fall through to the highlighter unchanged.

## How to reproduce

1. Get a message into the chat that contains `<speaker>some text</speaker>` (without a `name="..."` attribute).
2. The literal tag characters appear in the rendered message.

## Possible fix

Pick a side and align the two regexes — either teach the renderer to accept bare tags (treat speaker as null), or require name attributes everywhere and sanitize the bare ones out earlier. The right choice probably depends on #1127 (the parse/render refactor) — that's a more natural place to settle this.

---

Surfaced during code review of #1125. Pre-existing, not introduced by that PR.

cha1latte:
## What?

Fixes bare `<speaker>...</speaker>` wrappers rendering as literal text in the refactor chat message UI.

The shared `ChatMessage` speaker-tag renderer now accepts both bare and named speaker wrappers, so bare tags are stripped consistently while named speaker coloring still works.

Fixes #1126

## Why?

The pre-HTML detection path already treated the speaker name as optional, but the actual plain-text renderer only matched `<speaker="Name">...</speaker>`. That mismatch let existing or newly generated bare speaker tags leak into chat bubbles.

## How?

- Updated the shared speaker tag regex to match `<speaker>...</speaker>` and `<speaker="Name">...</speaker>`.
- Treated unnamed speaker tags as default-color dialogue.
- Kept named speaker color lookup intact for group/merged chat rendering.

## Machine Verification

- `pnpm exec vitest run --config scratch/vitest.issue-1126.config.ts`
  - Before fix on clean `upstream/refactor`: bare tag cases failed and leaked literal markup.
  - After fix: 5/5 passed.
  - Covered single bare wrapper, named wrapper regression, multiple bare wrappers, mixed bare/named wrappers, and negative control `<speakerish>...</speakerish>`.
- `pnpm check` passed:
  - architecture check
  - frontend typecheck
  - Rust `cargo check`
  - docs check
- `node C:\Users\celia\.codex\tools\marinara-proof-gate.mjs scratch\bugfix-verification.json` passed.
- Bunny review passed before PR open.

## Evidence

Issue 1126 before/after proof

Evidence URL verified with `curl.exe -I` returning `200 OK`.

## Manual Verification

None required for the core claim. The changed path was exercised directly with a component render harness against `ChatMessage`.

## Notes

Targeting `refactor` because this issue was requested for the refactor lane and the relevant renderer has moved from `packages/client/...` to `src/features/modes/shared/chat-ui/components/ChatMessage.tsx`.

### DIFF
--- a/src/features/modes/shared/chat-ui/components/ChatMessage.tsx
+++ b/src/features/modes/shared/chat-ui/components/ChatMessage.tsx
@@ -169,8 +169,8 @@ interface ChatMessageProps {
 /** Regex to match a plain image URL as the entire content. */
 const IMAGE_URL_RE = /^https?:\/\/\S+\.(?:gif|png|jpe?g|webp)(?:\?[^\s]*)?$/i;
 
-/** Regex to match <speaker="name">dialogue</speaker> tags. */
-const SPEAKER_TAG_RE = /<speaker="([^"]*)">([\s\S]*?)<\/speaker>/g;
+/** Regex to match <speaker>dialogue</speaker> and <speaker="name">dialogue</speaker> tags. */
+const SPEAKER_TAG_RE = /<speaker(?:="([^"]*)")?>([\s\S]*?)<\/speaker>/g;
 const INLINE_MARKDOWN_CONTAINER_RE =
   /\*\*\*[\s\S]+?\*\*\*|\*\*[\s\S]+?\*\*|__[\s\S]+?__|(?<!\*)\*(?!\*)[\s\S]+?(?<!\*)\*(?!\*)|==[\s\S]+?==|~~[\s\S]+?~~|(?<![_\w])_[^_]+?_(?![_\w])/g;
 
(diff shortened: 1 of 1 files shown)


cha1latte:
@coderabbitai review

Issue #1512, opened 2026-05-29
Import Profile is a destructive full-replace with no confirmation and wipes collections absent from the file


Romuromylus:
## Summary

Importing a profile replaces the entire local profile with no warning. Collections that are not present in the imported file are wiped to empty rather than left untouched.

## Repro

1. Start from an instance with data (e.g. 31 characters, 32 lorebooks, 20 personas, 5 chats).
2. Import a profile whose `fileStorage.tables` contains only `chats` and `messages`.
3. Result: chats/messages are replaced as expected, but characters -> 0, lorebooks -> 0, personas -> 0. Every collection not in the file is wiped.

## Root cause

`src-tauri/src/commands/storage/profile/legacy.rs` builds `replacements` for the full fixed table-mapping list and calls `replace_all_many_and_then(...)`. Any mapped collection missing from the imported file is replaced with an empty array. `src/features/shell/settings/components/ProfileImportSection.tsx` calls `importProfileFile` immediately after the file picker, with no confirmation dialog and no "this will replace your data" warning.

## Impact

A full, complete backup restores correctly. But importing a partial profile, an older backup missing a table, or a hand-built export silently destroys everything not included, with no prompt and no undo.

## Suggested directions (for discussion)

- Add a confirmation step before import that states the data will be replaced.
- Consider leaving collections absent from the file untouched instead of wiping them, or make replace-vs-merge explicit in the UI.

## Environment

refactor @ ef244c80, Windows 11, Tauri desktop (`pnpm tauri dev`).

---

Found during smoke test #1484 (Heavy Data / Legacy / Import Pass). Project board: https://github.com/orgs/Pasta-Devs/projects/4

Romuromylus:
Fixed by #1518 (merged into `refactor` as a9cc0827). The legacy import no longer wipes collections absent from the file (and a present-but-non-array table is now rejected rather than treated as a clear), and a confirmation dialog gates the import. The modern `data.collections` import path still full-replaces absent collections; that is tracked separately in #1521.

PR #1633, opened 2026-05-30
Guard modern profile import against wiping absent collections

### LINKED ISSUE #1521: Profile import: modern collections path still full-replaces collections absent from the file
Follow-up to #1512 (fixed for the legacy table path in #1518).

#1518 made the **legacy** import path (`fileStorage.tables`) skip-absent: a collection whose table key is missing from the file is left untouched instead of wiped. The **modern** `data.collections` import path (`src-tauri/src/commands/storage/profile.rs`, `import_profile_collections_with_restored_assets`) still rebuilds replacements for every mapped collection and falls back to an empty array for ones absent from the file, so a partial modern-format profile would still wipe the collections it omits.

In practice the refactor's own exports are always full backups, so this is not triggered by a normal export/import round-trip. But a hand-built or partial modern-format file has the same data-loss shape #1512 described for the legacy path.

Suggested: apply the same present-key guard to the modern collections path so absent collections are left untouched, while present-but-empty still clears.

Romuromylus:
## Why this change

Follow-up to #1521. #1518 fixed the **legacy** table import path so a partial backup leaves collections it omits untouched instead of wiping them. The **modern** `data.collections` path (`import_profile_collections_with_restored_assets`) still rebuilt a replacement for every known collection, falling back to an empty array for any collection the file omitted. A partial or hand-built modern-format profile therefore wiped every collection it did not carry, the same data-loss shape #1512 described for the legacy path.

The refactor's own exports are always full backups, so a normal export/import round-trip never hit this. A hand-built or partial modern file did.

## What changed

- `src-tauri/src/commands/storage/profile.rs`: in the modern collections loop, skip collections absent from the file (leaving the user's existing data untouched) and reject a present-but-non-array collection instead of coercing it to an empty array. A present-but-empty array is still treated as a deliberate clear. This mirrors the #1518 legacy guard exactly.
- Added three regression tests against the real storage layer: skip-absent leaves existing rows, present-but-empty clears, present-but-non-array is rejected with nothing replaced.

## Verification

- `pnpm check` passes locally (architecture + frontend tsc + cargo check + docs).
- `cargo clippy --workspace --all-targets -- -D warnings` clean.
- New unit tests exercise the exact data-loss path against the real storage engine and pass; full `profile` module suite (30 tests) green.

The end-user trigger (importing a hand-built/partial modern-format profile) is reached through the native file picker, which is not driven by the automated tiers; the storage-layer behavior it depends on is fully covered by the new tests.

## Linked issue

Closes #1521.

## Release Notes

* **Bug Fixes**
  * Profile collection imports now preserve existing stored data when collections are absent from the import file, instead of clearing them
  * Non-array formatted collection data is validated and rejected during import before any changes take effect, safeguarding against data loss

* **Tests**
  * Added tests verifying profile import behavior with absent collections, empty arrays, and invalid data formats

[Review Change Stack](https://app.coderabbit.ai/change-stack/Pasta-Devs/Marinara-Engine/pull/1633?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src-tauri/src/commands/storage/profile.rs
+++ b/src-tauri/src/commands/storage/profile.rs
@@ -177,9 +177,26 @@ where
     let mut replacements = Vec::new();
     let mut unsupported_prompt_overrides = 0usize;
     for collection in PROFILE_COLLECTIONS {
-        let mut rows = collections
-            .get(*collection)
-            .and_then(Value::as_array)
+        // A partial modern profile (a hand-built export, or a file missing a
+        // collection) must not wipe collections it does not carry. Skipping the
+        // replacement leaves the user's existing collection untouched; a
+        // collection that is present but empty is still an explicit clear and
+        // falls through to a normal empty replacement. Mirrors the legacy table
+        // path guard added in #1518.
+        let Some(collection_value) = collections.get(*collection) else {
+            continue;
+        };
+        // A present-but-non-array collection is malformed (e.g. `"characters": {}`).
+        // Coercing it to an empty array would silently clear the collection - the
+        // same data loss the absent-key skip above guards against. Reject the
+        // import instead so nothing is replaced.
+        if !collection_value.is_array() {
+            return Err(AppError::invalid_input(format!(
+                "Profile collection `{collection}` must be a JSON array"
+            )));
+        }
+        let mut rows = collection_value
+            .as_array()
             .cloned()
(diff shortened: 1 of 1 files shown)


Romuromylus:
@coderabbitai review

Issue #1513, opened 2026-05-29
Legacy import leaves boolean/array lorebook-entry fields as strings, breaking editor display and activation logic


Romuromylus:
## Summary

After a legacy profile import, several lorebook-entry fields are stored as JSON-encoded strings instead of native types. This makes the entry editor display the wrong toggle state, and because one generation path uses raw truthiness, it can also change which entries activate.

## What is wrong (confirmed on imported entry "NPC: Kafka", key "Kafka")

Read back via `storage_list { entity: "lorebook-entries" }`:

- `caseSensitive` = "false" (string)            (should be boolean)
- `constant` = "false" (string)
- `locked` = "false" (string)
- `preventRecursion` = "false" (string)
- `excludeFromVectorization` = "false" (string)
- `selective` = "true" (string), `selectiveLogic` = "and" (string)
- `enabled` = "true" (string)
- `matchWholeWords` = true (boolean)            <- correctly normalized
- `characterFilterIds` / `characterTagFilters` / `additionalMatchingSources` / `activationConditions` = "[]" (string, should be arrays)
- `keys` / `secondaryKeys` = arrays             <- correctly normalized

All 1492 entries in the dataset share this pattern.

## Confirmed symptoms

1. Editor: opening the entry shows Case Sensitive, Locked, No Recursion, and No Vector toggled ON, even though they are OFF on the source instance. `"false"` is truthy in JS, so the raw string reads as on. (Whole Words is correct because `matchWholeWords` is a real boolean.)
2. Save is NOT corrupting: editing an unrelated field and letting it auto-save round-trips the string values unchanged (verified on disk), so the stored data is not silently flipped to `true`.
3. Generation is inconsistent: `prompt-assembly.ts` uses a `boolish()` coercion and `makers.ts` uses `=== true`, but `keyword-scanner.ts` uses raw truthiness (`if (entry.constant)` at lines 454/508, `.filter((a) => !a.entry.preventRecursion)` at 552, and passes `caseSensitive` raw at 479). With string `"false"` being truthy, the scanner can wrongly treat entries as constant and wrongly suppress recursion.

## Root cause

The legacy normalizer (`src-tauri/src/commands/storage/profile/legacy.rs`) normalizes some fields (`matchWholeWords`, `keys`, `secondaryKeys`) but not `caseSensitive` / `constant` / `locked` / `preventRecursion` / `excludeFromVectorization` / `selective`, and it does not decode the string-encoded array fields.

[...]

PR #1637, opened 2026-05-30
Normalize lorebook-entry fields at the generic storage boundary

### LINKED ISSUE #1571: [Issue]: Generic storage writes can persist malformed lorebook-entry fields
## Summary

Generic storage writes can still persist malformed `lorebook-entries` records even though the legacy import path now normalizes string booleans and JSON-like arrays.

## Expected behavior

Every subsystem that writes `lorebook-entries` should cross one strong record-shape contract. Boolean, array, number, metadata, timing, and scope fields should either be normalized to the native shape or rejected before they reach persistent storage, regardless of whether the row came from the editor, bulk creation, import, tool-generated entries, remote runtime, or another generic storage caller.

## Actual behavior

The legacy import path was fixed for #1513, but the generic storage boundary remains loosely typed. Source audit against `upstream/refactor` found:

- `src/features/catalog/lorebooks/hooks/use-lorebooks.ts` creates, updates, copies, moves, reorders, and bulk-creates lorebook entries through `storageApi` with `Record<string, unknown>` / `unknown[]` payloads.
- `src/shared/api/storage-api.ts` normalizes stored JSON for chats and messages, but does not normalize `lorebook-entries`.
- `src/engine/contracts/schemas/lorebook.schema.ts` exports `createLorebookEntrySchema` and `updateLorebookEntrySchema`, but current grep found no active call sites using those schemas at the write boundary.
- Native import normalization in `src-tauri/src/commands/storage/imports/normalization.rs` handles legacy lorebook fields, but that does not protect later generic `storage_create` / `storage_update` calls.

That means a future or existing caller that bypasses the fixed legacy import path can still hand strings like `"false"` or JSON strings like `"[\"moon\"]"` across the storage boundary and rely on downstream UI/generation code to guess the intended shape.

## Steps to reproduce

1. Use a generic storage write path such as `storage_create` / `storage_update` for entity `lorebook-entries`.
2. Send a malformed but plausible legacy-style payload, for example `enabled: "false"`, `constant: "true"`, `keys: "[\"moon\"]"`, or `secondaryKeys: "[]"`.
3. Read the row through the lorebook editor, active lorebook scan, or `search_lorebook` tooling.
4. Observe that the row shape depends on caller-side assumptions instead of one enforced storage/lorebook-entry contract.

## Impact area

- [ ] App shell, navigation, startup, or providers
- [x] Catalog resources
- [ ] Chat mode
- [ ] Roleplay mode
- [ ] Game mode
- [ ] Shared mode UI
- [ ] Professor Mari or shell tools
- [x] React-free engine behavior, contracts, agents, or capability ports
- [x] Shared API wrapper or remote runtime routing
- [ ] Docs, contributor workflow, or templates

## Environment

- Branch or commit: source audit against `upstream/refactor` at `ab8af081`
- Run mode: source audit / smoke-test follow-up
- OS + version: Windows desktop, exact version not captured
- Node + pnpm versions: Not specified
- Rust version, if relevant: Not specified
- Remote Runtime URL configured? not relevant
- Provider, model, or integration, if relevant: not relevant

No screenshot. This was found during the focused system-boundary/data-contract audit after #1513 was closed.

Relevant source signals:

- `src/features/catalog/lorebooks/hooks/use-lorebooks.ts`
- `src/shared/api/storage-api.ts`
- `src/engine/contracts/schemas/lorebook.schema.ts`
- `src-tauri/src/commands/storage/imports/normalization.rs`

## Regression notes

Related to #1513, but not a duplicate. #1513 covered legacy import normalization. This issue covers the remaining generic storage write boundary for lorebook-entry records.

Also related to #1566 because Knip reports the lorebook entry schemas as unused, which is a useful signal that the intended TypeScript contract may not be wired into active write paths.

## Additional context

Suggested regression coverage should include at least one positive normalized row and one negative/malformed row for lorebook-entry create/update outside the legacy import path.

## Template check

Romuromylus:
## Why this change

Follow-up to #1571 (related to but not a duplicate of #1513). #1513 fixed legacy lorebook-entry **import** normalization, but the generic storage write boundary stayed loosely typed: the `lorebook-entries` arm of `normalize_typed_json_fields` only normalized `keys` and `secondaryKeys`. Any caller that bypassed the legacy import path (the editor, bulk create, copy/move/reorder, tool-generated entries, remote runtime) could still hand `enabled: "false"`, `constant: "true"`, `keys: "[\"moon\"]"`, or a JSON-string object across the boundary and rely on downstream UI/generation to guess the intended shape.

`normalize_typed_json_fields` is the single contract both write paths cross: `storage_create` through `with_entity_defaults`, and `storage_update` through `normalize_update_patch`. Strengthening that one arm closes the gap for every generic caller at once.

## What changed

- `src-tauri/src/commands/storage/shared.rs`: extend the `lorebook-entries` arm to normalize the full native record shape:
  - the nine boolean flags (`enabled`, `constant`, `selective`, `matchWholeWords`, `caseSensitive`, `useRegex`, `preventRecursion`, `locked`, `excludeFromVectorization`) via the existing `normalize_boolish_fields`,
  - the remaining array fields (`characterFilterIds`, `characterTagFilters`, `generationTriggerFilters`, `additionalMatchingSources`, `activationConditions`) alongside `keys`/`secondaryKeys`,
  - the `relationships`/`dynamicState` objects and the nullable `schedule` object.
  - Values that cannot be coerced are rejected rather than stored, matching the existing array-field contract.
- Added three tests: legacy-string coercion (positive), native row passes through unchanged (negative control), and an unparseable array field is rejected.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` clean.
- New tests plus the existing lorebook (19) and normalization (31) suites pass; the legacy import path (which also calls this function) is unaffected because its rows are already coerced before reaching it and the new object/schedule fields are absent on legacy rows.

The end-user write paths reach this boundary through generic `storage_create`/`storage_update`, fully covered by the new unit tests against the real normalization function.

## Known caveat: numeric fields

Numeric lorebook-entry fields (`probability`, `scanDepth`, `position`, `depth`, `order`, `sticky`, `cooldown`, `delay`, `ephemeral`, `groupWeight`) are intentionally left untouched. There is no numeric-coercion primitive at this boundary today and the legacy contract never covered them, so adding one is out of scope for this fix. If a malformed numeric caller ever appears it warrants a dedicated follow-up with its own primitive.

## Linked issue

Closes #1571.

### DIFF
--- a/src-tauri/src/commands/storage/shared.rs
+++ b/src-tauri/src/commands/storage/shared.rs
@@ -377,7 +377,40 @@ pub(crate) fn normalize_typed_json_fields(
             normalize_json_array_fields(object, &["tags", "characterIds", "personaIds"])?;
         }
         "lorebook-entries" => {
-            normalize_json_array_fields(object, &["keys", "secondaryKeys"])?;
+            // The generic storage boundary is the single contract every
+            // lorebook-entry write crosses (editor, bulk create, copy/move,
+            // tool-generated entries, remote runtime). Coerce every legacy-style
+            // shape - string booleans, JSON-string arrays, and JSON-string
+            // objects - to the native shape, or reject what cannot be coerced,
+            // so downstream UI/generation never has to guess the intended type.
+            normalize_json_array_fields(
+                object,
+                &[
+                    "keys",
+                    "secondaryKeys",
+                    "characterFilterIds",
+                    "characterTagFilters",
+                    "generationTriggerFilters",
+                    "additionalMatchingSources",
+                    "activationConditions",
+                ],
+            )?;
+            normalize_boolish_fields(
+                object,
+                &[
+                    "enabled",
+                    "constant",
+                    "selective",
+                    "matchWholeWords",
+                    "caseSensitive",
+                    "useRegex",
+                    "preventRecursion",
+                    "locked",
+                    "excludeFromVectorization",
+                ],
+            );
+            normalize_json_object_fields(object, &["relationships", "dynamicState"])?;
+            normalize_nullable_json_object_fields(object, &["schedule"])?;
         }
         "connections" => {
             normalize_nullable_json_object_fields(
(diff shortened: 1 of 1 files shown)


Romuromylus:
@coderabbitai review

Romuromylus:
@coderabbitai review

Romuromylus:
Clarification on the description's "rejected rather than stored" line: that applies to the **array and object** fields (they route through `normalize_json_field`, which returns an error on an uncoercible value). The **boolean** fields use the existing `normalize_boolish_fields` helper (shared with chat-presets), which is lenient: a value that is neither a bool, a recognized true/false string, nor a number falls back to `false` rather than erroring. That matches the repo's established boolean-coercion convention; it is normalize-with-default, not reject. Also, the numeric fields are intentionally left untouched (no numeric-coercion primitive at this boundary today) and tracked as a follow-up, so #1571's numeric expectation is deferred rather than fully closed here.

PR #1639, opened 2026-05-30
Stop the connections sidebar squishing during panel resize

### LINKED ISSUE #1586: [Issue]: connections sidebar squishes and lags during resize — framer-motion layout animations + transition-all
## Summary

The connections sidebar behaves incorrectly during resize — content squishes, squashes, and the resize feels sluggish. No other right-side panel has this problem.

## Expected behavior

Resize feels instant and smooth, content reflows cleanly like CharactersPanel or LorebooksPanel.

## Actual behavior

- Content squishes/squashes visually while dragging the resize handle
- Resize feels sluggish compared to all other right-side panels

Image
Image

## Steps to reproduce

1. Open the Connections sidebar
2. Drag the resize handle left and right
3. Compare to Characters or Lorebooks sidebar — those resize cleanly

## Environment

- Marinara Engine version: Refactor
- Install type: MacOS App from Source <---> HTTP Docker Backend on Ubuntu
- OS + version: macOS (Tauri) + Linux server (Docker/web)

## Additional context

Two root causes, both unique to ConnectionsPanel:

**1. framer-motion `Reorder` layout animations (the squish)**

ConnectionsPanel is the only right-side panel using `Reorder.Group` / `Reorder.Item` (for folder drag-to-reorder). These components track layout positions internally and fire animations on every layout change — including panel resize. Every pixel dragged triggers framer-motion measuring and re-animating all folder rows. No other panel uses `Reorder`.

**2. `transition-all` on multiple elements (the sluggishness)**

- `TTSConfigCard` root div: `transition-all` causes `width` to animate on every resize pixel
- `ConnectionRow` root div: same `transition-all`, one per connection, all firing simultaneously

Other panels use `transition-colors` or targeted transitions only.

✱Maybe Fixes that:**
- `TTSConfigCard` line 676 + `ConnectionRow` line 98: `transition-all` → `transition-colors`
- For the Reorder squish: either pass `isResizing` from AppShell to suppress animations during drag, or replace `Reorder.Group`/`Reorder.Item` with `dnd-kit`

## Template check

Romuromylus:
## Why this change

The connections right panel was the only right-side panel that lagged and squished while dragging the panel resizer. Two independent causes, both unique to it:

1. **`transition-all` on inner cards.** The `ConnectionRow` and `TTSConfigCard` root divs used `transition-all`, which animates `width`. Every connection card therefore ran a width transition on every resize frame.
2. **framer-motion `Reorder` layout animation.** `ConnectionsPanel` is the only right-side panel using `Reorder`. Each `Reorder.Item` defaults to `layout=true` and runs a layout-projection measure/animate on every width change, producing the squish/squash.

The AppShell `<aside>` already disables `transition-[width]` during drag, so the container was fine; the jank was entirely these two inner sources.

## What changed

- `ConnectionsPanel.tsx` / `TTSConfigCard.tsx`: swap the two `transition-all` root classes to `transition-colors`. The only animated effect on either element is a hover/expand color change, so behavior is preserved while the per-frame width animation is gone.
- `ui/model.ts` + `ui.store.ts`: add a transient `rightPanelResizing` flag and setter, beside the existing transient `centerCompact`. Intentionally **not** added to `partializeUiState`, so it is never persisted.
- `AppShell.tsx`: set the flag on right-panel drag start and clear it in the shared `finishResize` teardown (covering mouseup and the window-blur path).
- `ConnectionsPanel.tsx` (folder rows): read the flag and pass `layout=false` to `Reorder.Item` while resizing, suppressing the per-item layout projection. `layout` is omitted from `ReorderItemProps`'s public type even though the runtime forwards it, so it goes through a loosely-typed spread (commented). Drag-to-reorder is unchanged because the reorder measurement is independent of the layout flag.

## Verification

- `pnpm exec tsc -b` clean (validates the new store field and the typed-omit spread).
- New `ui.store.test.ts` cases pass: flag defaults to `false`, toggles via its setter, and is absent from `partializeUiState` (never persisted).
- Playwright boot repro: the app mounts with the new wiring and a `transition-colors` element does not animate `all`; no error from the new symbols.

## Manual

The perceived smoothness (framer-motion layout suppression) and drag-to-reorder-after-resize are visual/interaction behaviors not capturable by an assertion; verified by driving the connections panel in the browser tier. Worth a quick look on a real display.

## Risk

`layout={false}` is passed via a loosely-typed spread because framer-motion 12 omits `layout` from `ReorderItemProps` while still forwarding it at runtime; a future framer-motion major could change that forwarding (noted in a code comment). Mobile is unaffected (the resize gesture returns early on mobile, so the flag never sets).

## Linked issue

Closes #1586.

### DIFF
--- a/src/app/shell/AppShell.tsx
+++ b/src/app/shell/AppShell.tsx
@@ -197,6 +197,7 @@ export function AppShell() {
   const rightPanelOpen = useUIStore((s) => s.rightPanelOpen);
   const rightPanelWidth = useUIStore((s) => s.rightPanelWidth);
   const setRightPanelWidth = useUIStore((s) => s.setRightPanelWidth);
+  const setRightPanelResizing = useUIStore((s) => s.setRightPanelResizing);
   const closeRightPanel = useUIStore((s) => s.closeRightPanel);
   const trackerPanelEnabled = useUIStore((s) => s.trackerPanelEnabled);
   const trackerPanelOpen = useUIStore((s) => s.trackerPanelOpen);
@@ -397,6 +398,7 @@ export function AppShell() {
       document.body.style.userSelect = "none";
       rightPanelDragWidthRef.current = rightPanelWidth;
       setRightPanelDragWidth(rightPanelWidth);
+      setRightPanelResizing(true);
 
       const onMove = (moveEvent: MouseEvent) => {
         const nextWidth = clampWidth(
@@ -414,6 +416,7 @@ export function AppShell() {
         setRightPanelWidth(rightPanelDragWidthRef.current ?? useUIStore.getState().rightPanelWidth);
         rightPanelDragWidthRef.current = null;
         setRightPanelDragWidth(null);
+        setRightPanelResizing(false);
         document.body.style.cursor = originalCursor;
         document.body.style.userSelect = originalUserSelect;
         window.removeEventListener("mousemove", onMove);
@@ -425,7 +428,7 @@ export function AppShell() {
       window.addEventListener("mouseup", finishResize);
       window.addEventListener("blur", finishResize);
     },
-    [isMobile, rightPanelWidth, setRightPanelWidth],
+    [isMobile, rightPanelWidth, setRightPanelWidth, setRightPanelResizing],
   );
 
   const adjustSidebarWidth = useCallback(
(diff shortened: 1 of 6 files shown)


Romuromylus:
Known limitation: the left chat sidebar's folder list uses the same framer-motion `Reorder` pattern and will still show the squish when the *left* sidebar is drag-resized. This PR deliberately scopes to the right-side connections panel (the reported surface); generalizing the resize flag to the left sidebar is a small follow-up.

Romuromylus:
@coderabbitai review

Romuromylus:
@coderabbitai review

PR #1641, opened 2026-05-30
Surface invalid combat action responses instead of advancing combat

### LINKED ISSUE #1520: [Issue]: Malformed encounter action output silently advances combat
During the Roleplay A13 native encounter audit on `origin/refactor` commit `229e7d89`, malformed combat action model output was treated as a successful combat turn instead of surfacing a recoverable error.

Related smoke umbrella: #1486.

## Repro

1. In a Roleplay chat with the `combat` agent enabled and a deterministic local OpenAI-compatible test connection, click `Encounter`.
2. Click `Begin Combat`.
3. Let the model return a valid initial combat JSON object.
4. Submit a combat action where the model response to the action request is non-JSON / malformed.

## Expected

Malformed combat action output should be understandable and recoverable. The encounter should not fake a successful turn without telling the user that the model output was invalid. A reasonable result would be an inline combat error with retry/close options and unchanged combat state.

## Actual

The invalid action response did not surface an error. The encounter silently used deterministic fallback action resolution, advanced combat state, and in the A13 repro ended the encounter as `victory`.

Because the fallback path treated the turn as successful, the combat summary path then wrote back to chat. In the deterministic repro the visible chat got an assistant summary row after the malformed action path, even though the action model output was invalid.

## Evidence

Native A13 probe:

- initial encounter rendered `A13 Training Dummy`, party HP, enemy HP, and custom action controls
- normal action updated enemy/party HP and combat log as expected
- malformed action response was `A13_MALFORMED_NON_JSON_RESPONSE`
- UI did not show an invalid-output or retry error
- encounter advanced to `VICTORY`
- summary writeback occurred afterward

Code path:

- `completeJsonObject(...)` catches/returns `null` for failed parse/LLM errors
- `resolveRoleplayEncounterAction(...)` passes that through `sanitizeCombatActionResult(...)`
- `sanitizeCombatActionResult(...)` uses `fallbackActionResult(...)`
- `useEncounter.sendAction(...)` receives a valid-looking object and therefore does not render its `AI returned an invalid response. Try again.` error path

## Likely Owner

`src/engine/modes/roleplay/encounter/encounter-service.ts` and `src/features/modes/roleplay/encounter/hooks/use-encounter.ts`.

The service likely needs to distinguish "sanitized malformed fields from a valid response" from "no valid response at all", so the UI can show a retryable combat error without silently inventing a successful turn.

Romuromylus:
## Why this change

A malformed / non-JSON combat action response was silently treated as a successful turn. `completeJsonObject` returns `null` on unparseable output or an LLM error, but `resolveRoleplayEncounterAction` then ran `recordValue(parsed?.result) ?? parsed` (which is `null`) through `sanitizeCombatActionResult`, which rebuilds a fully valid-looking `CombatActionResult` from the deterministic fallback. The caller had no way to tell "sanitized a garbage response" from "a real turn", so it advanced party/enemy HP and the log, and could even reach victory + a summary writeback to the chat from a malformed reply.

## What changed

- `combat-encounter.ts`: add an optional `invalid?: boolean` to `EncounterActionResponse` (documented: the caller must surface a retryable error and leave combat unchanged).
- `encounter-service.ts`: when `completeJsonObject` returns `null`, return `{ result: fallback, invalid: true }` before the sanitize path runs. The fallback is returned only to satisfy the required `result` field; the caller must not apply it.
- `use-encounter.ts`: check `res.invalid` first and reuse the existing recoverable error path (`setError("AI returned an invalid response. Try again.")` + clear processing + early return), which runs before any `updateCombat`/`endCombat`/`generateSummary`, so combat state and the chat are untouched and the retry controls stay mounted. No new UI.

A present-but-partial response (a real record missing `result`/`combatStats`) is still a real turn and continues through the sanitizer unchanged, and the genuine empty-input `interrupted` branch stays reachable, because only a `null` parse flags invalid.

## Verification

- `pnpm exec tsc -b` clean; `pnpm check:architecture` clean (no new cross-layer import).
- New engine tests (`encounter-service.test.ts`) cover: malformed non-JSON output flags invalid, an LLM error flags invalid, a valid JSON turn resolves normally (not invalid), and a present-but-partial response is not flagged invalid.

## Risk / note

This also routes a transient LLM-call failure (timeout, 5xx) into the same retryable error instead of the previous silent fallback turn. That is the intended, more-correct behavior, but it does broaden the trigger beyond pure malformed JSON, so it is called out here. The hook + inline error rendering are covered by code reading; the engine root cause is covered by the unit tests.

## Linked issue

Closes #1520.

### DIFF
--- a/src/engine/contracts/types/combat-encounter.ts
+++ b/src/engine/contracts/types/combat-encounter.ts
@@ -223,6 +223,12 @@ export interface EncounterActionRequest {
 /** Response from encounter action resolution. */
 export interface EncounterActionResponse {
   result: CombatActionResult;
+  /**
+   * True when the model output could not be parsed into a usable combat turn.
+   * The caller must surface a retryable error and leave combat state unchanged
+   * instead of treating `result` as a real turn.
+   */
+  invalid?: boolean;
 }
 
 /** Payload for summarizing an encounter. */
--- a/src/engine/modes/roleplay/encounter/encounter-service.test.ts
+++ b/src/engine/modes/roleplay/encounter/encounter-service.test.ts
@@ -1,0 +1,139 @@
+import { describe, expect, it } from "vitest";
+import type { LlmGateway } from "../../../capabilities/llm";
+import type { StorageGateway } from "../../../capabilities/storage";
+import type { EncounterActionRequest } from "../../../contracts/types/combat-encounter";
+import { resolveRoleplayEncounterAction } from "./encounter-service";
+
+type JsonRecord = Record<string, unknown>;
+
+const CHAT: JsonRecord = {
+  id: "chat-1",
+  name: "Encounter",
+  mode: "roleplay",
+  personaId: null,
+  characterIds: [],
+  connectionId: "conn-1",
+  metadata: {},
+};
+
+/** Minimal storage that resolves the chat + a default connection and returns
+ *  empty collections for everything else, so the action path runs end to end. */
+function fakeStorage(): StorageGateway {
+  return {
+    list: async <T,>() => [] as T[],
+    get: async <T,>(entity: string, id: string) => {
+      if (entity === "chats" && id === "chat-1") return CHAT as T;
+      if (entity === "connections" && id === "conn-1") return { id: "conn-1", name: "Test" } as T;
+      return null as T | null;
+    },
+    create: async <T,>() => ({}) as T,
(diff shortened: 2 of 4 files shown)


Romuromylus:
@coderabbitai review

Romuromylus:
@coderabbitai review

PR #1645, opened 2026-05-30
Surface connected command failures instead of swallowing them

### LINKED ISSUE #1535: [Issue]: Connected command failures are stripped or swallowed during generation
## Summary

Connected command execution can strip command tags and swallow command failures across shared generation, leaving the transcript without a visible retryable error.

## Expected behavior

Connected command families should be gated by mode, and command execution failures should be emitted or persisted as visible errors instead of silent no-ops.

## Actual behavior

The generation path can remove command tags while failed command execution does not produce a clear user-facing error state.

## Steps to reproduce

1. Trigger a connected command during generation that is unsupported for the active mode or fails during execution.
2. Observe the generated message and runtime error state.
3. Confirm whether the command failure is visible and retryable, or whether the command is stripped and the failure disappears.

## Impact area

- [ ] App shell, navigation, startup, or providers
- [ ] Catalog resources
- [ ] Chat mode
- [ ] Roleplay mode
- [ ] Game mode
- [ ] Shared mode UI
- [ ] Professor Mari or shell tools
- [x] React-free engine behavior, contracts, agents, or capability ports
- [ ] Shared API wrapper or remote runtime routing
- [ ] Tauri/Rust storage, assets, LLM, integrations, security, imports, or HTTP runtime
- [ ] Docs, contributor workflow, or templates

## Environment

- Branch or commit: refactor; read-only failpath audit finding
- Run mode: generation runtime; focused native repro still needed
- OS + version: not verified
- Node + pnpm versions: not relevant yet
- Remote Runtime URL configured? not verified
- Provider, model, or integration, if relevant: connected command generation path

No screenshot attached.

## Regression notes

Unknown.

## Additional context

Suggested direction: gate command families by mode and emit/persist command errors before command stripping can turn a failed command into a silent success.

## Template check

Romuromylus:
## Why this change

`persistConnectedCommandTags` wrapped `executeCommand` in a blanket `.catch(() => null)`, and `parseCharacterCommands` strips command tags from the visible content **unconditionally** (regardless of whether the command ran). So when a connected command threw, the failure collapsed to `null`, emitted no event, and left the user with a clean message and no error or retry. Two integration commands (`haptic`, `spotify`) also returned a silent `null` when their integration was absent, while their tag was still stripped from the content.

## What changed

Mirror the existing `selfie_error` event end to end, for commands:

- `generation-events.ts` / `connected-commands.ts`: add a `command_error` variant to both the `GenerationEvent` and `ConnectedCommandEvent` unions, plus an `eventsPushCommandError` helper (sibling of `eventsPushSelfieError`).
- `connected-commands.ts`: replace the blanket `.catch(() => null)` around `executeCommand` with a `try/catch` that emits `command_error` and lets generation continue (no rethrow); emit `command_error` on the `haptic`/`spotify` integration-absent branches.
- `use-generate.ts`: consume `command_error` with a toast, exactly like `selfie_error`.

The fix is **mode-neutral** — no mode flag is added to the shared helper (per `marinara-mode-separation`); roleplay/game modes that route through the same helper now get the same visible error instead of a silent strip.

## Verification

- `pnpm exec tsc -b` clean; `pnpm check:architecture` clean (events ride the existing pass-through; no new caller wiring, no React/Tauri import added to engine).
- New `connected-commands.test.ts` cases: a throwing command emits `command_error` (tag still stripped), the loop continues after a failure, an unsupported `haptic` without integrations surfaces an error, and the happy path emits no error. Full `src/engine/generation` suite (193 tests) green.

## Risk / note

The toast consumer is untested, consistent with the existing untested `selfie_error`/`illustration_error` siblings (covered by tsc + the structural mirror). A message with many failing command tags fires one toast per failure (per-command granularity, matches the existing design); aggregation is out of scope. Raw error text surfaces in the toast, parity with `selfie_error`.

## Linked issue

Closes #1535.

### DIFF
--- a/src/engine/generation/connected-commands.test.ts
+++ b/src/engine/generation/connected-commands.test.ts
@@ -145,11 +145,7 @@ describe("persistConnectedCommandTags command failures", () => {
     const chats = new Map<string, Row>([["conversation-1", conversation]]);
     const storage = storageWithChats(chats);
 
-    const result = await persistConnectedCommandTags(
-      storage,
-      conversation,
-      '[haptic: action="vibrate"]',
-    );
+    const result = await persistConnectedCommandTags(storage, conversation, '[haptic: action="vibrate"]');
 
     expect(result.events.some((event) => event.type === "command_error")).toBe(true);
     const haptic = result.events.find(
@@ -164,11 +160,7 @@ describe("persistConnectedCommandTags command failures", () => {
     const chats = new Map<string, Row>([["conversation-1", conversation]]);
     const storage = storageWithChats(chats);
 
-    const result = await persistConnectedCommandTags(
-      storage,
-      conversation,
-      "<note>[12:01] Remember the door.</note>",
-    );
+    const result = await persistConnectedCommandTags(storage, conversation, "<note>[12:01] Remember the door.</note>");
 
     expect(result.events.some((event) => event.type === "command_error")).toBe(false);
     expect(result.executedCommands).toContain("note");
--- a/src/engine/generation/connected-commands.test.ts
+++ b/src/engine/generation/connected-commands.test.ts
@@ -95,3 +95,82 @@ describe("persistConnectedCommandTags connected notes", () => {
     expect(typeof notes[1]?.consumedAt).toBe("string");
   });
 });
+
+describe("persistConnectedCommandTags command failures", () => {
+  it("emits command_error when a command throws during execution instead of swallowing it", async () => {
+    const conversation = { id: "conversation-1", mode: "conversation", notes: [] };
+    const chats = new Map<string, Row>([["conversation-1", conversation]]);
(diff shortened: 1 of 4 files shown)


Romuromylus:
Scope note: this PR surfaces command-execution exceptions (and the haptic/spotify integration-absent branches). The soft-failure `return null` paths in `executeCommand` (an `update_character`/`update_persona`/`update_lorebook` for an entity that doesn't exist, a `cross_post`/`dm` with no target chat, a `fetch` with a missing row, a `spotify` search that finds no track) still strip their tag silently. Distinguishing a genuine "target not found" from an intentional no-op is a follow-up worth its own change; I'll file it separately rather than widen this PR.

Romuromylus:
@coderabbitai review

Romuromylus:
@coderabbitai review

PR #2379, opened 2026-06-05
Flush pending storage writes on app quit

### LINKED ISSUE #2319: [Issue]: No flush-on-quit — writes within the 750ms storage debounce window are lost when the app is closed
## Summary

**Severity: Medium — Data-loss / durability.** Ordinary writes (new message, edit, swipe, settings) go through the debounced FileStorage path and sit only in the in-memory cache for up to `STORAGE_SAVE_DEBOUNCE_MS = 750ms` before reaching disk. There is no flush-on-quit anywhere in `src-tauri`, and the only Drop backstop is gated on `Arc::strong_count == 1`, which the in-flight debounce thread defeats. So quitting (or crashing) within ~750ms of the last edit silently loses that write.

## Expected behavior

On app shutdown the storage layer should synchronously flush all pending dirty collections to disk before the process exits, independent of `Drop`/`strong_count` timing.

## Actual behavior

- Writes are debounced: `write_collection` (`src-tauri/crates/storage/src/lib.rs:1718-1722`) → `cache_collection(dirty=true)` + `schedule_dirty_flush`.
- `schedule_dirty_flush` (`lib.rs:820-835`) does `let storage = self.clone();` then spawns a detached thread that sleeps 750ms before flushing — holding a cache-Arc clone alive for the whole window.
- `Drop for FileStorage` (`lib.rs:1954-1960`) only flushes when `Arc::strong_count(&self.cache) == 1`; the in-flight thread's clone keeps it ≥ 2 during exactly the pending window, so Drop skips the flush.
- No exit-time flush exists: a repo-wide search for `RunEvent` / `on_window_event` / `ExitRequested` / `WindowEvent::CloseRequested` returns zero matches; the builder ends at `.run(tauri::generate_context!())` (`src/lib.rs:281`) with no event callback. `flush()` is public but only called from backup, gallery recovery, and import rollback — never on quit.

On a normal Tauri window-close the underlying event loop typically exits the process without unwinding, so `Drop` very likely never runs at all — either path yields the same loss.

## Steps to reproduce

1. Send or edit a message (or change a setting).
2. Close the app window (or the process exits/crashes) within ~750ms, before the debounce timer fires.
3. The pending dirty collection (e.g. `messages.json`, `settings.json`) is never written; the change is gone on next launch.

## Impact area

- [x] App shell, navigation, startup, or providers

## Environment

- Branch or commit: refactor @ b0e5561d
- Run mode: Tauri desktop
- Remote Runtime URL configured? not relevant

Static source audit; cited to file:line above.

## Regression notes

Found via static audit of the refactor branch; not verified against main. Loss window is bounded (~750ms, single last action) and recoverable by re-typing, hence medium.

## Additional context

Suggested fix: switch to `.build(ctx)?` + `app.run(move |app_handle, event| { if let RunEvent::ExitRequested { .. } | RunEvent::Exit = event { if let Some(state) = app_handle.try_state::<AppState>() { let _ = state.storage.flush(); } } })`. `flush()` already takes the write lock and calls `flush_dirty_collections_locked`, so it is safe from the run callback. Optionally also flush on `WindowEvent::CloseRequested`. Keep Drop as a secondary backstop but remove its `strong_count` blind spot.

## Template check

Romuromylus:
## Linked issue

Closes #2319

## Why this change

Storage writes are debounced ~750ms before hitting disk. The Tauri app exited via `.run(generate_context!())` with no exit handler, so any write still inside the debounce window when the user closed the app was lost. The `Drop` backstop on `FileStorage` is gated on being the sole cache owner, which the in-flight debounce thread defeats during exactly that window.

## What changed

- Build the app and drive it with `.build(ctx).expect(...).run(|handle, event| ...)`, flushing storage on `RunEvent::ExitRequested` / `RunEvent::Exit` so pending writes are persisted before the process exits.
- The `FileStorage` `Drop` guard is intentionally left **unchanged**: dropping a still-shared, dirty storage (e.g. during a read-only health probe) must not eagerly persist seeded defaults; the explicit on-quit flush is the correct, scoped fix.

## Refactor impact

Primary owner: app shell + Rust storage (`src-tauri/src/lib.rs`, `src-tauri/crates/storage/src/lib.rs`)

Impact areas reviewed:

- `run()` keeps its `()` signature, so `main.rs` is untouched. `flush()` drains the dirty cache synchronously and is already used on other paths.

Boundary notes:

- The exit handler (app layer) calls `storage.flush()` (public storage-crate API). No new cross-layer coupling.

Pressure points touched:

- `src-tauri/src/lib.rs` (run loop only — command registration unchanged)

## Validation

- [ ] Matching validation command passes locally

### Manual verification notes

- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test --workspace` — green, including new `flush_persists_pending_debounced_writes_to_disk` (a dirty, not-yet-written collection lands on disk after `flush()` and the dirty flag clears). Existing `health_liveness_skips_writable_probe_by_default` still passes (Drop semantics unchanged).

## Feature Discoverability

- [x] N/A because this PR is a durability bugfix and does not add a new thing users need to find.

Reason:

- Persists existing writes on quit.

## Docs and release impact

- [x] No docs changes needed

### DIFF
--- a/src-tauri/crates/storage/src/lib.rs
+++ b/src-tauri/crates/storage/src/lib.rs
@@ -4567,6 +4567,32 @@ mod tests {
     }
 
     #[test]
+    fn flush_persists_pending_debounced_writes_to_disk() {
+        // The on-quit RunEvent handler calls storage.flush() to drain writes that are
+        // still sitting in the debounce window (#2319). Verify flush() actually lands a
+        // pending (dirty, not-yet-written) collection on disk and clears the dirty flag.
+        let root = temp_storage_root("flush-persists-pending");
+        let storage = FileStorage::new(&root).unwrap();
+        storage
+            .cache_collection("characters", &[json!({ "id": "pending" })], true)
+            .unwrap();
+        assert!(storage.dirty_collection_count() > 0, "write should be pending");
+
+        storage.flush().unwrap();
+
+        assert_eq!(
+            storage.dirty_collection_count(),
+            0,
+            "flush should clear the dirty collections"
+        );
+        // A fresh instance reads from disk, proving the pending write was persisted.
+        let reopened = FileStorage::new(&root).unwrap();
+        assert_eq!(reopened.list("characters").unwrap()[0]["id"], "pending");
+
+        fs::remove_dir_all(root).unwrap();
+    }
+
+    #[test]
     fn list_projected_caches_clean_projection_shapes_until_file_changes() {
         let root = temp_storage_root("list-projected-caches-clean-shapes");
         let storage = FileStorage::new(&root).unwrap();
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -288,7 +288,11 @@ pub fn run() {
                 tauri::RunEvent::ExitRequested { .. } | tauri::RunEvent::Exit
             ) {
                 if let Some(state) = app_handle.try_state::<crate::state::AppState>() {
-                    let _ = state.storage.flush();
+                    if let Err(error) = state.storage.flush() {
+                        // Best-effort: a failed quit-time flush must not block shutdown, but it
+                        // must not be silent either, so a dropped write is diagnosable.
+                        log::error!("failed to flush pending storage writes on quit: {error}");
+                    }
                 }
             }
         });
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -278,6 +278,18 @@ pub fn run() {
             storage_commands::update_commands::update_check,
             storage_commands::update_commands::update_apply,
         ])
-        .run(tauri::generate_context!())
-        .expect("error while running tauri application");
+        .build(tauri::generate_context!())
+        .expect("error while building tauri application")
+        .run(|app_handle, event| {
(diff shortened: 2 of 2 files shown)


Romuromylus:
@coderabbitai review

Romuromylus:
On the Bunny escalation that a cancellable shutdown still abandons a failed flush: I'm keeping this as best-effort-with-logging and not adding prevent-exit, deliberately.

Using ExitRequested's api.prevent_exit() when flush() fails would risk a strictly worse bug than the one #2319 fixes: on a persistent failure (disk full, lock poisoning, read-only volume) the user can't resolve the condition before closing, so preventing exit traps them in an app that will not quit. There's no safe 'retry once then allow' here — the next close re-fires ExitRequested, flush fails again, and exit is prevented again.

#2319's scope is the routine data loss: writes sitting in the 750ms debounce window when the app closes normally. That's fully fixed by flushing on ExitRequested/Exit. The added log::error! makes the rare hard-failure case diagnosable instead of silent, which is the right trade for a shutdown path. flush() itself is unchanged and its persistence is covered by the new unit test; the quit-event wiring isn't unit-testable without a Tauri harness, noted in the validation section.

PR #2383, opened 2026-06-05
Memoize conversation message rendering to stop the streaming re-parse storm

### LINKED ISSUE #2304: [Issue]: Streaming render storm — SplitMessageGroup defeats ConversationMessage memo, re-parsing the whole transcript ~22x/sec
## Summary

**Severity: Medium — Performance.** `SplitMessageGroup` rebuilds a fresh combined message object on every render and passes it as the `message` prop to `ConversationMessage` (a `React.memo` with the default shallow comparator). The new object reference always fails the memo, so during streaming — when the view re-renders ~22×/sec — every split assistant bubble in the mounted window re-runs its markdown parse/sanitize, even though only the streaming bubble's content actually changes.

## Expected behavior

Non-streaming message bubbles should not re-render (and re-parse markdown) on every streaming buffer commit. Only the bubble whose content is actively changing should re-render.

## Actual behavior

`ConversationView` subscribes to `streamBuffer` / `thinkingBuffer` (`src/features/modes/conversation/components/ConversationView.tsx:408-409`). The store commits the buffer every 45 ms (`STREAM_BUFFER_COMMIT_INTERVAL_MS = 45`, `use-generate.ts:90`) — ~22×/sec for the full duration of each generation — so `ConversationView` re-renders ~22×/sec.

`ConversationMessage` is `memo()` with the **default** shallow comparator (`ConversationView.tsx:338`, no custom comparator). `SplitMessageGroup` (`ConversationView.tsx:1435`) is an **unmemoized** plain function invoked inline in the render loop (`1251`). On every render it builds a new object literal:

```ts
const combinedContent = items.map((gi) => gi.msg.content).join("\n\n");
const combinedMsg = { ...firstItem.msg, content: combinedContent, extra: lastItem.msg.extra };
return (<ConversationMessage key={firstItem.key} message={combinedMsg} ... />);
```

The new `combinedMsg` reference fails the child memo, forcing a re-render. The re-render is genuinely costly: `MessageContent` (`180`) is unmemoized and calls `renderMarkdownBlocks` (a full line-by-line regex markdown→ReactNode parse, `markdown.tsx:355`) directly in its render body (`214`) with no `useMemo`; the `renderedContent` memo (`484`) only caches the macro-resolved *string*, not the markdown parse.

In Discord-style conversation mode, multi-paragraph assistant replies are split into `__blockN` groups (`817-843`), so most visible bubbles route through `SplitMessageGroup`. The mounted window holds up to 160 messages (`MAX_MOUNTED_TRANSCRIPT_MESSAGES = 160`), so the bulk of the transcript re-parses markdown ~22×/sec for the entire duration of every reply, scaling linearly with on-screen message count. The single-message path (`1295-1318`) passes the stable `msg` reference, so its memo correctly bails — confirming the split-group path is the leak.

## Steps to reproduce

1. Open a conversation-mode chat with several multi-line assistant messages already on screen.
2. Send a message and let the assistant stream a reply.
3. Profile renders: every `SplitMessageGroup`-rendered bubble in the window re-renders/re-parses ~22×/sec, not just the streaming one.

## Impact area

- [x] Chat mode
- [x] Shared mode UI

## Environment

- Branch or commit: refactor @ 47b16701
- Run mode: Tauri desktop (same React frontend in server/web shell)

Static source audit; claims cited to file:line above.

## Regression notes

Found via static audit of the refactor branch; not verified against main.

## Additional context

Suggested fix: wrap `SplitMessageGroup` in `React.memo` and memoize `combinedMsg` / `combinedContent` with `useMemo` keyed on `items` (only recompute when item ids/content change); memoize `MessageContent` so the markdown parse only re-runs when content changes; and ideally gate `streamBuffer` / `thinkingBuffer` so non-streaming groups never receive live streaming props at all.

## Template check

Romuromylus:
## Linked issue

Closes #2304

## Why this change

While a message generates, `ConversationView` re-renders on every ~45ms stream-buffer commit (~22x/sec). `MessageContent` ran `renderMarkdownBlocks` directly in its render body and was not memoized, so every mounted message (up to the full render window) re-parsed its entire markdown on each commit — the streaming render storm.

## What changed

- Wrap `MessageContent` in `React.memo` and compute the parsed output in a `useMemo` keyed on `content` / `mentionNames`, so `renderMarkdownBlocks` only re-runs for a message whose content actually changed.
- Stabilize the `onImageOpen` callback (`useCallback`) so the memo can bail.
- The `useMemo` runs unconditionally before the image-URL early return, keeping the Rules of Hooks intact.

## Refactor impact

Primary owner: chat / conversation mode UI (`src/features/modes/conversation/components/ConversationMessage.tsx`)

Impact areas reviewed:

- `MessageContent` is conversation-mode-scoped; `renderedContent` (macro-resolved string) and `mentionNames` are already memoized upstream, so the memo bails for unchanged messages during streaming.

Boundary notes:

- none — no cross-layer imports; pure render optimization.

Pressure points touched:

- none

## Validation

- [ ] Matching validation command passes locally

### Manual verification notes

- `pnpm vitest run src/features/modes/conversation/components/MessageContent.test.tsx` — 3 passed (no re-parse on unchanged content across re-renders with a new `onImageOpen` identity; re-parse on content change; image URL bypasses parse).
- `pnpm exec tsc -b` — clean. `pnpm exec eslint` on the touched files — clean.

## Feature Discoverability

- [x] N/A because this PR is a performance bugfix and does not add a new thing users need to find.

Reason:

- Render optimization only.

## Docs and release impact

- [x] No docs changes needed

### DIFF
--- a/src/features/modes/conversation/components/ConversationMessage.tsx
+++ b/src/features/modes/conversation/components/ConversationMessage.tsx
@@ -177,7 +177,7 @@ function highlightMentions(nodes: ReactNode[], names: string[], keyPrefix: strin
 }
 
 /** Renders message content, showing image URLs as inline images */
-function MessageContent({
+export const MessageContent = memo(function MessageContent({
   content,
   mentionNames,
   onImageOpen,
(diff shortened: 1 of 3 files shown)


Romuromylus:
Scope note: this PR eliminates the dominant cost #2304 measured — `renderMarkdownBlocks` re-running across the transcript ~22x/sec — by memoizing both the parse (`MessageContent`) and the combined-message object (`SplitMessageGroup`). Fully preventing `ConversationMessage` from re-rendering at all during streaming additionally requires memoizing the parent's per-group element list and `characterMap` in `ConversationView` (the `items` arrays are rebuilt every render), which is a larger structural change I'm leaving as a follow-up rather than bundling into this fix.

Romuromylus:
@coderabbitai review

PR #2399, opened 2026-06-06
Replace atomic-update deep-equality conflict check with cheap file-stamp guard

### LINKED ISSUE #2301: [Issue]: Every message create/edit/swipe deep-clones and deep-compares the entire global messages + message-swipes collections
## Summary

**Severity: High — Performance.** Every message mutation (create, edit, add/switch/delete swipe, branch copy) routes through `write_message_and_swipes -> Storage::update_collections_atomically`, which reads, deep-clones, and deep-**compares** the entire **global** `messages` and `message-swipes` collections. Cost scales with the *total* number of messages/swipes across every chat, not the active chat, and runs under the storage write lock on every turn.

## Expected behavior

A single message create/edit/swipe should cost work proportional to the active chat (or to the changed rows), not to the entire message corpus. The conflict check in particular should not deep-compare the whole collection.

## Actual behavior

`update_collections_atomically` (`src-tauri/crates/storage/src/lib.rs:604-663`), invoked for `["messages", "message-swipes"]` by `write_message_and_swipes` (`commands/storage/message_swipes.rs:359-390`), performs, per single mutation, over the global collections (N = all messages, M = all swipes), all under the write lock:

- `read_collection_no_recovery` returns `cached.rows.clone()` — full deep-clone of N (+M). A cache hit does **not** avoid the clone (`cached_rows`, `lib.rs:730-740`, always `.clone()`s the whole Vec).
- `entry.rows.clone()` into `original_rows` — deep-clone #2.
- Conflict check: `read_collection_no_recovery(collection)? != *rows` — deep-clone #3 **plus** a full `serde_json::Value` structural deep-equality over the entire N (+M):

```rust
for (collection, rows) in &original_rows {
    if self.read_collection_no_recovery(collection)? != *rows {
        return Err(AppError::new("storage_conflict", ...));
```

- `entry.rows.clone()` into `replacements` — deep-clone #4.
- `replace_all_many_locked` (`lib.rs:1902`) — `serde_json::to_vec_pretty(rows)` full re-serialize of both collections.

So roughly **4 full deep-clones + 1 full deep-equality + 1 full serialize** of the entire global messages collection and the entire swipes collection, per mutation.

Hot path confirmed: `create_message` (`message_swipes.rs:831/841/846`), `set_active_swipe`, patch/edit (874), branch copy (911), and deletes (759/798) all route through `write_message_and_swipes`. `messages` / `message-swipes` are **single global collections** (one in-memory `Vec<Value>` across all chats; rows carry a `chatId` only for in-memory filtering) — there is no per-chat sharding, so cost scales with the total corpus. The append fast-path in `create_locked` (`lib.rs:344-351`) is dead for message creation: `create_message` never calls `create_locked`, and after the first chat load the collection is already cached.

## Steps to reproduce

1. Build up a corpus of several thousand messages across chats.
2. Send a message, add/switch/delete a swipe, or edit a message.
3. Observe each mutation pays multiple full-corpus clones + a full deep-equality pass (latency grows with total message count, not the active chat).

## Impact area

- [x] Chat mode
- [x] Roleplay mode
- [x] Game mode

## Environment

- Branch or commit: refactor @ 47b16701
- Run mode: Tauri desktop (same storage layer in server mode)

Static source audit; claims cited to file:line above.

## Regression notes

Found via static audit of the refactor branch; not verified against main.

## Additional context

Two fix directions:
1. **Cheap conflict guard** — replace the O(n) `read_collection_no_recovery != original_rows` deep-equality with the already-computed `CollectionFileStamp.content_signature` (a version/stamp guard), eliminating the most wasteful step without changing the storage format.
2. **Per-chat sharding** — store messages/swipes one collection (file) per `chatId` so reads/writes/compares scale with the active chat. Larger change, addresses the clone cost as well.

## Template check

Romuromylus:
## Linked issue

Closes #2301

## Why this change

- `update_collections_atomically` ran on every message create/edit/swipe. Its concurrent-modification guard cloned the entire global `messages` + `message-swipes` collections and deep-compared them by structural `serde_json::Value` equality (O(n) over the whole corpus) under the storage write lock — pure overhead that scales with chat history size.

## What changed

- `src-tauri/crates/storage/src/lib.rs`: the guard now captures a cheap `CollectionFileStamp` (file length + mtime + content hash) per target collection at load and re-checks it at commit, instead of cloning and deep-comparing all rows. The `CollectionFileStamp`/`collection_file_stamp` primitive already exists in the file and is used by the projected-list cache.
- The pre-load `flush_dirty_collections_locked()` plus the atomic-update mutex mean targets are clean on disk and only an external/concurrent write changes a stamp, so the `storage_conflict` contract is preserved. Public method signature unchanged — callers (message_swipes.rs / chats.rs / entities.rs) untouched.

## Refactor impact

Primary owner: Rust storage

Impact areas reviewed:

- `src-tauri/crates/storage/src/lib.rs` (`update_collections_atomically`)

Boundary notes:

- None. Private method body in the `marinara-storage` crate; no mode-specific logic (shared infra). Per-chat sharding (issue's larger alternative) intentionally out of scope.

Pressure points touched:

- None.

## Validation

- [x] Matching validation command passes locally

### Manual verification notes

- `cargo test --manifest-path src-tauri/Cargo.toml -p marinara-storage` — 45 passed, including all existing conflict/recovery tests (`rejects_duplicate`, `rejects_reentrant_writes`, `reads_targets_without_recovery`, `reentrant_reads_do_not_recover`, `reads_and_replaces_multiple_collections`). Local inline tests confirmed the cheap guard still rejects a genuine out-of-band modification (`storage_conflict`) and does not false-positive on a normal sequential write (kept local per repo convention).
- `cargo clippy --manifest-path src-tauri/Cargo.toml --workspace --all-targets -- -D warnings` — clean.

## Feature Discoverability

- [ ] Updated `src/features/shell/discovery/`
- [x] N/A because this PR is a performance bugfix in storage internals.

Reason:

- No user-facing surface change.

## Docs and release impact

- [x] No docs changes needed

### DIFF
--- a/src-tauri/crates/storage/src/lib.rs
+++ b/src-tauri/crates/storage/src/lib.rs
@@ -634,10 +634,13 @@ impl FileStorage {
             loaded
         };
 
-        let original_rows = entries
+        let original_stamps = entries
             .iter()
-            .map(|entry| (entry.collection.clone(), entry.rows.clone()))
-            .collect::<Vec<_>>();
+            .map(|entry| {
+                let path = self.collection_path(&entry.collection)?;
+                Ok((entry.collection.clone(), collection_file_stamp(&path)?))
+            })
+            .collect::<AppResult<Vec<_>>>()?;
 
         let output = update(&mut entries)?;
 
@@ -646,8 +649,9 @@ impl FileStorage {
             .write()
             .map_err(|_| AppError::new("lock_error", "Storage lock poisoned"))?;
         self.flush_dirty_collections_locked()?;
-        for (collection, rows) in &original_rows {
-            if self.read_collection_no_recovery(collection)? != *rows {
+        for (collection, original_stamp) in &original_stamps {
+            let path = self.collection_path(collection)?;
+            if collection_file_stamp(&path)? != *original_stamp {
                 return Err(AppError::new(
                     "storage_conflict",
                     format!("Collection changed during atomic update: {collection}"),
--- a/src-tauri/crates/storage/src/lib.rs
+++ b/src-tauri/crates/storage/src/lib.rs
@@ -610,7 +610,12 @@ impl FileStorage {
         F: FnOnce(&mut [AtomicCollectionRows]) -> AppResult<T>,
     {
         let _atomic_update = self.begin_atomic_update()?;
-        let mut entries = {
+        // Load the rows and capture each collection's file stamp under the SAME
+        // write lock, so the conflict baseline reflects exactly the bytes the rows
+        // were read from. Sampling the stamp after the lock is released would let a
+        // concurrent writer slip in between the read and the stamp, baking its change
+        // into the baseline and hiding it from the commit-time conflict check.
+        let (mut entries, original_stamps) = {
             let _guard = self
                 .lock
                 .write()
@@ -618,10 +623,11 @@ impl FileStorage {
             self.flush_dirty_collections_locked()?;
 
             let mut loaded = Vec::with_capacity(collections.len());
+            let mut original_stamps = Vec::with_capacity(collections.len());
             let mut seen_paths = HashSet::new();
             for collection in collections {
                 let path = self.collection_path(collection)?;
-                if !seen_paths.insert(path) {
+                if !seen_paths.insert(path.clone()) {
                     return Err(AppError::invalid_input(format!(
                         "Duplicate collection update: {collection}"
                     )));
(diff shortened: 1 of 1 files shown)


Romuromylus:
This re-flag is a false positive on the current head. The baseline stamp is captured **inside** the same `write()` lock as the row read, not after it. In `update_collections_atomically`, the `let (mut entries, original_stamps) = { let _guard = self.lock.write()...; self.flush_dirty_collections_locked()?; for collection in collections { ... loaded.push(read_collection_no_recovery(...)?); original_stamps.push((..., collection_file_stamp(&path)?)); } ... }` block holds `_guard` for the entire load-and-stamp loop, so no concurrent writer can interleave between a collection's row read and its stamp. The pre-loop `flush_dirty_collections_locked()` guarantees the on-disk file equals the cached rows, so the stamp matches exactly the bytes the rows were read from. The only window that remains is between releasing the load lock and re-acquiring it at commit, which is the legitimate conflict window the commit-time re-check detects. No change needed.

PR #2401, opened 2026-06-06
Cap TTS provider response while streaming to prevent OOM

### LINKED ISSUE #2324: [Issue]: TTS speak() buffers the entire provider response into memory before enforcing the 20MB cap
## Summary

**Severity: Low — Unbounded buffering / DoS.** TTS `speak()` reads the entire provider response body into memory with `response.bytes().await` before the 20 MB cap is checked, so the cap cannot prevent a memory blow-up. A misbehaving/hostile TTS endpoint can stream a multi-GB body (within the up-to-30-minute request timeout) and OOM the process before the limit fires.

## Expected behavior

The body should be streamed with a running byte counter and aborted once `MAX_TTS_AUDIO_BYTES` (20 MB) is exceeded, and/or a `Content-Length` pre-check applied before downloading — mirroring the established pattern in `custom_tools.rs`.

## Actual behavior

`src-tauri/src/commands/storage/integrations/tts.rs:425-447`:

```rust
let bytes = response
    .bytes()
    .await
    .map_err(|error| AppError::new("tts_response_error", error.to_string()))?;
...
if bytes.len() > MAX_TTS_AUDIO_BYTES {
    return Err(AppError::invalid_input("TTS audio response is too large"));
}
```

The whole body is buffered first; the cap at line 445 runs after. There is no `Content-Length` pre-check and no streaming counter. The request timeout is configurable up to 30 minutes (`MAX_TTS_REQUEST_TIMEOUT_MS = 30 * 60_000`). The codebase already does this correctly elsewhere: `custom_tools.rs:287-300` has a `content_length()` pre-check plus a `.chunk()` loop with a running counter; the TTS path has neither.

## Steps to reproduce

1. Configure a TTS source pointing at an endpoint that returns a multi-GB body with an audio content-type.
2. Call `tts/speak`.
3. The full body buffers into memory before the 20 MB check; large enough bodies OOM the process.

## Impact area

## Environment

- Branch or commit: refactor @ b0e5561d
- Run mode: Tauri desktop (also server mode)
- Provider, model, or integration, if relevant: TTS source

Static source audit; cited to file:line above.

## Regression notes

Found via static audit of the refactor branch; not verified against main. TTS config comes from stored app-settings (not the per-request body), so exploitation requires a hostile/broken user-chosen endpoint (desktop) or a post-auth config write (server), hence low.

## Additional context

Suggested fix: add a `response.content_length()` pre-check before download and replace `bytes().await` with a `.chunk()` loop carrying a running counter that aborts once `MAX_TTS_AUDIO_BYTES` is exceeded, mirroring `custom_tools.rs`. (`images/providers.rs:474` shares the same weakness and could be fixed alongside.)

## Template check

Romuromylus:
## Linked issue

Closes #2324

## Why this change

- TTS `speak()` fully buffered the provider response into memory via `response.bytes().await` and only enforced the 20MB cap afterward. A hostile or misconfigured endpoint could stream a multi-GB body (with the request timeout up to 30 min) and OOM the process before the cap ever ran.

## What changed

- `src-tauri/src/commands/storage/integrations/tts.rs`: replaced `bytes().await` with a new `read_capped_audio_bytes` helper that does a `content_length()` pre-check against `MAX_TTS_AUDIO_BYTES` and then streams via a `.chunk()` loop with a saturating running counter, aborting the moment the cap is exceeded. Mirrors the in-repo `read_limited_webhook_text` pattern in `custom_tools.rs`. The post-buffer size check is removed (now enforced during streaming); the same `invalid_input` "TTS audio response is too large" message is preserved on both reject paths.

## Refactor impact

Primary owner: Rust storage (integrations)

Impact areas reviewed:

- `src-tauri/src/commands/storage/integrations/tts.rs` (`speak`, new `read_capped_audio_bytes`)

Boundary notes:

- None. Same-file integration change; no IPC/command-signature/return-shape change. Valid (<20MB) responses behave identically. Sibling call sites sharing the same anti-pattern (http.rs, images/providers.rs) are intentionally left out of scope (no drive-by refactor) and can be addressed separately.

Pressure points touched:

- None.

## Validation

- [x] Matching validation command passes locally

### Manual verification notes

- `cargo test --manifest-path src-tauri/Cargo.toml -p marinara-engine tts` — 13 passed, including local-only tests for the `content_length` pre-check (declared-41MB body rejected before download) and the streamed-overflow abort (no Content-Length, >20MB stream aborts); happy path unchanged.
- `cargo clippy --manifest-path src-tauri/Cargo.toml --workspace --all-targets -- -D warnings` — clean.

## Feature Discoverability

- [ ] Updated `src/features/shell/discovery/`
- [x] N/A because this PR is a resource-safety bugfix.

Reason:

- No user-facing surface change.

## Docs and release impact

- [x] No docs changes needed

### DIFF
--- a/src-tauri/src/commands/storage/integrations/tts.rs
+++ b/src-tauri/src/commands/storage/integrations/tts.rs
@@ -422,10 +422,7 @@ async fn speak(state: &AppState, body: Value) -> AppResult<Value> {
         .and_then(|value| value.to_str().ok())
         .unwrap_or(fallback_content_type)
         .to_string();
-    let bytes = response
-        .bytes()
-        .await
-        .map_err(|error| AppError::new("tts_response_error", error.to_string()))?;
+    let bytes = read_capped_audio_bytes(response).await?;
     if !status.is_success() {
         let detail = provider_error_detail(&String::from_utf8_lossy(&bytes), 500);
         return Err(AppError::with_details(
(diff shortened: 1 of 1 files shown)


Romuromylus:
Good catches, both stem from the error-classification harden I added on top of the core fix. Since this PR's actual job (issue #2324) is the streaming OOM cap, I have reverted that harden and kept only the cap. The remaining behavior (an oversized body propagating the cap's `invalid_input` regardless of status) is the original narrow edge: it requires a provider to stream a >20MB body, which for a non-2xx error page essentially never happens, and a genuinely oversized 2xx audio payload really is too large. Reclassifying that cleanly (without mislabeling mid-stream network errors as "too large", which was the LOW you flagged) is more surface than the edge warrants, so I am leaving it as-is. The OOM protection, which is the load-bearing fix, is unchanged.

PR #2540, opened 2026-06-09
perf(db): make the file-backed store flush async so it stops blocking the event loop

### LINKED ISSUE #2528: [Issue]: File-store flush synchronously serializes whole table + copies file to .bak on the event loop
## Summary

**Severity: Medium — performance.** Each file-store flush synchronously serializes the entire dirty table and `copyFileSync`s the whole previous file to a `.bak`, plus several `fsyncSync` calls, all on the Node event loop. `messages` is a single global table holding every message across every chat, so on a large transcript each flush blocks the loop for a multi-MB serialize + multi-MB file copy, stalling concurrent requests on the shared server. Cost grows linearly with accumulated chat history.

## Expected behavior

Flush should not block the event loop with `O(total DB size)` synchronous work on a hot write path. The serialize + backup copy + fsync should run via async `fs/promises` (so the flush yields), and/or the whole-file backup copy should be replaced with rename-based double-buffering.

## Actual behavior

The 750 ms debounce correctly prevents a write *per op* (the naive "rewrites whole file per op" claim is false — `saveFileSnapshots` only writes dirty tables). But each flush that *does* fire pays `O(rows in the dirty table)` synchronously. `atomicWriteFile` (`packages/server/src/db/file-backed-store.ts:329-376`, driven by `saveFileSnapshots` at `:1226`):

```ts
if (refreshBackup && existsSync(path) && !looksNulFilled(path)) {
  copyFileSync(path, bakTmpPath);   // whole-file copy, synchronous
  flushFile(bakTmpPath);            // fsyncSync
  renameSync(bakTmpPath, bakPath);
}
writeFileSync(tmpPath, content);     // content = JSON.stringify(entire table), synchronous
flushFile(tmpPath);                  // fsyncSync
renameSync(tmpPath, path);
flushFile(dirname(path));            // fsyncSync on dir
```

`messages` is a single global table (`:154`); `rows("messages")` returns the entire in-memory array, and `markDirty("messages")` fires on every insert/update/delete. So during active chatting each debounced flush re-serializes and whole-file-copies the entire messages table. For a long transcript (multi-MB `messages.json`), that is a multi-MB synchronous serialize + multi-MB synchronous copy + multiple fsyncs blocking the loop for every other user's request, growing with one user's chat history.

## Steps to reproduce

1. Accumulate a large `messages` table (a long transcript or many chats; multi-MB `messages.json`).
2. Chat actively (or run autonomous agents writing per turn) so flushes fire on the debounce.
3. Observe each flush stringify + whole-file-copy the entire table synchronously; on a multi-user server, one user's flush blocks the event loop for all others.

## Environment

- Marinara Engine version: 1.6.1 (`staging @ a9fdae8`)
- Install type: source
- OS + version: n/a — static source audit
- Browser or app shell: n/a

Static source audit, cited to file:line above.

## Additional context

Fix: swap the `*Sync` fs calls in `atomicWriteFile` for `fs/promises` async variants so the flush yields the event loop, and/or replace the whole-file `copyFileSync` backup with rename-based double-buffering (keep the just-replaced file as the backup) so each flush is one serialize + one write rather than copy-then-write. Partitioning the hot `messages` table per chat would cap serialize/copy cost to the active chat, but that is a larger change.

Related: this compounds with the updater-flush data-loss issue (the same store), but is a distinct cost (synchronous flush vs missing flush).

Regression notes: found via static audit; `file-backed-store.ts` is byte-identical on `main` (`65a0395`, `git diff` empty), so this is a shared-engine performance characteristic, not a `main`→`staging` regression.

## Template check

Romuromylus:
## Linked issue

Closes #2528

## Why this change

The file-backed store's flush path was fully synchronous and ran O(total table size) work on the Node event loop:

```
markDirty() -> flush() -> saveFileSnapshots() -> atomicWriteFile() per dirty table
```

`atomicWriteFile` did `copyFileSync` (a whole-file `.bak` copy), `writeFileSync`, and three `fsyncSync` calls. `messages` is a single global file-backed table that `markDirty`s on every insert/update/delete, so during active chat each debounced flush synchronously re-serialized **and** whole-file-copied the entire `messages` table (multi-MB on a long transcript), blocking every other request on the shared server. The cost grows linearly with accumulated chat history; the 750ms debounce only prevents a write *per op*, not the O(table size) cost *per flush*.

## What changed

`packages/server/src/db/file-backed-store.ts`:
- Converted `flushFile` and `atomicWriteFile` to `fs/promises` (`open` + `FileHandle.sync()`, `copyFile`, `writeFile`, `rename`, `unlink`).
- Made `saveFileSnapshots` async and `await` each `atomicWriteFile` (sequential — preserves ordering and bounds concurrent handles to one, matching prior behavior).
- `await` the now-async `saveFileSnapshots()` in the already-async `flush()`.
- Dropped the now-unused sync `fs` imports (`copyFileSync`/`fsyncSync`/`writeFileSync`/`renameSync`/`unlinkSync`), keeping the ones still used by `looksNulFilled`/`loadFileSnapshots`/etc.

**Crash-safety is preserved exactly**: same tmp-path naming, same `.bak` refresh gating (`refreshBackup && existsSync && !looksNulFilled`), same fsync-before-rename on the `.bak` tmp / main tmp / parent dir, same cleanup-tmp-on-error then rethrow. The only behavioral change is that the multi-MB serialize→copy→fsync now yields the event loop instead of blocking it. `JSON.stringify(rows)` stays synchronous (the deeper per-chat-partitioning win is out of scope here).

All three `atomicWriteFile`/`saveFileSnapshots` call sites are awaited; `flush()`'s fire-and-forget callers (markDirty debounce, safety timer, beforeExit handler, controller) already used `void this.flush()` and still return a Promise, so no signature change is needed at any of them.

Not a regression: byte-identical on `main`; shared-engine perf characteristic.

## Validation

### Manual verification notes

`pnpm check` passes locally. Added a local-only node:test (`file-backed-store-flush.test.ts`, gitignored) driving the store through `createFileNativeDB` against a temp `FILE_STORAGE_DIR`:
- **Roundtrip**: insert a row, `flush()` returns a Promise, the table JSON contains the row, and no `.tmp-*` files remain (atomic rename completed).
- **Backup double-buffer**: a second write of the pre-existing table produces a valid `.bak` holding the prior version (proves `copyFile → fsync → rename` works through `fs/promises`).
- **Idempotency**: a flush with nothing newly dirty does not throw and does not rewrite/corrupt the file.

Local run: **3/3 pass**.

Still to verify manually against the deploy (non-regression smoke, not the gate): create a character + chat, send several messages, refresh and restart the service, and confirm all data persists (durable flush on shutdown via `close() → flush(true)`), with `tables/messages.json` valid and a `.bak` present after activity.

## Docs and release impact

- [x] No docs changes needed


Romuromylus:
Self-review on this branch flagged a durability narrowing the async conversion introduced: `flush()` cleared `this.dirtyTables` *after* awaiting the write, so a `markDirty()` that interleaved during the per-table IO (on a table already serialized this pass) could be erased by the post-await `clear()` and stay unpersisted until the next write to that table. The synchronous version had no such window.

Pushed a follow-up that snapshots `this.dirtyTables` and resets it to a fresh Set *before* the await, passes the snapshot into `saveFileSnapshots`, and merges it back on failure so failed tables retry. Added a local regression test covering a write issued while a flush is in flight (passes).

PR #2542, opened 2026-06-09
perf(lorebook): bound the default per-turn keyword-scan window

### LINKED ISSUE #2526: [Issue]: Lorebook keyword scan re-scans full chat history every turn (O(history × entries))
## Summary

**Severity: Medium — performance.** On every generation turn, the lorebook keyword scanner concatenates the entire chat history into one string and runs a keyword test over it once per active entry, synchronously on the Node event loop before the model call. Cost is `O(total history chars × active entry count)` and grows linearly with both chat length and lorebook size, because the global scan depth is `0` ("scan all messages") and the per-chat context limit is off by default.

## Expected behavior

The lorebook keyword scan window should be bounded independently of the prompt context window — scan only the last N messages (a SillyTavern-style global scan depth) so per-turn cost is `O(scanDepth × entryCount)`, constant in chat length. Keys typically appear in recent turns, and per-entry scan-depth overrides plus constant/sticky entries already exist, so a sane default would not change activation for the common case.

## Actual behavior

`packages/server/src/services/lorebook/keyword-scanner.ts:410-411` builds the scan text from the full history when `scanDepth` is 0, and the per-entry loop (`:426-496`) tests each non-constant entry against it:

```ts
const messagesToScan = scanDepth > 0 ? messages.slice(-scanDepth) : messages;
const combinedText = messagesToScan.map((m) => m.content).join("\n");
```

The driver passes `scanDepth: 0` (`packages/server/src/services/lorebook/index.ts:844-845`, `// Scan all messages`), and `contextMessageLimit` is off by default (`generate.routes.ts:2017-2019`), so the scan covers the entire conversation. `processLorebooks` runs per turn (`generate.routes.ts:3617/3673/4896`). The `MAX_LOREBOOK_ENTRIES` cap (100) bounds budget *selection*, not the keyword scan, and there is no memoization of the activated set across swipes/regenerates — so a long roleplay with a large worldbook re-runs ~100 full-history substring/regex scans every turn, blocking the event loop and stalling other users on the shared server.

## Steps to reproduce

1. Open a long chat (e.g. a months-old roleplay with thousands of messages) that has many active keyword lorebook entries and no `contextMessageLimit` set.
2. Send a message / regenerate.
3. Each turn re-scans the full history against every active entry before the model call; the per-turn cost grows as the chat lengthens.

## Environment

- Marinara Engine version: 1.6.1 (`staging @ a9fdae8`)
- Install type: source
- OS + version: n/a — static source audit
- Browser or app shell: n/a

Static source audit, cited to file:line above.

## Additional context

Fix: give the global scan a sane default depth (e.g. 2–10) at `index.ts:844-845` instead of `scanDepth: 0`, rather than relying on the off-by-default `contextMessageLimit`. Per-entry `scanDepth` overrides already exist (`keyword-scanner.ts:457-463`) and constant/sticky entries bypass keyword matching, so a default depth bounds per-turn cost to `O(scanDepth × entryCount)` without changing common-case activation. Alternatively, memoize the activated-entry set per `(latest-message-id, entry-set-hash)` so unchanged history is not re-scanned on every regenerate/swipe.

Regression notes: found via static audit; `keyword-scanner.ts` is byte-identical on `main` (`65a0395`) and the `scanDepth: 0` driver/default-off `contextMessageLimit` are the same on both branches, so this is shared baseline behavior, not a `main`→`staging` regression.

## Template check

Romuromylus:
## Linked issue

Closes #2526

## Why this change

The generate-path lorebook scan set `scanOpts.scanDepth = 0` ("scan all messages"). In `keyword-scanner.ts`, an entry whose effective `scanDepth` resolved to `null` fell back to `combinedText`, which — with the global `scanDepth` of 0 — was the **entire** chat history. So each turn re-scanned the full transcript for those entries (O(history × entries)), and the cost grew with transcript length.

The common case is already bounded: a lorebook's `scanDepth` DB-defaults to 2 and `applyLorebookDefaults` propagates it to null-`scanDepth` entries before the scan, so those slice to the last 2 messages. The unbounded full-history scan only fires for an entry whose effective depth is genuinely unset (`null`) **or** an explicit `0`.

## What changed

> **Maintainer note / wants your eyes:** this touches user-facing scan semantics, so I kept it deliberately conservative. Two things to sanity-check: (1) the default value `10`, and (2) the rule below that an explicit `scanDepth` of `0` is preserved as "scan all". Happy to adjust the constant or the policy.

- Added `LIMITS.LOREBOOK_DEFAULT_SCAN_DEPTH = 10` (`packages/shared/src/constants/defaults.ts`).
- `packages/server/src/services/lorebook/index.ts`: the generate-path `scanOpts.scanDepth` now uses `LIMITS.LOREBOOK_DEFAULT_SCAN_DEPTH` instead of `0`, so an entry that leaves `scanDepth` unset caps its scan to the last 10 messages instead of the whole history.
- `packages/server/src/services/lorebook/keyword-scanner.ts`: the per-entry scan-window selection now distinguishes the three cases explicitly:
  - `scanDepth === 0` → **scan full history** (a deliberate "scan all" choice, preserved),
  - `scanDepth > 0` → scan that many recent messages (unchanged),
  - `scanDepth === null` → inherit the now-bounded global default.

So a user who intentionally sets a lorebook/entry to `scanDepth 0` is unaffected; only the genuinely-unset fallback is bounded. The knowledge-router scan path passes no `scanDepth` and keeps the scanner's own default, so it is unchanged.

Not a regression: byte-identical on `main`; long-standing shared-code perf characteristic.

## Validation

### Manual verification notes

`pnpm check` passes locally. Added a local-only node:test (`scan-depth.test.ts`, gitignored) driving `scanForActivatedEntries` directly with a keyword present only in the oldest of 15 messages and the new bounded global default:
- null entry `scanDepth` → keyword 10+ messages back is **not** scanned (the fix),
- explicit entry `scanDepth 0` → still reaches the oldest message (deliberate scan-all preserved),
- entry `scanDepth` larger than history → scans everything (override honored),
- a keyword inside the bounded window still activates (negative control).

Local run: **4/4 pass**.

Still to verify manually against the deploy:
- A normal chat where recent-turn lorebook entries still activate as before.
- A lorebook/entry explicitly set to `scanDepth 0` still matches keywords far back in history.

## Docs and release impact

- [x] No docs changes needed


Romuromylus:
Small correction to the description: the knowledge-router scan path passes an explicit global `scanDepth: 0` (scan-all), which routes through the unchanged global-option logic, rather than passing no `scanDepth`. The behavioral claim stands (that path is unaffected) - only the wording was imprecise. The default value of 10 is still worth a maintainer sanity check, as flagged in the body.

PR #2648, opened 2026-06-19
Fix game-mode voice auto-play overlap (#2647)

### LINKED ISSUE #2647: [Issue]: Game-mode voice auto-play loops/duplicates the same line until stopped
## Summary

In **game (GM) mode** with voice auto-play enabled, the narrated voice line keeps restarting/duplicating — the same line plays over and over (audibly overlapping with itself) until playback is manually stopped.

## Expected behavior

Each segment/side-line voice clip auto-plays once when it becomes ready, then advances to the next. A line should never restart from the top on its own.

## Actual behavior

While a scene's voice clips are still being generated, the active line restarts repeatedly and overlaps itself, continuing until the user hits stop.

## Root cause (analysis)

This is specific to the game-mode voice queue in `packages/client/src/components/game/GameNarration.tsx`. The roleplay/VN path uses the `ttsService` singleton (`packages/client/src/lib/tts-service.ts`), which commits its dedup synchronously via `stop()` + a `sequence` counter and is not affected.

The "already auto-played" guards are written **asynchronously** but read on a **synchronous churn signal**, so the guard is perpetually stale:

- The main-line auto-play effect (`GameNarration.tsx:2866`) only sets its dedup ref `lastAutoPlayedVoiceKeyRef` inside the `onStarted` callback, which fires when the browser actually begins playback (`audio.onplaying` / `play().then(...)`) — tens of ms later.
- That effect (and the side-voice effect at `GameNarration.tsx:2890`) depend on `gameVoiceVersion`.
- `gameVoiceVersion` is bumped every time a voice blob finishes generating (`GameNarration.tsx:2782`), plus on side-voice start/block/retry (`2918`, `2929`, `1852`).

So while clips are still generating, `gameVoiceVersion` increments rapidly and re-runs the auto-play effect. If the current clip's `onStarted` hasn't fired yet, the dedup ref is still stale, so the effect calls `playGameVoiceKey(activeVoiceKey)` again → `playGameVoiceKeys()` runs `stopGameVoicePlayback()` and restarts the same line from the top (`GameNarration.tsx:1751`–`1826`). Generation churn faster than playback-start ⇒ continuous restart, and the stop/restart race on the `<audio>` element is the audible overlap. The side-voice effect has the same shape and even bumps the version itself in `onStarted` (`2918`), feeding the loop.

## Proposed fix

Close the async gap by marking a key as "attempt in flight" **synchronously** when `playGameVoiceKey`/`playGameVoiceKeys` is invoked (a ref the auto-play effects also check), then:
- on `onStarted` → commit to the real dedup ref/Set and clear in-flight;
- on `onBlocked` → clear in-flight so the existing autoplay-blocked retry path still works (dedup was deferred to `onStarted` originally to allow retry-on-block, so a purely synchronous commit would regress that; the in-flight ref preserves it).

## Steps to reproduce

1. Enable a TTS provider and turn on game-mode voice auto-play.
2. Start/advance a game-mode scene that produces multiple voice segments and/or side-lines (so several clips generate in sequence).
3. Observe the active voice line restart and overlap itself repeatedly until stopped.

## Environment

- Marinara Engine version: 1.6.1 (also present on `staging`)
- Install type: source
- OS + version: Linux (WSL2)
- Browser or app shell: web client

## Additional context

Discovered while investigating per-character voice playback. Affects game mode only; roleplay/VN/conversation auto-play (the `ttsService` singleton path) is unaffected. Fix to be made against `staging` per `CONTRIBUTING.md`.

## Template check

aeriondyseti:
## Why this change

Closes #2647.

In **game (GM) mode** with voice auto-play enabled, narrated voice lines could restart and duplicate themselves until manually stopped. The overlapping playback made game-mode TTS unusable for scenes with live voice generation.

## What changed

- Game voice playback now tracks every live `<audio>` element in a Set, so stopping playback tears down every active clip instead of only the most recent one.
- Game voice auto-play now marks main and side voice keys as in flight synchronously, preventing `gameVoiceVersion` churn from relaunching the same ready clip before playback reports started.
- `play()` rejection handling now treats a still-playing audio element as a normal start instead of dropping it from tracking, avoiding orphaned clips that cannot be stopped later.
- No intentional behavior change outside game-mode voice auto-play.

## Validation

- [x] `pnpm check` passed locally in a temporary checkout on June 20, 2026. The temp shell emitted a Node engine warning (`v20.11.1` vs repo target `>=24 <26`), but the full check completed successfully.
- [x] GitHub `pnpm-validate` passed.
- [x] GitHub `container-build-test` passed.

## Manual test plan (unchecked — for human verification)

- [ ] In game mode with a TTS provider + voice auto-play, start a fresh scene that generates multiple voice lines; confirm a single voice plays per line with no echo/overlap.
- [ ] Manually verify per-character / narrator voice routing still works in game mode.
- [ ] Manually verify pause / resume / restart and the stop button.
- [ ] Manually verify an autoplay-blocked clip (no user gesture yet) still retries after interaction (`NotAllowedError` -> `onBlocked` path).
- [ ] Manually verify roleplay/VN and conversation TTS auto-play are unchanged.


SpicyMarinara:
Thank you so much!

PR #2681, opened 2026-06-20
fix(agents): await tool-loop result so a tool failure can't corrupt co-grouped agents

### LINKED ISSUE #2680: [Issue]: Tool agent's tool-call exception corrupts co-grouped agents (missing await in executeAgent) → siblings overwritten as failed
## Summary

A tool-using agent whose tool call throws does not just fail itself — it corrupts the results of **every other agent** sharing its provider+model job group, overwriting their already-successful output with `success: false` error results.

## Expected behavior

When a single tool agent's tool call throws, only that agent should be marked failed (a failed `AgentResult` via `makeError`). Co-grouped agents (e.g. the built-in `tracker`/`expression` batch agents, or sibling tool agents) that ran successfully should keep their real output, and the main user turn should be unaffected.

## Actual behavior

A single tool agent's rethrow rejects the whole group, and every agent in that group — including co-grouped batch agents and sibling tool agents that already **succeeded** — is overwritten with a `success: false` error result, discarding their real output.

Root cause chain:

- `packages/server/src/services/agents/agent-executor.ts` — `executeAgent` returns the tool path with `return executeAgentWithTools(...)` and **no `await`** (line ~305). Because there is no `await`, the function's own `try/catch` (catch at line ~420, which would have produced a failed `AgentResult` via `makeError`) never runs for the tool path.
- When a tool call throws, `executeAgentWithTools` rethrows (`throw err`, line ~525). With no `await` at the call site, `executeAgent` returns a **rejected** promise instead of a failed `AgentResult`.
- `packages/server/src/services/agents/agent-pipeline.ts` — `executeGroup` joins batch and tool agents for the same provider+model group with `Promise.all([...])` (line ~188), with the tool agents inside a `Promise.all` (lines ~179–186). One tool agent's rejection rejects that `Promise.all`, rejecting the whole `executeGroup`.
- `executePhase` catches the rejected group via `Promise.allSettled` and, at lines ~245–262, overwrites **every** agent in the group with `success: false` error results — including co-grouped batch agents and sibling tool agents that already succeeded.

## Steps to reproduce

1. Configure a custom tool-using agent that shares the **default-for-agents** connection (same provider + model) with the built-in batch agents (e.g. `tracker`, `expression`).
2. Trigger a turn where the custom agent makes a tool call that throws (e.g. the tool's `executeToolCall` raises).
3. Observe that the co-grouped batch agents' feedback is reported as failed even though they completed successfully.

## Environment

- Marinara Engine version: 2.0.0 (reproduced against `staging`)
- Install type: source
- OS + version: any
- Browser or app shell: n/a (server-side agent pipeline)

In the server log you would see `[agent-pipeline] Group REJECTED in phase "..."` followed by the group's whole agent list, even though only one tool agent actually failed.

## Additional context

Two fix options:

- **(a)** Add `await` at `agent-executor.ts:305` (`return await executeAgentWithTools(...)`) so the existing catch at ~420 converts the rethrow into a failed `AgentResult` for **only** that agent. Minimal; preserves per-agent isolation.
- **(b)** Stop rethrowing at `agent-executor.ts:525` and instead return a failed `AgentResult` for that agent.

Option (a) is the minimal fix. A PR implementing it is being opened against `staging`.

## Template check

kolacheee:
## Linked issue

Closes #2680

## Why this change

A tool-using agent whose tool call throws should fail **only itself** — but today it corrupts every other agent that shares its provider+model job group:

- `executeAgent` returned the tool path with `return executeAgentWithTools(...)` and **no `await`** (`agent-executor.ts`). Without the `await`, a rethrow from the tool loop (`executeToolCall` throwing → `throw err`) escaped this function's own `try/catch` — the catch that would have produced a failed `AgentResult` via `makeError` — and surfaced as a **rejected promise**.
- In `agent-pipeline.ts`, `executeGroup` joins batch + tool agents for the same provider+model group with `Promise.all([...])`. One tool agent's rejection rejected that `Promise.all`, rejecting the whole group.
- `executePhase` then caught the rejected group via `Promise.allSettled` and overwrote **every** agent in the group — including co-grouped batch agents (e.g. `tracker`, `expression`) and sibling tool agents that already **succeeded** — with `success: false` error results, discarding their real output.

This is reachable with a realistic config: a custom tool-using agent sharing the default-for-agents connection with the built-in batch agents.

## What changed

- Added `await` to the tool-loop call in `executeAgent` (`return await executeAgentWithTools(...)`), so a rethrow is caught by the function's existing `catch` and converted into a failed `AgentResult` for **only** the failing agent — restoring per-agent isolation.
- Added an explanatory comment at the call site so the `await` isn't "optimized away" by a future reader.

Minimal, behavior-preserving for the success path: the no-tools path already returned a failed `AgentResult` on error, and this makes the tool path do the same.

## Validation

### Manual verification notes

`pnpm check` (TypeScript + ESLint + build) passed in the automated run that produced this change. A human should still confirm:

- **Manually verify isolation:** configure a custom tool-using agent that shares the default-for-agents connection (same provider + model) with the built-in batch agents (`tracker`, `expression`). Force its tool call to throw (e.g. a tool whose `executeToolCall` raises). Confirm that **only** the failing tool agent is reported `success: false`, and the co-grouped batch agents keep their real output (no `[agent-pipeline] Group REJECTED` log for the whole group).
- **Manually verify sibling tool agents:** if another tool agent in the same group succeeds, confirm it is no longer overwritten as failed.
- **Manually verify the main turn:** confirm the main user generation turn is unaffected by the failing agent.

## Docs and release impact

- [x] No docs changes needed

n/a — server-side agent pipeline change, no UI surface.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling in agent execution to ensure tool-related failures are properly captured and converted to failed results, improving overall reliability and error reporting.

### DIFF
--- a/packages/server/src/services/agents/agent-executor.ts
+++ b/packages/server/src/services/agents/agent-executor.ts
@@ -300,9 +300,13 @@ export async function executeAgent(
     const maxTokens = applyProviderMaxTokensOverride(provider, normalizeAgentMaxTokens(config.settings.maxTokens));
     const streamResponses = context.streaming !== false;
 
-    // If tools are available, use the tool call loop
+    // If tools are available, use the tool call loop.
+    // `await` so a rethrow from the tool loop is caught by this function's
+    // catch below and converted into a failed AgentResult for THIS agent only,
+    // instead of rejecting the promise and corrupting co-grouped agents in the
+    // pipeline (see executeGroup's Promise.all).
     if (toolContext && toolContext.tools.length > 0) {
-      return executeAgentWithTools(
+      return await executeAgentWithTools(
         config,
         messages,
         provider,


kolacheee:
@coderabbitai review

PR #2776, opened 2026-06-22
fix(storage): make lastContactByCharacter work on the file-storage backend

### LINKED ISSUE #2775: lastContactByCharacter throws "groupBy is not a function" on the file-storage backend
### Summary

`chats.lastContactByCharacter()` (`packages/server/src/services/storage/chats.storage.ts`) builds a Drizzle query using `.groupBy()` plus a `sql` `MAX()` aggregate. The default **file-storage** backend's query-builder shim (`packages/server/src/db/file-backed-store.ts`) implements `select` / `from` / `where` / `orderBy` but **not** `groupBy`, so the call throws at runtime:

```
TypeError: db.select(...).from(...).where(...).groupBy is not a function
    at Object.lastContactByCharacter (.../dist/services/storage/chats.storage.js)
    at Object.<anonymous> (.../dist/routes/conversation.routes.js)
```

### Impact

`conversation.routes.ts` calls this when the conversation view loads, so on the **default file-storage backend** that route 500s repeatedly. It works on the libsql backend (real Drizzle supports `.groupBy()`), which is why it can pass unnoticed in libsql setups.

### Root cause

`.groupBy()` is used in exactly one place in the server (this function), and the file-backed store never implemented it. It's also the only call relying on a `sql` `MAX()` aggregate, which the file store doesn't evaluate.

### Reproduce

1. Run on the default (file-storage) backend.
2. Open a chat / load the conversation view.
3. Server logs the `TypeError` above (HTTP 500).

### Suggested fix

Aggregate in JS instead of via SQL `GROUP BY` / `MAX()`: select the plain `characterId` + `createdAt` columns (both supported by the file store) and reduce to latest-per-character in code. Backend-agnostic; no schema/API change. PR to follow.

Branch: `staging`.

kolacheee:
## Linked issue

Closes #2775

## Why this change

- `chats.lastContactByCharacter()` built its query with `.groupBy()` plus a `sql` `MAX()` aggregate. The default **file-storage** backend's query-builder shim (`file-backed-store.ts`) implements `select`/`from`/`where`/`orderBy` but **not** `groupBy` (and does not evaluate `sql` `MAX()` aggregates), so the conversation route **500s with `groupBy is not a function`** on that backend. It works on libsql, where real Drizzle supports `groupBy`.
- Net effect: on the default backend, loading the conversation view fails and last-contact-per-character data is unavailable.

## What changed

- `packages/server/src/services/storage/chats.storage.ts`: replaced the SQL `GROUP BY` / `MAX()` aggregation with an in-JS reduction. The query now selects plain `characterId` + `createdAt` columns (using only `select`/`from`/`where`, which both backends support) and folds the rows into a `{ characterId: latestCreatedAt }` map. `created_at` is a TEXT (ISO 8601) column, so lexicographic `>` is chronological.
- Dropped the now-unused `sql` import from `drizzle-orm`.
- `.groupBy()` is used in exactly this one place in the server, so dropping the aggregate here is smaller and safer than teaching the file-store shim SQL aggregation.
- Behavior is unchanged on the libsql backend (identical result, computed in JS instead of SQL). No schema or API changes.

## Validation

### Manual verification notes

- **Already run:** `pnpm --filter @marinara-engine/server lint` (`tsc -b ../shared && tsc --noEmit`) passes. Full `pnpm check` (TypeScript + ESLint across the workspace) has not yet been run.
- **Manually verify on the file-storage backend:** open a chat / load the conversation view → confirm the route no longer 500s and last-contact data still populates. (The original `groupBy is not a function` 500 was reproduced from server logs before the change; in-app confirmation of the fix is pending a rebuild.)
- **Manually verify on the libsql backend:** confirm last-contact-per-character data is unchanged (regression check, since this path was previously SQL-side).

## Docs and release impact

- [x] No docs changes needed

_Assessment: storage-internal refactor with no API/schema/behavior change and no version bump, so no docs or release-notes updates appear necessary — left unchecked for the contributor to confirm._

N/A — no UI changes (storage-internal fix).

## Summary by CodeRabbit

* **Refactor**
  * Optimized backend data retrieval for character contact history.

### DIFF
--- a/packages/server/src/services/storage/chats.storage.ts
+++ b/packages/server/src/services/storage/chats.storage.ts
@@ -1,7 +1,7 @@
 // ──────────────────────────────────────────────
 // Storage: Chats
 // ──────────────────────────────────────────────
-import { eq, desc, and, gt, inArray, isNull, isNotNull, sql } from "drizzle-orm";
+import { eq, desc, and, gt, inArray, isNull, isNotNull } from "drizzle-orm";
 import type { DB } from "../../db/connection.js";
 import {
   chats,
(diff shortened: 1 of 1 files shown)


kolacheee:
@coderabbitai review

PR #2807, opened 2026-06-22
fix(client): use dynamic programmatic file picker to fix Linux/Firefox silent import failures

Minsklatte:
## Why this change is needed
When attempting to import custom CSS themes or JS/JSON extensions on certain browser/OS environments (specifically Firefox on Gtk/Linux desktops), clicking the import button successfully opens the native file picker, but selecting a file and clicking 'Open' silently drops the event with absolutely no console errors or toasts.

### Root Cause
1. **Layout Deactivation (`display: none` / `hidden`)**: Certain layout engines (Gecko/Firefox Gtk portals) do not deliver file-picker payloads or fire `change` events on elements with 0x0px visual size or de-activated display trees.
2. **Cyclic Nesting**: Nesting file `<input>` nodes inside `<label>` elements creates mouse-click delegation loops in Gtk/Firefox that abort the change event thread on return.
3. **Stale Refs / Detached Nodes**: React StrictMode remount cycles can cause `fileRef.current` click captures to trigger on detached DOM elements, which successfully open native dialogs but cannot bubble events back to React's document root.

### The Fix
We completely replaced all hidden HTML file-picker nodes with a purely programmatic, dynamic file picker trigger (`triggerFilePicker`). 
- Creates a clean native `<input type="file">` dynamically in global scope when clicked.
- Listens natively to the change event, runs the clean handler callback, and instantly removes the element from the DOM.
- This bypasses all React ref/mounting lifecycle gotchas, has non-zero size, is completely immune to Gtk-portal deactivation, and has been proven 100% reliable on Firefox, Chrome, and Gtk Linux portals.

### DIFF
--- a/packages/client/src/components/panels/SettingsPanel.tsx
+++ b/packages/client/src/components/panels/SettingsPanel.tsx
@@ -3323,7 +3323,6 @@ function ThemesSettings() {
   const updateTheme = useUpdateTheme();
   const deleteTheme = useDeleteTheme();
   const setActiveTheme = useSetActiveTheme();
-  const fileRef = useRef<HTMLInputElement>(null);
   const activeCustomTheme = syncedThemes.find((theme) => theme.isActive) ?? null;
   const isSavingTheme = createTheme.isPending || updateTheme.isPending || setActiveTheme.isPending;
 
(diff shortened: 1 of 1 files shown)


SpicyMarinara:
thank you. <3

PR #2875, opened 2026-06-25
fix: preserve knowledge router prompt injections

### LINKED ISSUE #2874: [Issue]: Knowledge Router's output doesn't appear anywhere
## Summary

In the last few days, Knowledge Router would sometimes work on one message, then *not* work on the next few. I did not manage to figure out the trigger. Since last night's update, I can't get it to show the output in my existing chat at all. The call goes through, it creates the output, but it doesn't show up within the app.

## Expected behavior

When Knowledge Router is activated and connected to a lorebook, it should scan the entries and return the selection. That should be visible in "Peek prompt", and the main model should take the info in consideration.

## Actual behavior

Everything goes through: according to my API calls, the lorebook was read -> it returned, say, 500-1000 tokens. But then Knowledge Router output is nowhere to be seen in the "Assembled Prompt" menu. So the process *is* happening, but the output is invisible to me and the model both.

Fresh new chat with the same settings works when it's just a few messages; old chats don't. 

**Tried to fix it:** Reinstalled the Engine, uninstalled + installed fresh build, wiped my data (specifically the agent configs in ...\data\storage\tables). Tried different models, default preset / my own preset. Putting Knowledge Router as a section, disabling it as a section. Wiping all Lorebooks from the chat's connections (disabling, then manually deleting them from every single place that mentions lorebookID in the .jsonl and then uploading it back in; the UI showed no lorebooks after) and reconnecting them, etc.

I also tried to test it with instructions to the model (put in a lorebook entry), and it did not respond to it, so it's likely not "sneakily" passing in the background.

## Steps to reproduce

0. Open established chat
1. Knowledge Router enabled + enabled to show as section
2. Preset has Knowledge Router inserted as a section
3. Activate Knowledge Router in chat
4. Lorebook connected to chat + manually connected to Knowledge Router just to be sure
5. Send a message
6. API call goes through, i.e. connection works
7. Model delivers reply
8. Knowledge Router is NOT in the final prompt (section missing entirely) + the info was not acknowledged

## Environment

- Marinara Engine version: v2.0.5
- Install type: installer
- OS + version: Windows 11 Pro
- Browser or app shell: Browser

It doesn't show up in console. Other agents work.

## Template check

TheLonelyDevil9:
## What?
Fix Knowledge Router output being omitted from assembled prompts in established chats.

## Why?
Router requests can complete and return lorebook selections, but established chats can lose that output before final prompt assembly, so the main model cannot use it.

## How?
- Store separate Knowledge Router and Knowledge Retrieval prompt injections as protected prompt injection messages instead of appending them to chat history.
- Keep the runtime-section replacement path intact when a preset section handles the agent output.
- Add a prompt regression that forces long-history context fitting and verifies the router injection survives while old history is trimmed.

## Testing?
- pnpm regression:prompt
- pnpm check

## Anything Else?
Closes #2874.

## Summary by CodeRabbit

* **New Features**
  * Improved how injected agent instructions are added to prompts, with cleaner formatting for different agent types and wrapping styles.
* **Bug Fixes**
  * Prompts now preserve injected instructions more reliably when long conversation history is trimmed for model limits.
* **Tests**
  * Added regression coverage for prompt fitting to ensure older context is removed while injected routing instructions remain included.

### DIFF
--- a/packages/server/src/routes/generate.routes.ts
+++ b/packages/server/src/routes/generate.routes.ts
@@ -203,6 +203,7 @@ import {
   buildLockedPersonaTrackerPatch,
   extractImageAttachmentDataUrls,
   appendNonLeadingSystemMessagesToLastUser,
+  appendSeparateAgentInjectionMessage,
   computeSummaryHideIds,
   injectIntoOutputFormatOrLastUser,
   isManualTrackerCharacterId,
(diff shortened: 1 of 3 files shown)


SpicyMarinara (approved):
Verified the CodeRabbit regression request is addressed by the assistant/history-tail prompt fitting case. Local validation: pnpm regression:prompt; pnpm check.

WaterBloom:
Thank you for looking into it @TheLonelyDevil9! Do I understand right it's in the live version already?

I just tried it (reset all my settings etc.) and it still doesn't work in a long chat, 55k total prompt. Request did go through, showing me received 16,625 → output 695 tokens.

Strangely enough, I also put some extra text into the section (normal text before {{agent::knowledge-router}}) and the section _still_ completely disappears from the final assembly.

PR #3115, opened 2026-06-30
fix(client): O(1) segment index lookup in game narration (#3104)

### LINKED ISSUE #3104: [Issue]: Severe client freeze (multi-second main-thread blocks, GPU spikes) on chats using the world-state / character-tracker agents
**What's wrong**

Opening a chat that uses the **world-state** and/or **character-tracker** agents puts the client into a severe main-thread freeze: interactions take tens of seconds, the GPU spikes, and the tab is effectively unresponsive ("memory-leak vibes"). It is reported in **all chat modes** (Conversation, Roleplay, and Game), only on chats that ran these agents, and it **stops the moment you navigate to home, a new chat, or a chat that did not use those agents**. It is worst on **older chats** (created/used on earlier versions) that have accumulated a lot of tracked-character / world-state data.

This is a **client-side** problem (the browser main thread is blocked), not slow generation — it happens while the chat just sits open, and clears when you leave the chat.

1. Open a chat that has the **character-tracker** and/or **world-state** agent enabled and that has been used for many turns (so the tracker has accumulated several present characters / world-state), ideally on an install with a large character/persona library.
2. Let the chat open, then try to interact (click a message, scroll, delete a message, type a reply).

You'll see: each interaction takes many seconds to tens of seconds; the GPU revs; deleting/replying causes a heavy momentary freeze. Navigating to the home page or a non-agent chat makes it instantly responsive again.

**Evidence captured from affected users (DevTools)**

- **Web Vitals INP = 188,848 ms.** Two recorded `pointer` interactions: one `115,472 ms` (input delay **69,013 ms**, processing **46,411 ms**), one `188,848 ms` (input delay **49,009 ms**, processing **46,607 ms**, presentation **93,233 ms**). The ~46 s *processing* + ~49–69 s *input delay* mean the main thread is doing ~46 s synchronous chunks of work back-to-back.
- Network requests left **pending >50 s** (fetch callbacks starved behind the blocked main thread).
- Opening an affected Roleplay chat lazy-loads exactly these chunks: **`TrackerDataSidebar` (~175 kB)**, `RoleplayHUD` (~29 kB), `world-state-display` (~12 kB), `TrackerSizeTierIcon` — i.e. the tracker-panel / world-state UI.
- One user: *"it's loading every persona and character in my app, even ones not in this chat."*

**Where it is (localized — see confidence note below)**

The hotspot is the **tracker-panel feature** that renders the world-state / character-tracker agent output. It mounts only when those agents are active (which is why the lag tracks "chats with agents" across all modes, and clears on chats without them):

- `packages/client/src/features/tracker-panel/components/TrackerDataSidebar.tsx`, `TrackerSectionList.tsx`, `components/sections/WorldStatePanel.tsx`, `components/character-card/CharacterTrackerCard.tsx`, and `packages/client/src/components/chat/RoleplayHUD.tsx` — the components in the lazy chunks above.
- **Whole-library over-fetch + parse** (matches "loading every character/persona"): `packages/client/src/features/tracker-panel/hooks/use-tracker-sprite-lookup.ts:16-49` calls `useCharacters()` (the entire library) and then maps `parseCharacterDisplayData(character)` over **every** character row to build a name→id lookup — even though only the chat's tracked characters are rendered. `packages/client/src/features/tracker-panel/hooks/use-tracker-panel-model.ts:70` likewise loads **all** personas. For a library of hundreds of cards, building that lookup is an O(library) JSON-parse pass that re-runs whenever the character-list reference changes.
- The per-tracked-character render: `components/sections/CharacterTrackerPanel.tsx:76` maps a `CharacterTrackerCard` per present character; old chats accumulate many present characters.

**What's happening (leading mechanism)**

A chat with the world-state/character-tracker agents mounts the tracker panel, which (a) loads and **parses the entire character + persona library** to build sprite/name lookups it only needs the chat's characters for, and (b) renders a card per accumulated tracked character. The cost scales with **library size × tracked-character count**, both of which are large for long-lived older chats on big installs — producing the ~46 s synchronous main-thread blocks. Because the panel re-renders on every gameState/agent update and on edits, the freeze recurs continuously while the chat is open.

**Ruled out** (investigated and excluded, so reviewers don't re-chase them):
- Macro corruption / re-expansion of stored agent macros — the engine deliberately prevents this (`{{agent::TYPE}}` is a terminal flat substitution placed after all macro passes); stored tracker text is never re-run through `resolveMacros`.
- Server generation cost / tracker-context assembly — those are per-send, can't cause a freeze while the chat sits idle.
- The game-session auto-switch effect (`resolveCurrentGameSessionChatId`) — it early-returns for non-game chats, so it cannot affect Conversation/Roleplay.
- The macro engine itself — benchmarked on the compiled build; ~336 ms worst case for a 200 KB prompt, guarded by `MAX_MACRO_EXPANSIONS`.

**Suggested direction**

1. Pin the exact hot frame with a **Performance → Bottom-Up (self-time)** profile recorded while an affected chat is open — it will name the precise function inside `features/tracker-panel/**`.
2. Likely fixes once confirmed: scope `use-tracker-sprite-lookup` to the **chat's** characters instead of parsing the whole library; memoize/virtualize the tracked-character card list; avoid re-parsing character cards on every list-reference change; and bound work to the number of present characters actually rendered.

**Confidence note**

This is a **localization**, not a pinned root-cause line. The evidence (the loaded tracker chunks, the INP breakdown, the whole-library parse in `use-tracker-sprite-lookup`, and the symptom matrix) strongly points at the tracker-panel render, but the single exact O(n) frame should be confirmed with a Bottom-Up profile or a local repro (a chat with the character-tracker/world-state agents and many accumulated present characters on a large-library install) before fixing.

---
*AI-investigated against v2.0.8 (`main`). Symptom data from affected users' DevTools (INP/Network); code paths read and cited. The exact hot function is not yet pinned — please capture a Bottom-Up profile or reproduce locally to confirm before fixing.*

kolacheee:
## Linked issue

Partially addresses #3104 (the **game-mode** half). This does **not** close #3104 — the roleplay/conversation lag is a separate, still-open thread (different render path; this change does not touch it).

## Why this change

A contributor confirmed the lag does **not** exist in v2.0.6, so #3104 is a regression. Bisecting the render-path diff (v2.0.6 → staging) surfaced one change that fits the instrumented signature (8–22s `render+commit` scaling with chat length, game mode):

In `GameNarration.tsx`, the stacked-log narration renderer computes, at the **top of each per-segment render**:

```js
const liveSegmentIndex = segments.findIndex((candidate) => candidate.id === seg.id);
```

`segments` is the **entire chat's** flattened, uncapped segment array, and `renderStackedLogSegment` runs **once per visible segment**. So the cost is `rendered × total` = **O(n²) in segment count, per render**. On long agent-heavy chats (hundreds of messages → thousands of segments) that's thousands of full-array scans per render — matching the measured multi-second freeze, and scaling with chat length exactly as users report.

In v2.0.6 this renderer was display-only and had no `findIndex`. The eager lookup was added in **v2.0.7, commit `e1e378b7e`** ("Prepare v2.0.7 follow-up fixes"), when the segment renderer gained an inline action-button suite. `GameNarration` is rendered solely by `GameSurface`, so this is game-mode only.

## What changed

Build an `id → index` Map once per `segments` change (a `useMemo` keyed on `segments`) and look it up in O(1) at both call sites (the delete-navigation handler and the jump-to-segment row). This collapses the per-render cost from O(n²) to O(n) (one Map build) + O(1) lookups.

Behavior-identical: `findIndex` returned `-1` when absent; `Map.get(...) ?? -1` returns the same. No visual or behavioral change — purely the cost of the index lookup.

## Validation

### Manual verification notes

Done by the agent (not a substitute for human verification):
- `pnpm check` → exit 0 (TypeScript + ESLint + client build all clean).

Still needs manual verification:
- **Recommended before merge:** a DevTools Performance profile on a long, agent-heavy **game-mode** chat, before vs. after, confirming `renderStackedLogSegment` / `findIndex` self-time drops out (per #3104's "confirm perf fixes with a profile, don't trust code-reading alone").
- Open the **Logs** panel on a long game chat and confirm: stacked-log entries still render correctly, the **delete** button still deletes the right message/segment, and the **jump-to-segment** (click a past log row) still jumps to the correct segment.
- Confirm short/empty game chats and non-game chats are unaffected.

## Docs and release impact

- [ ] No docs changes needed

No UI change — this only changes how the per-segment index is computed (eager O(n²) scan → O(1) Map lookup).

## Summary by CodeRabbit

* **Performance**
  * Improved game narration log rendering by using faster segment lookup, making log updates and navigation more responsive.
* **Bug Fixes**
  * Fixed segment selection behavior in stacked and modal log views so “jump back” actions reliably target the correct entry.

### DIFF
--- a/packages/client/src/components/game/GameNarration.tsx
+++ b/packages/client/src/components/game/GameNarration.tsx
@@ -1670,6 +1670,16 @@ export function GameNarration({
     sourceMessagesById,
   ]);
 
+  // Fast O(1) segment-id → index lookup, rebuilt only when `segments` changes.
+  // The per-segment log renderers below previously each called
+  // `segments.findIndex(...)`, making stacked-log rendering O(n²) in segment
+  // count — the root of the #3104 game-mode freeze on long chats.
+  const segmentIndexById = useMemo(() => {
+    const map = new Map<string, number>();
+    segments.forEach((segment, index) => map.set(segment.id, index));
+    return map;
+  }, [segments]);
+
   // Clamp activeIndex when segments shrink (e.g. new party chat clears old dialogue)
   useEffect(() => {
     if (segments.length > 0 && activeIndex >= segments.length) {
@@ -3705,7 +3715,7 @@ export function GameNarration({
     const sourceMessageRole = sourceMessageId ? (sourceMessagesById.get(sourceMessageId)?.role ?? null) : null;
     const sourceRole = seg.sourceRole ?? sourceMessageRole;
     const isUserAuthoredSource = sourceRole === "user" || sourceMessageRole === "user";
-    const liveSegmentIndex = segments.findIndex((candidate) => candidate.id === seg.id);
+    const liveSegmentIndex = segmentIndexById.get(seg.id) ?? -1;
     const canEditMessage = !!onEditMessage && !!sourceMessageId && isUserAuthoredSource;
     const canEditSegment =
       !!onEditSegment &&
@@ -4865,7 +4875,7 @@ export function GameNarration({
                       const sourceRole = seg.sourceRole ?? sourceMessageRole;
                       const isUserAuthoredSource = sourceRole === "user" || sourceMessageRole === "user";
                       const isActiveSeg = active?.id === seg.id;
-                      const liveSegmentIndex = segments.findIndex((s) => s.id === seg.id);
+                      const liveSegmentIndex = segmentIndexById.get(seg.id) ?? -1;
                       const canJumpToSeg =
                         !!latestAssistant &&
                         sourceMessageId === latestAssistant.id &&


SpicyMarinara:
@coderabbitai review

SpicyMarinara (approved):
Approved for staging after green required checks and CodeRabbit reported no actionable comments.

PR #3114, opened 2026-06-30
[codex] Reduce chat area render scope

SpicyMarinara:
## Summary

- Scope ChatArea character and persona data to the active chat instead of loading full libraries on the hot render path.
- Add an active-persona endpoint/query for the non-game fallback persona case.
- Scope Conversation emoji/sticker display-name lookups to active participants.
- Lower the mounted transcript render window for long Conversation and Roleplay chats.
- Update the v2.0.9 changelog.

## Why

User performance traces showed `chat-area render+commit` taking 6-24 seconds while tracker-panel commits stayed near 0-8ms. The chat surface was paying library-wide lookup costs during active chat renders.

## Validation

- pnpm check

## Manual Verification

- [ ] Manually verify a long Conversation chat opens and scrolls without multi-second chat-area commits.
- [ ] Manually verify Roleplay message rendering still shows character/persona styling, creator notes CSS, emojis, and stickers.
- [ ] Manually verify Game mode still has access to the broader character library where required.

## Summary by CodeRabbit

* **New Features**
  * The app now consistently uses the currently active persona in chat-related views and emoji/sticker display.
  * Long conversations should load more smoothly, with fewer messages kept mounted at once.

* **Bug Fixes**
  * Reduced render stalls in chats by narrowing lookups and rendering work to the active conversation.
  * Improved character and persona display accuracy in game and roleplay chats.

PR #3118, opened 2026-06-30
fix(shared): make macro display resolution linear, not super-linear (#3104)

### LINKED ISSUE #3104: [Issue]: Severe client freeze (multi-second main-thread blocks, GPU spikes) on chats using the world-state / character-tracker agents

kolacheee:
## Linked issue

Relates to #3104 — this is the **primary per-resolution-cost** fix (the macro engine), complementary to #3114 ("Reduce chat area render scope", the re-render-frequency side). It should **not** auto-close #3104 until an affected user confirms the freeze is gone (see below).

## Why this change

A Bottom-Up CPU profile from an affected user pinned the #3104 freeze: **`resolveMacros` at 75% self-time (~38s)**. `resolveMacros` runs **per message during display render in every mode** (`ChatMessage`, `ConversationMessage`, `GameNarration`), so this is the cross-mode cause behind the game / roleplay / conversation reports.

The root cause: `resolveMacros` **eagerly and unconditionally rebuilt `personaText`** — resolving all five persona fields, each of which **recurses back into `resolveMacros`** when it contains macros — on *every* call, *before* checking whether the template even uses `{{persona}}`. Worse, that rebuild is **re-paid at every nested field-resolution frame**, so it compounds. On agent chats with large, self-referential persona fields, per-message display resolution goes super-linear. (Introduced by the v2.0.7 macro rewrite, `afa0cc3ab`.)

## What changed

Two **behavior-preserving** changes in `packages/shared/src/utils/macro-engine.ts`:

1. **Top-level early-out.** Every step of `resolveMacros` is triggered by either `{{` (live macros) or the `\x1e` deferred-character-token sentinel. A template containing neither has nothing to resolve, so it's returned unchanged (with the same `trimResult`/`clampMacroOutput` finalize) instead of running the comment strip, bracket/conditional expansion, the global substitution passes, and the persona build. This covers the bulk of display content (plain prose).
2. **Lazy `personaText`.** Build it only when `{{persona}}` is actually present in the result, keeping the **string-replace form** so `$`-sequences substitute exactly as before. This stops the unconditional (and per-recursion-frame) persona rebuild.

**No budget caps were touched.** `MAX_MACRO_EXPANSIONS` / `MAX_MACRO_OUTPUT_LENGTH` / depth limits are unchanged, so this does **not** trade speed for unresolved (literal) macros — macros still fully resolve, always for API prompts and on user-visible display content.

## Validation

### Manual verification notes

Done by the agent:
- `pnpm check` → exit 0 (TypeScript + ESLint + build).
- **Byte-identical output proven** old (staging) vs new across: plain prose, `{{user}}`/`{{char}}`, `{{persona}}` with a `$&` sequence (string-replace semantics preserved), char-field macros, `{{#if}}` conditionals, and a `\x1e` deferred-character token (passed through untouched). Output matched exactly in every case.
- Timing on a synthetic ~12KB self-referential persona context: plain msg **59ms → 0ms**, `{{char}}` msg **57ms → 0.13ms**, `{{persona}}` msg **58ms → 1.26ms** (~50–450×).

Still needs manual verification (decisive — don't merge on synthetic/code-reading alone):
- **An affected user re-tests the real laggy chat on this branch** and confirms the multi-second `chat-area` freeze is gone (their data is what produced the 38s; the synthetic case only proves the *shape*).
- Spot-check that `{{user}}`/`{{char}}`/`{{persona}}` and conditionals still render correctly in a real roleplay/conversation/game chat, and that an API prompt (peek-prompt) still fully resolves macros.

## Docs and release impact

- [ ] No docs changes needed

No UI change — macro output is identical; only the cost of producing it changes.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved macro processing so plain text templates are handled faster, with less unnecessary work.
  * Reduced extra processing for persona content by generating it only when needed.

### DIFF
--- a/packages/shared/src/utils/macro-engine.ts
+++ b/packages/shared/src/utils/macro-engine.ts
@@ -1138,6 +1138,15 @@ export function resolveMacros(template: string, ctx: MacroContext, options: Reso
     return clampMacroOutput(template, options);
   }
   getMacroBudget(options);
+  // #3104: content with no macro syntax has nothing to resolve. Every step below
+  // is triggered by either "{{" (live macros) or the "\x1e" deferred-character
+  // token sentinel, so a template containing neither is returned unchanged —
+  // skipping the comment strip, bracket/conditional expansion, the global
+  // substitution passes, and the persona-field build. Output is identical.
+  if (!template.includes("{{") && !template.includes("\x1e")) {
+    const passthrough = options.trimResult !== false ? template.trim() : template;
+    return clampMacroOutput(passthrough, options);
+  }
   let result = template;
   const fieldResolutionDepth = options.fieldResolutionDepth ?? 0;
   const resolveNestedFieldMacros = (value: string): string => {
@@ -1150,16 +1159,6 @@ export function resolveMacros(template: string, ctx: MacroContext, options: Reso
       fieldResolutionDepth: fieldResolutionDepth + 1,
     });
   };
-  const personaText = [
-    ctx.personaFields?.description,
-    ctx.personaFields?.personality,
-    ctx.personaFields?.backstory,
-    ctx.personaFields?.appearance,
-    ctx.personaFields?.scenario,
-  ]
-    .map((part) => (typeof part === "string" ? resolveNestedFieldMacros(part) : part))
-    .filter((part): part is string => typeof part === "string" && part.trim().length > 0)
-    .join("\n");
   const deferCharacterMacros = options.deferCharacterMacros;
   const characterReplacement = (field: keyof typeof DEFERRED_CHARACTER_MACRO_TOKENS): string => {
     if (deferCharacterMacros === "all" || (deferCharacterMacros === "names" && field === "char")) {
@@ -1184,7 +1183,23 @@ export function resolveMacros(template: string, ctx: MacroContext, options: Reso
 
   // ── Static substitutions ──
   result = result.replace(/\{\{user(?:Name)?\}\}/gi, ctx.user);
-  result = result.replace(/\{\{persona\}\}/gi, personaText);
+  // #3104: build the persona field text lazily — only when {{persona}} is actually
+  // present — so the recursive 5-field resolution is not paid on every message
+  // (it used to run unconditionally on every resolveMacros call). String form is
+  // kept so $-sequences in persona text substitute exactly as before.
+  if (/\{\{persona\}\}/i.test(result)) {
+    const personaText = [
+      ctx.personaFields?.description,
+      ctx.personaFields?.personality,
+      ctx.personaFields?.backstory,
+      ctx.personaFields?.appearance,
+      ctx.personaFields?.scenario,
+    ]
+      .map((part) => (typeof part === "string" ? resolveNestedFieldMacros(part) : part))
+      .filter((part): part is string => typeof part === "string" && part.trim().length > 0)
+      .join("\n");
+    result = result.replace(/\{\{persona\}\}/gi, personaText);
+  }
   result = result.replace(/\{\{personaDescription\}\}/gi, () =>
     resolveNestedFieldMacros(ctx.personaFields?.description ?? ""),
   );
--- a/packages/shared/src/utils/macro-engine.ts
+++ b/packages/shared/src/utils/macro-engine.ts
@@ -1139,11 +1139,12 @@ export function resolveMacros(template: string, ctx: MacroContext, options: Reso
   }
   getMacroBudget(options);
   // #3104: content with no macro syntax has nothing to resolve. Every step below
-  // is triggered by either "{{" (live macros) or the "\x1e" deferred-character
-  // token sentinel, so a template containing neither is returned unchanged —
-  // skipping the comment strip, bracket/conditional expansion, the global
-  // substitution passes, and the persona-field build. Output is identical.
-  if (!template.includes("{{") && !template.includes("\x1e")) {
+  // is triggered by "{{" (live macros), the "\x1e" deferred-character token
+  // sentinel, or the "\x00" trim-marker sentinel, so a template containing none
+  // of them is returned unchanged — skipping the comment strip,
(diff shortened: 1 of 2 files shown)


kolacheee:
@coderabbitai review

SpicyMarinara:
@coderabbitai review

SpicyMarinara (approved):
Approved for staging after applying CodeRabbit regression-test feedback and passing prompt regression + required checks.

Issue #3164, opened 2026-07-03
[Issue]: Regex scripts apply per-streaming-chunk & cause severe UI freezes


atlasveldine:
## Summary

Using RegEx scripts that replace with macros results in massive UI lag, rendering Marinara Engine effectively unusable. Thank fuck for whoever came up with the idea to implement such a competent agent with the built-in `Mari`, because otherwise I’d not have anything else to report beyond “shit be fucked.” Practically everything ↓ below ↓ is from `Mari`.

---

## Mari’s Breakdown

**File/Function:** `ChatRoleplaySurface`, specifically the useMemo in message rendering around line ~1700. It calls `applyToAIOutput` (the client-side regex function, imported from `PinnedImageOverlay`, function `re`) with a `macroResolver` that gets updated every streaming chunk. Since the memo depends on content, every single token invalidates it, re-running the full regex pipeline for all visible messages simultaneously. Each regex script compiles a fresh `RegExp`, resolves macros in `find`/`replace`/`trimStrings` via the macro engine, and worse, if replacements contain `{{random::…}}` or other dynamic macros, they produce new results every chunk instead of remaining stable. This blocks the main thread for hundreds/thousands of milliseconds (profiler showed 702ms in `macro-engine-qQkcw39S.js` at a bare minimum, but went well into the 3,000ms range).

## Suggested Fix

Move regex application out of the per-message useMemo, apply only to the final message content post-stream, and cache resolved macro values for each script so they don't re-evaluate on every render. Optionally, offload the heavy lifting to a web worker.

## Expected behavior

RegEx scripts should probably apply only *once*, when the stream finishes and the message is completed. If the message is edited later, re-apply at that point. Compare to SillyTavern/RisuAI: they don't burn the main thread on every token. (Although they do have their own issues with UI lag from other sources, still.)

## Actual behavior

RegEx is applied per-chunk when streaming, and bizarrely even randomly re-applies outside of streams, causing massive compounding UI lag.

## Steps to reproduce

Enable regex script(s) with macro-containing replace strings (even simple ones) and start streaming a response. The UI will freeze. Disabling all regex scripts restores full performance.

Specifically, if you *really* want to trigger this, try RegEx like:

```regex
the
```

With a replacement of e.g.:

```text
{{random::one::two::three::four::five::six@0.33::seven::eight@0.9::nine@0.1::ten@0.2}}
```

[...]

PR #3166, opened 2026-07-03
fix(client,shared): stop regex-script macro replacements freezing the UI (#3164)

kolacheee:
## Linked issue

Relates to #3164 — please have the reporter confirm on this branch before closing (their profiler produced the 702–3000ms numbers; the benchmarks below reproduce the mechanism, not their exact hardware).

## Why this change

#3164 reports severe UI freezes with regex scripts whose replacements contain macros (e.g. replace `the` with a weighted `{{random::…}}`). Verified legitimate, with the repro benchmarked against current staging. Two compounding causes:

1. **Multiplication (primary).** The per-message display pipeline (regex scripts + macro resolution) is individually cheap (~1.2ms for 5 scripts, desktop), but it re-ran for **every mounted message on every ChatArea render**. `useQueries` returns a fresh result-array identity each render, so `chatCharacterRows` → `characterMap` were rebuilt — new Map, new value objects — on every re-render, plus a guaranteed rebuild every 60s (presence clock) and on every chat metadata write (constant on agent chats). The new identity defeats `memo(ChatMessage)` and invalidates every message's `displayContent` memo. 160 mounted messages × 5 scripts ≈ 200–610ms per storm on a fast desktop — the reported 702–3000ms on slower hardware — re-paid on chat open, mid-stream agent-store flips, message invalidations, and the idle 60s tick (the reporter's "lag even outside of generation, bafflingly").
2. **Fixed per-resolution cost + streaming churn.** Every macro-bearing replacement resolution paid an unconditional `Intl.DateTimeFormat(...).formatToParts()` (~45% of the pipeline's fixed cost) even with no date macro present, and N scripts sharing one replacement did N identical resolutions. During streaming, the display random seed included message content, so every streamed chunk re-rolled every `{{random}}`/`{{roll}}` — the visible text churn in the report.

## What changed

Four coordinated fixes:

- **ChatArea identity stabilization** (`ChatArea.tsx`): `chatCharIds` keys on the raw `characterIds` field; `chatCharacterRows` reuses the previous array when every element is reference-unchanged; `characterMap` reuses its previous identity when a rebuild is value-equal (comparator covers all 14 value fields; `avatarCrop` compared by value). Unrelated re-renders, presence ticks, and metadata writes that don't change displayed character data no longer re-run the display pipeline for any message, across all three chat surfaces. Genuine changes still propagate.
- **Lazy date/time in the macro engine** (`packages/shared/.../macro-engine.ts`): the six date/time parts are built only when a date/time macro is present; `now` is still captured once per invocation so all six agree on the instant. Output identical (date strings can't contain `$` sequences).
- **Per-resolver template cache** (`chat-macros.ts`): one resolver instance serves all of a message's replacement/trim/pattern strings; identical templates with a fixed (context, seed) resolve identically, so repeats are cache hits. The cache **hard-disables from the first variable-op macro onward** — conditionals and the `{{name}}` catch-all can read variables without matching the var-op pattern, so any write poisons cachability for the rest of the instance.
- **Identity-based display random seeds** (`ChatMessage.tsx`, `GameNarration.tsx`, `ConversationMessage.tsx`): seed by `message.id` + active swipe index (+ segment/part index where applicable) instead of content. Streaming stops re-rolling picks every chunk; swipes keep distinct picks; regenerations (new swipe index) roll fresh. `ConversationMessage` previously passed **no seed at all** (`Math.random()`), re-rolling finished messages on every recompute — it now gains stability it never had.

**Deliberately not done** (so reviewers don't ask): per-script `RegExp`/`isPatternSafe` caching — measured at 0.00007ms/0.0005ms per application, three orders of magnitude below the resolver cost; throttling or freezing regex application during streaming — skipping application would leak content that display scripts exist to hide (scripts keep applying to every streamed frame); the deep ChatArea presence-overlay/`combine` refactor — follow-up scope, the comparator shims already deliver the identity stability.

### Measured (tsx harness against this branch vs staging)

| Scenario | Before | After |
| --- | --- | --- |
| One `{{random::…}}` replacement resolution (fresh resolver) | 0.172 ms | 0.012 ms |
| Message with 5 scripts sharing one replacement | ~0.86 ms | ~0.013 ms |
| Message with 5 distinct replacements | ~0.86 ms | ~0.06 ms |
| Whole-window storm (160 msgs × 5 scripts) on unrelated re-render | 200–610 ms desktop / 702–3000 ms reported | does not occur (identity stable) |

### Documented behavior deltas

- Historical messages re-roll their displayed `{{random}}` picks **once** on upgrade (picks are render-time only, never persisted; edit/copy paths use raw content). Worth a release-note line.
- Identical unseeded templates within one resolution pass now give one consistent pick instead of independent `Math.random()` rolls (affects multi-script user-input paths; arguably the desired consistency).
- All in-flight streams share the streaming placeholder seed until finalization, then settle once on the real message id.

## Validation

### Manual verification notes

- Lint (all three packages), shared build, and client build pass. The full `pnpm check` **server build step fails on clean staging on Windows** (pre-existing v2.1.0 `build.mjs` unquoted-`execPath` spawn bug, reproduced with this branch's changes stashed — separate fix needed; not caused by this PR).
- 20-case local regression matrix passes (15 macro-engine + 5 chat-macros), including: the stale-cache-after-`setvar` case (a cached conditional re-reads correctly after a later variable write), variable write/read sequencing through the cached resolver, same-seed `{{random}}` determinism across resolver instances, and date/time macros resolving after the lazy Intl construction. (`*.test.ts` is gitignored per repo convention — matrix kept local.)
- Benchmarks above.

Still needs manual verification:
- **The #3164 reporter re-tests their regex setup on this branch** (their repro: several scripts replacing common words with weighted `{{random::…}}`) and confirms the freeze is gone during streaming and idle.
- Spot-check in a real browser: regex scripts still apply to streamed text as it arrives; `{{random}}` picks hold steady during a stream and settle once at completion; swiping shows different picks per swipe; presence dots still update (60s schedule transitions and status changes); character edits (name/avatar/colors) still propagate to open chats in all three modes.

## Docs and release impact

- [x] No docs changes needed

Suggested release-note line: "Displayed {{random}}/{{roll}} results are now stable per message/swipe instead of re-rolling on every render or streamed chunk; existing messages will re-roll their displayed picks once after updating."

No visual change — displayed content is identical except that `{{random}}`/`{{roll}}` picks are now stable during streaming instead of flickering per chunk.

### DIFF
--- a/packages/client/src/components/chat/ChatArea.tsx
+++ b/packages/client/src/components/chat/ChatArea.tsx
@@ -354,6 +354,42 @@ function toCharacterMapValue(char: CharacterRow): CharacterMapValue {
   }
 }
 
+// [#3164] Value comparators so the characterMap memo can keep its previous
+// identity when a rebuild produced equal contents. A new map identity re-runs
+// the regex+macro display pipeline for every mounted message, so renders
+// triggered by the presence clock or unrelated metadata writes must not renew
+// it. The field list must cover every field of the CharacterMap value type —
+// a missed field would make a real change invisible to consumers.
+function areCharacterMapValuesEqual(a: CharacterMapValue, b: CharacterMapValue): boolean {
+  return (
+    a.name === b.name &&
+    a.description === b.description &&
+    a.personality === b.personality &&
+    a.backstory === b.backstory &&
+    a.appearance === b.appearance &&
+    a.scenario === b.scenario &&
+    a.example === b.example &&
+    a.avatarUrl === b.avatarUrl &&
+    a.nameColor === b.nameColor &&
+    a.dialogueColor === b.dialogueColor &&
+    a.boxColor === b.boxColor &&
+    a.conversationStatus === b.conversationStatus &&
+    a.conversationActivity === b.conversationActivity &&
+    // avatarCrop is a small plain object — compare by value, not reference.
+    (a.avatarCrop === b.avatarCrop ||
+      JSON.stringify(a.avatarCrop ?? null) === JSON.stringify(b.avatarCrop ?? null))
+  );
+}
+
+function areCharacterMapsEqual(a: CharacterMap, b: CharacterMap): boolean {
+  if (a.size !== b.size) return false;
+  for (const [id, value] of b) {
+    const previous = a.get(id);
+    if (!previous || !areCharacterMapValuesEqual(previous, value)) return false;
+  }
+  return true;
+}
+
 const ChatConversationSurface = lazy(async () => {
   const module = await import("./ChatConversationSurface");
   return { default: module.ChatConversationSurface };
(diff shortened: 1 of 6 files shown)


kolacheee:
@coderabbitai review

PR #3174, opened 2026-07-03
Remove text attachment prompt truncation

SpicyMarinara:
## Summary
- Removes the 60,000-character pre-truncation for readable text attachments before prompt context fitting.
- Lets large uploaded text files use the selected model/connection context window instead of Marinara slicing them first.
- Adds a prompt regression and changelog note.

## Why
A user report described an apparent ~15k token cap for the last user message after uploading a text file. That maps closely to the old 60,000-character attachment cap. This was not a max-output setting issue, and it happened before provider/model context fitting had a chance to decide what actually fits.

Normal lorebook entry content does not appear to have a 50k content cap in the lorebook schema; lorebook budget and activation rules are separate, and provider/model context still remains the real upper bound.

## Validation
- pnpm check
- pnpm regression:prompt
- git diff --check

## Manual verification
- Upload a readable text file larger than 60,000 characters and inspect/debug the generated prompt to confirm the attachment is no longer pre-truncated before context fitting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Large readable text-file attachments are no longer cut off before being included in prompts, so uploaded text can use the model’s full available context.
  * Prompt content now preserves the complete attachment text instead of showing a truncation notice for these files.

* **Tests**
  * Added regression coverage to ensure long text attachments remain intact when added to prompt content.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 
 ### Fixed
 
+- Fixed readable text-file attachments being pre-truncated to 60,000 characters before prompt context fitting, so large uploaded text files can use the selected model's actual context window.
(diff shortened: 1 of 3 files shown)


SpicyMarinara:
@coderabbitai review

PR #3175, opened 2026-07-03
[codex] Restore chat UI fixes and remove lorebook entry write cap

SpicyMarinara:
## Summary
- Recovers the quote-format caret preservation fix so typing apostrophes does not jump the cursor.
- Restores the Roleplay send-button/agent cleanup behavior so post-generation agents such as Illustrator keep the busy spinner without keeping the live stream bubble alive as a duplicate message.
- Removes the agent/tool write-path content cap for `save_lorebook_entry` and updates its tool description so large lorebook entries are preserved.
- Adds a prompt regression covering large `save_lorebook_entry` content.

## Why
The quote/caret and Illustrator cleanup fixes were local but not yet in `staging`. Users could still see the apostrophe caret jump, and Roleplay agent work could leave confusing UI state while Illustrator finished image generation.

The reported lorebook per-entry limit comes from the agent/tool write path trimming `save_lorebook_entry` content to 64 KiB before storage. The main lorebook editor schema and SQLite storage already allow larger content, so this removes the hidden agent-side truncation.

## Validation
- pnpm regression:prompt
- pnpm check
- git diff --check

## Manual verification needed
- In Roleplay mode with streaming enabled, trigger Illustrator and confirm the last text message does not appear as a duplicate while the image is generating.
- Type apostrophes in the chat input with quote formatting enabled and confirm the caret stays in place.
- Use an agent/tool write to save a lorebook entry larger than 64 KiB and confirm the stored entry is not truncated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 
 ### Fixed
 
+- Restored chat input and generation cleanup behavior so post-generation agents such as Illustrator keep the UI busy state without leaving a duplicate live-stream message visible, and preserved textarea caret position while quote formatting runs on apostrophes.
+- Removed the agent/tool write-path size cap on lorebook entry content so large entries are no longer truncated before storage.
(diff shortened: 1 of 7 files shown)


SpicyMarinara:
@coderabbitai review

PR #3221, opened 2026-07-04
feat: Reactions command toggle + fix group reaction attribution (#3219, #3220)

### LINKED ISSUE #3219: Conversation Commands: no per-command toggle for emoji reactions
**What's missing**

Emoji reactions from characters (the `[react: …]` command from #3210 / #3211) are controlled by the master **Commands** toggle in Chat Settings, but unlike Selfies, Music, Memories, etc. there is no "Reactions" card in the per-command grid underneath it. Two problems with that:

1. It's not obvious that reactions need Commands at all — if Commands is off, characters silently never react (this confused me while playtesting: models just typed emoji as message text instead).
2. You can't turn reactions off without losing every other command, or turn everything else off and keep just reactions.

**What's requested**

Add a "Reactions" toggle card to the per-command grid (Chat Settings → Commands, and the same grid in the chat setup wizard), wired up like the existing keys: when off, the react instruction is left out of the prompt and `[react:]` tags are ignored, same as Selfies/Music/etc. Default on, like the others.

**How to see the gap yourself**

1. Open a conversation chat → Chat Settings → Commands section.
2. Turn the master Commands toggle on — cards appear for Schedule Updates, Cross-Post, Selfies, Memories, Scenes, Music, Haptics, Influence, Notes, UNO, Chess… but nothing for Reactions, even though reactions ride the same pipeline.

*Follow-up to #3210, from maintainer playtesting of #3211.*

### LINKED ISSUE #3220: Group chats: character reactions are always credited to the first character
**What happens**

In a group conversation chat with Commands enabled, when characters use the `[react: …]` emoji-reaction command (#3210 / #3211), the reaction chip is always credited to the **first character in the chat** — hover the chip and the tooltip says "reacted by <first character>" every time, no matter which character was reacting in the story.

**Try it yourself**

1. Conversation chat with 2–3 characters, Commands on.
2. Chat until characters start reacting (or ask them to: "someone react to my message").
3. Hover each reaction chip. Every character reaction says "reacted by" the same character — whichever is listed first in the chat — even when the reply text makes clear a different character was the one reacting.

**Why it happens**

Merged group replies put all characters' turns in one response, split by `Name:` line prefixes, and each `[react:]` tag is attributed to the character whose section it sits in. But models consistently emit the react tag as the **first line of the whole reply, above any character's section** (confirmed by inspecting `extra.conversationCommandContent` of real replies — three out of three looked like this):

```
[react: emoji="🍿" to "User"]

Floyd Leech: oooh, corn! gimme some!
Jade Leech: It seems we have become the evening's entertainment.
Aurey Louvette: Eating while we argue... the sheer lack of decorum.
```

Text above the first `Name:` prefix falls back to the generation's primary character — the first one — in `parseCharacterCommandsBySpeaker` (`packages/server/src/services/conversation/character-commands.ts`). So a top-of-reply tag always credits character #1, deterministically.

**Suggested fix direction**

Two halves: tell models (in group chats) to write the tag inside the reacting character's own section so attribution is intentional, and make the fallback for tags above the first section attribute to the first section's speaker (the turn the tag visually opens) instead of always the primary character.

*Found while playtesting #3211; raw reply excerpts above are from the live data of that session.*

kolacheee:
Fixes #3219, fixes #3220

## Summary

Two follow-ups from maintainer playtesting of the reaction feature (#3211):

1. **Reactions toggle (#3219)**: character emoji reactions get their own card in the per-command Commands grid (Chat Settings and the chat setup wizard), wired like Selfies/Music/etc. — when off, the react instruction stays out of the prompt and `[react:]` tags are filtered by the command pipeline. Default on.
2. **Group attribution fix (#3220)**: character reactions in group chats were always credited to the first character in the chat. Real replies showed why: models emit the `[react:]` tag as the first line of the whole merged reply, above any character's `Name:` section, and that leading region fell back to the generation-primary (first) character.

## Why

While playtesting, reactions silently didn't work until the master Commands toggle was found (nothing indicates reactions depend on it), and once they worked, every chip's tooltip read "reacted by <first character>" regardless of who was reacting in the story. Inspecting `extra.conversationCommandContent` of the actual replies confirmed the tag-above-first-section placement in 3 out of 3 reacting replies.

## What changed

| Layer | Change |
|---|---|
| `packages/shared/src/types/chat.ts` | `"react"` added to `CONVERSATION_COMMAND_KEYS`. |
| `packages/server/src/routes/generate.routes.ts` | React instruction gated on the new per-command toggle; merged group chats (not individual-order mode, which forbids name prefixes) get added guidance instructing models to write the tag **inside the reacting character's own section** (never above the first name line) and stating explicitly that **multiple characters may each react in the same reply** (playtesting showed models emitting exactly one tag per merged reply, 5/5); a react targeting the persona name or "User" resolves to the user's latest message when no chat member shares the name (members take precedence); `getConversationCommandKey` maps `react` so the existing command filter enforces the toggle. |
| `packages/server/src/services/conversation/character-commands.ts` | `parseCharacterCommandsBySpeaker`: the leading region (text above the first recognized `Name:` prefix) now attributes to the speaker of the first named section — the turn the header visually opens — instead of the generation-primary character. No named sections → unchanged fallback. Affects all commands parked above the first section, not just reacts. |
| `packages/client/src/components/chat/ChatSettingsDrawer.tsx` + `ChatSetupWizard.tsx` | "Reactions" toggle card ("Let characters react to messages with emoji badges."), placed with the other command cards. |

Verified by executing the real parser: a header-placed tag now attributes to the first section's speaker, an in-section tag to that section's speaker, and a sectionless reply falls back exactly as before.

## Sweep hardening (adversarial regression/perf review of #3211 + this branch)

A multi-agent sweep (hunting the #3104/#3164 lag classes specifically) confirmed and fixed five more issues here:

- **Quadratic `stripLeadingMessageTimestamps`** on whitespace runs (measured 1.1s/call at 40KB of newlines, called per reacted history message per generation) — the leading whitespace class is now same-line-only: 100KB drops from ~6.9s to 0.68ms, output byte-identical on real timestamp shapes. Both new call sites also cap content size BEFORE stripping.
- **Redundant store scans per `[react:]` command** (each get/update is a full-table scan in the file-native store; 5 reacts on a 30k-message install measured ~2s of event-loop blocking): the chat-member build is now cached per request, the resolved target message is reused instead of re-fetched, and the swipe mirror skips the active swipe `updateMessageExtra` already writes.
- **Attribution vs leaked timestamps**: `parseCharacterCommandsBySpeaker` now segments the timestamp-stripped shape (matching the client display and the react resolver), so `[12:01] Alice: hey [react: …]` credits Alice, not the next character. Verified by execution.
- **Junk react tags**: quote-bearing bodies (`[react: she said "hi"]`) no longer mint literal-text chips — they're rejected and left visible like the old grammar; `emoji = "😂"` (spaces around `=`) now parses as intended.
- **Phantom themed box**: whitespace-only narration segments no longer render an empty `[data-card-css]` wrapper; and per-segment `ReactionAddButton`s now mount their picker + custom-emoji query subscription only while open (they render once per speaker segment across the transcript).

The sweep's remaining HIGH finding — the whole transcript re-rendering every streaming frame because `useConversationCustomEmojis/Stickers` return fresh Maps per render — **predates these PRs** (it's the #3164 disease on another prop) and is filed + fixed separately: #3223 / PR #3224.

## Known limitations

- A tag above the first section is inherently ambiguous — the model doesn't say who meant it. Attributing it to the section it opens is deterministic and positionally coherent, but the real cure is the new instruction pushing models to place tags inside their own sections; the fallback only covers non-compliant replies.
- The leading-region attribution change applies to every command type (e.g. a `[selfie]` above the first section), for consistency — flagging in case a different fallback is preferred there.

## Test plan (manual verification needed)

- [ ] Chat Settings → Commands: a "Reactions" card appears in the grid (and in the setup wizard's command step); light + dark mode, mobile width.
- [ ] With Reactions off: characters stop reacting, no raw `[react:]` text leaks into visible messages, and Peek Prompt no longer contains the react instruction.
- [x] With Reactions on in a group chat: when characters react, hover the chips — the "reacted by" tooltip credits different characters over time, not always the first one. *(Verified in live playtest against stored data — Floyd/Jade/Aurey all credited across one session; see verification report comment.)*
- [ ] Peek Prompt in a group chat shows the new placement sentence ("write the tag inside that character's own section..."); in a manual/individual-order group chat it does NOT appear.
- [x] Over a longer group session, replies sometimes carry reactions from more than one character (the multi-react guidance) — verified in raw output (`extra.conversationCommandContent`) or via chips from different reactors on the same turn. *(Verified: single replies carried 2–3 tags; chips from three characters on one message; see verification report comment.)*
- [x] Ask a character to react to your message: the chip lands on your latest message (models targeting `to "User"`/persona name now resolve explicitly). *(Verified in live playtest — models emit `to "User"` routinely and chips landed on the user's message; the collision-precedence edge is code-traced only.)*

## Docs touched

- `CHANGELOG.md` (Unreleased → Added + Fixed).

## Summary by CodeRabbit

* **New Features**
  * Added support for character reactions in Conversation mode, including a new Reactions command toggle in settings and setup.

* **Bug Fixes**
  * Improved reaction targeting in group chats so reactions are attributed to the correct character.
  * Prevented empty message segments from rendering as visible placeholders.
  * Fixed reaction parsing to ignore malformed emoji inputs and leave invalid text unchanged.
  * Improved reaction performance and responsiveness by reducing unnecessary work during message processing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 
 - Conversation mode: emoji reactions can now target an individual character's part of a merged multi-character reply. Each speaker segment gets its own add-reaction button (next to the name) and its own reaction row; in the prompt, characters see reactions as a `[User reacted with 😹]` line directly under the exact part they were aimed at. Reactions without a segment target keep applying to the whole message, so existing chats and 1:1 conversations are unaffected (#3210).
 - Conversation mode: characters can now react to each other, not just to the user. A character writing `[react: emoji="🙄" to "Character Name"]` puts the reaction on that character's most recent part (same reply or recent history), rendered as a chip under it and shown to everyone in the prompt (e.g. `[User reacted with 😹, Aurey reacted with 🙄]`).
+- Conversation mode: character emoji reactions now have their own **Reactions** card in the per-command Commands grid (Chat Settings and the chat setup wizard), so they can be toggled independently like Selfies or Music (#3219).
 
 - Added the Agent Suite to the Chat Settings drawer's Agents section: a window listing the agents active in the current chat where you can view and edit everything they have stored — agent memory, tracker state, and custom-agent outputs — manually or with AI-assisted rewrites (select text, give an instruction, optionally attach grounding context such as character cards or active-lorebook entries, and pick a connection) (#3160).
 - Added first-class scene video generation for Game Mode, Roleplay, and Visual Novel galleries, including Video Generation connections for Gemini Omni and xAI Imagine, editable `game.video` prompts, manual Gallery video actions, per-image Animate buttons, Gallery video previews with prompt copy, live View Latest media, and draggable/resizable pinned video overlays.
@@ -16,6 +17,7 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 
 ### Fixed
 
+- Fixed group-chat character reactions always being credited to the first character in the chat: commands placed above the first `Name:` line of a merged reply now attribute to the speaker whose section they open, group chats now instruct models to write the `[react:]` tag inside the reacting character's own section, and a react aimed at the user's persona name (or "User") explicitly targets the user's latest message (#3220).
 - Renamed the editable scene-video prompt template from `game.omniVideo` to `game.video`, with legacy override fallback, and shortened scene-video prompts for smaller video providers by summarizing narration into a compact story beat, excerpting source illustration prompts, and loosening default motion guidance.
 - Removed the hard-coded three-sprite limit from Roleplay sprite selection, setup, and display paths so chats can enable all uploaded sprite owners they need (#3169).
 - Let Image Captioning use any non-image-generation connection instead of hiding local or custom multimodal models behind model-name heuristics (#3170).
--- a/packages/client/src/components/chat/ChatSettingsDrawer.tsx
+++ b/packages/client/src/components/chat/ChatSettingsDrawer.tsx
@@ -406,6 +406,11 @@ const CONVERSATION_COMMAND_TOGGLE_OPTIONS: Array<{
     description: "Let characters ring you for a Conversation call.",
   },
   {
+    id: "react",
+    label: "Reactions",
+    description: "Let characters react to messages with emoji badges.",
+  },
+  {
     id: "uno",
     label: "UNO",
     description: "Let characters start a game of UNO at the table when you agree to play.",
--- a/packages/client/src/components/chat/ChatSetupWizard.tsx
+++ b/packages/client/src/components/chat/ChatSetupWizard.tsx
@@ -164,6 +164,7 @@ const CONVERSATION_COMMAND_TOGGLE_OPTIONS: Array<{
   { id: "influence", label: "Influence", description: "Let characters influence a connected chat." },
   { id: "note", label: "Notes", description: "Let characters save durable notes for a connected chat." },
   { id: "call", label: "Calls", description: "Let characters ring you for a Conversation call." },
+  { id: "react", label: "Reactions", description: "Let characters react to messages with emoji badges." },
   { id: "uno", label: "UNO", description: "Let characters start a game of UNO at the table when you agree to play." },
   { id: "chess", label: "Chess", description: "Let characters accept a one-on-one chess challenge at the table." },
 ];
--- a/packages/client/src/components/chat/ConversationMessageGrouped.tsx
+++ b/packages/client/src/components/chat/ConversationMessageGrouped.tsx
@@ -196,6 +196,10 @@ export function ConversationMessageGrouped({
             ) : null;
 
           if (!grp.speaker) {
+            // Whitespace-only narration: render nothing — an empty
+            // [data-card-css] wrapper would paint a phantom themed box, the
+            // same invariant the speaker branches guard with segHasText.
+            if (!segHasText) return null;
             return (
               <div
                 key={i}
--- a/packages/client/src/components/chat/ReactionAddButton.tsx
+++ b/packages/client/src/components/chat/ReactionAddButton.tsx
@@ -4,7 +4,7 @@
 // user's reaction on the message. Standard emojis come from the picker; the global
 // custom emojis are shown in a pick-only "Custom" grid. Conversation mode only.
 // ──────────────────────────────────────────────
-import { useRef, useState } from "react";
+import { useRef, useState, type RefObject } from "react";
 import { SmilePlus } from "lucide-react";
 import { cn } from "../../lib/utils";
 import { EmojiPicker } from "../ui/EmojiPicker";
@@ -21,7 +21,6 @@ interface ReactionAddButtonProps {
 export function ReactionAddButton({ onPick, className, tabIndex }: ReactionAddButtonProps) {
   const [open, setOpen] = useState(false);
   const buttonRef = useRef<HTMLButtonElement>(null);
-  const { data: customEmojis } = useCustomEmojis();
 
   return (
     <>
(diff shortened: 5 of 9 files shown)


kolacheee:
## Verification report

Record of what was actually tested on this branch, split by kind. Everything in the first two sections is reproducible; commands/inputs included. Items **not** covered are listed at the end so review effort goes there.

### 1. Live playtest (real install, group chat with 3 characters; dev server running this branch)

Verified against stored data (`extra.conversationCommandContent` = raw model output with tags, `extra.reactions` = persisted chips), not just screenshots:

- **The #3220 bug reproduced first**: before the fix, 3 out of 3 reacting replies carried the `[react:]` tag above the first `Name:` line, and every chip was credited to the chat's first character regardless of who reacted in-story.
- **Attribution fix + placement instruction working**: in the post-fix session, models placed tags *inside* character sections (e.g. `Floyd Leech: [react: emoji="🤣" to "Aurey Louvette"]`), and chips credited the actual reactor. A same-reply character→character react resolved to the target's part *above* the reactor's turn (the narrative-causality cutoff), e.g. Floyd's 🤣 landed on Aurey's "tactical retaliation" part as a segment chip.
- **Multi-react instruction working**: after the instruction commit, single merged replies carried 2–3 react tags — one user message ended up with chips from Floyd 😸 **and** Jade 🧐, another with Floyd 🤣 + Aurey 🙄 + Jade 🤭 from one generation. Untargeted mid-section reacts defaulted to the user's latest message; a leftover header tag was credited to the section it opens.
- **`to "User"` targets land on the user's message** — models consistently emit `to "User"` and the chips landed correctly (this is the path the explicit persona/"User" resolution now owns).
- Context for #3219: with the master Commands toggle off, models emitted literal emoji *as message text* instead of tags — reactions silently dead, which is why the per-command card + discoverability matters.

### 2. Executed against the real code (scripted, reproducible)

Run from `packages/server` with `npx tsx` importing the actual modules (not reimplementations), unless noted:

| Test | Input | Result |
|---|---|---|
| Attribution: leaked timestamp | `parseCharacterCommandsBySpeaker('[12:01] Alice: hey [react: emoji="X"]\nBob: hi', [Alice, Bob], 'alice')` | credited **alice** (pre-fix: bob) |
| Attribution: header tag | tag above first section (the observed model shape) | credited first section's speaker; sectionless reply → unchanged fallback |
| Attribution: in-section tag | `Bob: hi [react: emoji="B" to "Alice"]` | credited **bob**; ran again after the staging merge — identical |
| Grammar parity | old forms `emoji="😂"` / `"😂"` / bare / `:name:` + targeted forms quoted/bare | all parse as before; targets extracted; strip leaves no residue |
| Junk rejection | `[react: she said "hi" loudly]` | **no command, tag stays visible** (old grammar parity; previously became a literal-text chip) |
| Malformation tolerance | `[react: emoji = "😂"]` (spaces around `=`) | parses as 😂 |
| ReDoS regression test | `'[react: ' + ' '.repeat(200_000)`, no `]` | **0.56 ms** (the reviewed-out draft regex: 21 s at 4 000 chars) |
| Timestamp-strip quadratic | `stripLeadingMessageTimestamps('\n'.repeat(100_000))` | **0.68 ms** after fix (before: ~6.9 s; measured 69 ms @ 10K → 170 s @ 500K, clean O(n²)) |
| Timestamp-strip parity | leading/indented/stacked `[HH:MM]`/date tokens, mid-line tokens, `Name: ` lines | output byte-identical to the old regex on all shapes |

### 3. Adversarial review rounds (multi-agent, each finding independently double-verified against the code)

Three rounds ran over this branch: one on the initial commit (caught the individual-order prompt leak and the persona-name precedence, both fixed here), and a combined regression/perf sweep over #3211+this branch hunting the #3104/#3164 lag classes specifically. Sweep verifiers benchmarked the real file-native store: one `[react:]` persisted through ~10–14 full-table scans (**118 ms/react at 10k stored messages, 351 ms at 30k — ~2 s of event-loop blocking for a 5-react reply**), which motivated the member-cache/target-reuse/active-swipe-skip commit. The sweep's remaining HIGH finding predates these PRs and is filed separately (#3223 → PR #3224).

### 4. Merge-conflict resolution (v2.1.0 release landed on staging mid-branch)

CHANGELOG was the only hard conflict: resolved by dropping the branch's Unreleased entries that shipped in 2.1.0 (each one confirmed present in the released section before deletion) and keeping only this PR's entries. The two auto-merged server files were checked hunk-by-hunk (staging's changes are in Professor Mari preset parsing + unrelated line ranges), and the attribution tests above were re-run post-merge. `pnpm check` and `pnpm version:check` green.

### Not verified — please cover in review

- The **Reactions toggle card UI** (Chat Settings grid + setup wizard): added after the playtest session, never seen in a browser. Light/dark, mobile width.
- **Toggle OFF end-to-end**: instruction absent from Peek Prompt, no raw tag leaks (code-traced only).
- The **individual-order group chat** gating (placement sentence absent there) — code-traced only.
- Persona-name collision precedence (character literally named like the persona stays targetable) — code-traced only.

kolacheee:
@coderabbitai review

PR #3224, opened 2026-07-05
fix: stop conversation emoji/sticker maps re-rendering the whole transcript every streaming frame (#3223)

### LINKED ISSUE #3223: Conversation mode: whole transcript re-renders every streaming frame (unmemoized emoji/sticker maps defeat message memo)
**What happens**

In conversation mode, every mounted message re-renders on every frame of a streaming reply (React DevTools "highlight updates" shows the whole transcript flashing per token frame). On long chats (100+ mounted messages, group chats with multi-speaker replies) this costs roughly 15–35ms per frame on a desktop — visible stutter during generation, worse on laptops or when a local model is saturating the same CPU.

**Try it yourself**

1. Conversation chat with a long history (scroll up so 80+ messages are mounted); custom emojis/stickers don't need to be configured — the bug is unconditional.
2. Open React DevTools → Components → enable "Highlight updates when components render".
3. Send a message. While tokens stream in, every message in the transcript flashes on every frame, not just the streaming one.

**Why it happens**

`useConversationCustomEmojis()` and `useConversationCustomStickers()` (`packages/client/src/hooks/use-conversation-custom-emojis.ts` / `use-conversation-custom-stickers.ts`) build a fresh `list` array and a fresh `Map` on every render — there's no memoization, and the `useQueries` inputs churn identity per render anyway. `ConversationView` passes those Maps as `emojiMap`/`stickerMap` props to every `memo(ConversationMessage)`, so the memo's shallow compare fails for all mounted messages on every `ConversationView` render — and ConversationView renders once per animation frame during any token stream. This is the same class as #3164/#3104 ("cheap per-message work × every mounted message × every render"): each message then re-runs its markdown/custom-emoji/mention pipeline per segment plus (since #3211) its per-segment reaction-button subtrees, measured at ~3.3ms of pure element work per pass for 100 messages × 4 segments before reconciliation.

`ChatArea` already fixed the identical disease for `chatCharacterRows` in #3166 with a ref-backed element-wise reuse (ChatArea.tsx ~750) — these two hooks are the remaining identity-breakers on the ConversationMessage prop surface (characterMap, personaInfo and the callbacks are already stable).

**Suggested fix**

Apply the same row-stability pattern inside both hooks: compare the freshly-built list element-wise against a ref-cached previous one (all fields are primitives) and return the cached `{ list, map }` object when unchanged.

*Found by a regression/perf sweep of the reaction PRs (#3211/#3221) — the bug predates them, but per-segment reactions multiplied the work living under the broken memo boundary.*

kolacheee:
Fixes #3223

## Summary

`useConversationCustomEmojis()` and `useConversationCustomStickers()` returned a fresh `list` array and a fresh `Map` on every render. `ConversationView` passes those Maps as `emojiMap`/`stickerMap` props to every `memo(ConversationMessage)`, so the memo's shallow compare failed for **every mounted message on every ConversationView render** — and ConversationView renders once per animation frame during any token stream. Net effect: the whole transcript re-rendered per streaming frame (~15–35 ms/frame measured on a 100-message group chat), the same disease class as #3104/#3164.

The fix applies the ref-backed reuse pattern ChatArea already uses for `chatCharacterRows` (the #3164 fix): build the resolved list as before, compare it element-wise against the previous one (all fields are primitives), and return the cached `{ list, map }` object when unchanged. Both hooks fixed identically.

## Why

Found by a regression/performance sweep of the reaction PRs (#3211/#3221). The bug predates them — but per-segment reactions added more per-message subtrees under the broken memo boundary, so fixing the boundary is what makes the transcript cheap again during streaming.

## Known limitations

- The element-wise compare is O(emoji+sticker count) per render — trivial, and the price of the `useQueries` inputs having no stable identity to memo against (a plain `useMemo` cannot work here, which is why the ref pattern exists).

## Test plan (manual verification needed)

- [ ] React DevTools → "Highlight updates when components render": in a long conversation chat, stream a reply — only the streaming message (and composer chrome) should flash per frame, not the whole transcript. Before this fix, every mounted message flashed.
- [ ] Custom emojis/stickers still render in messages, the composer autocomplete, the picker's Custom tab, and the sticker selector; adding/renaming/deleting a custom emoji updates them (the cache must invalidate on real changes).
- [ ] Long group chat scroll + streaming feels smoother on a loaded machine (subjective, but the measurable proxy is the DevTools flashing).

## Docs touched

- None (internal perf fix; no user-facing behavior change).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency when rendering conversations by keeping custom emoji and sticker results stable across re-renders.
  * Reduced unnecessary updates by reusing previously computed emoji and sticker collections when the content hasn’t changed.
  * Helped preserve UI memoization, which can prevent flicker or extra rendering work in transcript views.

### DIFF
--- a/packages/client/src/hooks/use-conversation-custom-emojis.ts
+++ b/packages/client/src/hooks/use-conversation-custom-emojis.ts
@@ -5,6 +5,7 @@
 // emojis override a global of the same name (owner-wins). Used by the message
 // renderer, the composer autocomplete, and the picker's Custom tab.
 // ──────────────────────────────────────────────
+import { useRef } from "react";
 import { useQueries } from "@tanstack/react-query";
 import { api } from "../lib/api-client";
 import { useChatStore } from "../stores/chat.store";
@@ -36,6 +37,7 @@ function displayName(row: { data?: unknown; name?: unknown } | undefined, fallba
 }
 
 export function useConversationCustomEmojis(): { list: ConversationCustomEmoji[]; map: Map<string, string> } {
+  const stableResultRef = useRef<{ list: ConversationCustomEmoji[]; map: Map<string, string> } | null>(null);
   const activeChatId = useChatStore((s) => s.activeChatId);
   const { data: activeChat } = useChat(activeChatId);
   const personaId = (activeChat as { personaId?: string | null } | undefined)?.personaId ?? null;
@@ -90,7 +92,31 @@ export function useConversationCustomEmojis(): { list: ConversationCustomEmoji[]
     }
   });
 
-  const list = [...byName.values()];
-  const map = new Map(list.map((emoji) => [emoji.name, emoji.url] as const));
-  return { list, map };
+  // [#3223] The derivation above rebuilds fresh objects every render (the
+  // useQueries result arrays churn identity per render), but the resolved set
+  // rarely changes. Reuse the previous { list, map } when it is element-wise
+  // identical — every mounted ConversationMessage receives `map` as a prop, so
+  // a fresh Map per render broke React.memo for the whole transcript on every
+  // streaming frame (the #3164 disease; same ref-reuse pattern as ChatArea's
+  // chatCharacterRows).
+  const next = [...byName.values()];
+  const previous = stableResultRef.current;
+  if (
+    previous &&
+    previous.list.length === next.length &&
+    next.every((emoji, index) => {
+      const before = previous.list[index]!;
+      return (
+        before.name === emoji.name &&
+        before.url === emoji.url &&
+        before.source === emoji.source &&
+        before.editable === emoji.editable
+      );
+    })
+  ) {
+    return previous;
+  }
+  const result = { list: next, map: new Map(next.map((emoji) => [emoji.name, emoji.url] as const)) };
+  stableResultRef.current = result;
+  return result;
 }
--- a/packages/client/src/hooks/use-conversation-custom-stickers.ts
+++ b/packages/client/src/hooks/use-conversation-custom-stickers.ts
@@ -5,6 +5,7 @@
 // stickers override a global of the same name (owner-wins). Used by the message
 // renderer and the sticker selector.
 // ──────────────────────────────────────────────
+import { useRef } from "react";
 import { useQueries } from "@tanstack/react-query";
 import { api } from "../lib/api-client";
 import { useChatStore } from "../stores/chat.store";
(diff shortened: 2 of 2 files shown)


kolacheee:
## Verification report

What was actually tested for this fix, and what still needs a browser pass.

### Root cause, verified in the code by two independent review passes

- `useConversationCustomEmojis()` / `useConversationCustomStickers()` build a fresh `list` array and fresh `Map` in the render body with no memoization (the `useQueries` inputs churn identity every render, so a plain `useMemo` keyed on them could never hold — which is why the ref pattern is required).
- `ConversationView` passes both Maps as `emojiMap`/`stickerMap` props into every `memo(ConversationMessage)`; the shallow compare therefore fails for **every mounted message on every ConversationView render**.
- `useThrottledStreamBuffer` re-renders ConversationView once per animation frame during any token stream (per character in typewriter mode) → the whole transcript re-renders per frame. All other props on the `ConversationMessage` surface were checked and are identity-stable (`characterMap` via ChatArea's #3164 fix, `personaInfo`, callbacks), so these two hooks are the only remaining memo-breakers.

### Measured cost (scripted benchmark of the real render-path code)

An esbuild-bundled bench of the actual message-content pipeline (markdown + custom-emoji + sticker + mention passes) measured **~3.3 ms of pure element/regex work per full-transcript pass** at 100 messages × 4 segments — before React reconciliation of ~100 message subtrees plus the per-segment reaction-button subtrees added by #3211. Realistic full re-render estimated at 15–35 ms/frame on desktop, i.e. visible streaming stutter, worse with a local model saturating the same CPU.

### Fix verification

- The reuse pattern is the same one already proven in `ChatArea.chatCharacterRows` (the #3164 fix): element-wise compare of primitives (`name`/`url`/`source`/`editable`) against a ref-cached previous list; identical → return the cached `{ list, map }` object, otherwise cache and return fresh (so real emoji/sticker changes still propagate to all consumers — messages, composer autocomplete, picker Custom tab, sticker selector).
- Both hooks fixed identically; `pnpm check` (TypeScript strict + ESLint + production build) green.

### Not verified — please cover in review

- The headline test needs a browser: **React DevTools → "Highlight updates when components render"**, long conversation chat, stream a reply. Before this fix the entire transcript flashes every frame; after, only the streaming message (and composer chrome) should. This is the one-look confirmation the fix works.
- Custom emoji/sticker rendering after add/rename/delete (cache must break exactly when content changes) — code-traced only.

kolacheee:
@coderabbitai review

PR #3228, opened 2026-07-05
[codex] Fix Mari parser flags and Game dynamic image prompts

### LINKED ISSUE #3222: [Issue]: mari parser omits some boolean flags from BOOLEAN_FLAGS, silently swallowing positionals
## Summary

Prof Mari's argument parser only treats a flag as valueless if its name is in a hardcoded `BOOLEAN_FLAGS` set, but 13 flags that are used purely as booleans are missing from that set. When any of them appears immediately before a non-flag token, the parser consumes that token as the flag's "value", silently discarding a required positional.

## Expected behavior

`mari chats messages --tail <chat-id>` should list the tail of the chat's messages regardless of whether `--tail` comes before or after the `<chat-id>` positional, since `--tail` is a boolean flag.

## Actual behavior

`--tail` swallows `<chat-id>` as its value. `positionals[0]` is then `undefined`, so the command throws its usage error (`Usage: mari chats messages <chat-id> …`) even though a chat id was supplied. `hasFlag(flags, "tail")` still returns `true` (the stored value isn't `false`), so the flag reads as "set" while having destroyed the positional.

## Steps to reproduce

1. Run any `mari` subcommand that takes a positional, with a missing-from-set boolean flag placed *before* that positional. Minimal case: `mari chats messages --tail <chat-id>`.
2. Observe the command fails with the `Usage: mari chats messages <chat-id> …` error.
3. Compare with `mari chats messages <chat-id> --tail`, which works — the only difference is flag order.

## Environment

- Marinara Engine version: 2.1.0
- Install type: source
- OS + version: N/A — static source review at commit `a90e7158` (branch `staging`)
- Browser or app shell: N/A

None

## Additional context

Identified by source review; **not yet reproduced against a running server**.

Root cause: `parseArgs` treats a `--flag` as valueless only when its name is in `BOOLEAN_FLAGS` (`packages/server/src/services/mari-db/mari-db.service.ts:628`), otherwise it consumes the next token (`:629`). `BOOLEAN_FLAGS` (`:139`) lists 9 names, but `hasFlag(...)` is used as a pure boolean for 13 more that aren't listed: `tail`, `global`, `enable`, `disable`, `constant`, `no-constant`, `changed`, `cached`, `staged`, `patch`, `full`, `help`, `no-global`. None of the 13 is ever read as a value via `flagString` (verified across the file), so none has any reason to be excluded. The sibling parser at `packages/server/src/services/mari-db/mari-images.service.ts:106` already lists `help` in its own `BOOLEAN_FLAGS`, confirming these are meant to be boolean. Nothing in the per-command usage strings, the top-level help, or the Professor Mari system prompt instructs the model to place flags after positionals, so the safe ordering is not guaranteed.

Suggested fix: add the 13 names to `BOOLEAN_FLAGS` at `mari-db.service.ts:139`. `BOOLEAN_FLAGS` is referenced only by `parseArgs` (`:628`), so the change is side-effect-free.

## Template check

### LINKED ISSUE #3225: [Feature]:  Toggle for Dynamic LLM Prompt Generation in GM Mode Assets

The current GM Mode image generation relies on rigid, hardcoded prose templates in the prompt review panel (e.g., aggressively forcing a humanoid fallback rule: *"Unless the description says otherwise, do not infer an animal species..."*). 

While this prevents wild hallucinations on cloud image APIs, it completely breaks compatibility for local text-to-image pipelines. Local diffusion checkpoints—especially fine-tunes like *Illustrious* or *SDXL*—rely heavily on dense tag weights and completely choke on conversational, negative prose structures. 

Furthermore, applying a hardcoded "humanoid/person" bias template universally across GM Mode bleeds into and muddies up non-character assets like environmental backgrounds and cinematic action scenes. By trying to hardcode these limits to save minor text token overhead, the current setup ironically makes the feature work *only* with massive cloud LLMs, locking out local hardware setups.

## Proposed solution

Add an optional settings toggle under the Illustrator / Image Generation configuration: 
`[ ] Enable Dynamic LLM Prompt Generation for GM Mode Assets`

When enabled, instead of wrapping the asset context in a hardcoded prose template, the engine sends a quick, dedicated text API call to the connected LLM. The LLM is tasked with translating the active context into a highly optimized positive prompt string before passing it to the image provider.

This should apply universally across all three image generation triggers in GM Mode:
* **NPC Portraits:** Allows clean tag weights for diverse character species/races without arbitrary human-bias text overrides.
* **Backgrounds:** Prevents hardcoded character/humanoid rules from bleeding into and ruining environmental landscape generation.
* **Key Moments:** Gives the LLM the freedom to describe complex cinematic action prompts dynamically without being chained to a static template.

## Alternatives considered

Leaving the templates as hardcoded text strings was considered, but it fundamentally breaks local image generation workflows. Because local models require specialized formatting (like raw **Danbooru / e621 style tag layouts**), a static conversational template cannot serve both cloud APIs and local checkpoints simultaneously. 

## Additional context

This infrastructure isn't a massive structural shift for the codebase—it is already proven to work seamlessly within Marinara Engine's own architecture via the **"Selfie responses"** mechanism in normal Chat Mode. This suggestion simply extends that flexible, on-demand pipeline to GM Mode.

Additionally, making this an *optional toggle* completely protects the engine's token-conservation philosophy. Users who are highly sensitive to text API credits can leave it off and use the default baked-in string. Power users running local hardware or who don't mind the minor token cost can choose to flip it on.

## Template check

SpicyMarinara:
## Summary

- Fixes Mari CLI boolean flag parsing so flags such as `--tail`, `--raw`, `--patch`, `--strict`, and `--staged` no longer consume positional arguments.
- Adds an optional Game Illustrator setting, **Dynamic LLM Prompt Generation for GM Mode Assets**, for NPC portraits, location backgrounds, and key-moment scene illustrations.
- Registers `game.imagePromptDirector` so the dynamic rewrite prompt is editable, applies it to prompt previews and final generation, and falls back to deterministic prompts if the LLM rewrite fails.
- Updates the v2.1.0 changelog.

Fixes #3222.
Fixes #3225.

## Validation

- `pnpm --filter @marinara-engine/server lint`
- `pnpm --filter @marinara-engine/client lint` (passes with existing `ConversationView.tsx` exhaustive-deps warning)
- `pnpm regression:prompt`
- `pnpm check` (passes with existing client lint warning and Vite large-chunk warning)

## Manual Verification Needed

- Manually verify Game Mode Chat Settings shows the new Dynamic LLM Prompt Generation toggle under Illustrator.
- Manually verify prompt review previews use the rewritten prompt when the toggle is on.
- Manually verify generated NPC portraits, backgrounds, and key-moment illustrations fall back to old prompts if the prompt-model call fails.

## Summary by CodeRabbit

* **New Features**
  * Added an optional toggle for dynamic prompt generation for game assets, letting generated portraits, backgrounds, and illustrations be rewritten before image creation.
  * Expanded game asset prompt handling so image requests can use more context-aware prompts when enabled.

* **Bug Fixes**
  * Fixed CLI parsing so common boolean flags no longer consume the following positional argument, improving command reliability.


SpicyMarinara:
@coderabbitai review

PR #3241, opened 2026-07-05
fix: use mari-message-content styles for File Browser markdown preview

kolacheee:
Fixes #3240

## Summary

One-line class swap in `FileEditorModal.tsx`: the markdown preview wrapper used Tailwind `prose prose-sm max-w-none` classes, but `@tailwindcss/typography` is not installed anywhere in the client, so those classes were inert and the preview rendered as unstyled plain text. The wrapper now uses `mari-message-content whitespace-pre-wrap break-words text-sm`, the same pattern every working markdown consumer uses (`ChatMessage`, `GameJournal`, `HomeProfessorMariChat`).

## Why

Users previewing `.md` files in the Game Assets file browser see headings, lists, and tables collapse into undifferentiated plain text, while the exact same markdown renders correctly in chat. All of the markdown renderer's output styles (`.mari-md-heading`, `.mari-md-ul`, `.mari-md-table`, …) are scoped under `.mari-message-content` in `globals.css`, so the wrapper must carry that class for any styling to apply. The renderer also joins paragraph runs with `\n`, which is why the wrapper additionally needs `whitespace-pre-wrap` — without it, paragraph spacing collapses.

## Test plan

- [x] Manually verify in browser: open a `.md` file in the Game Assets file browser, switch to Preview, and confirm headings, bullet lists, and tables are styled (matching how the same markdown looks in a chat message)
- [x] Manually verify paragraph spacing: a file with multiple paragraphs separated by blank lines shows visible spacing between them
- [x] Manually verify in both light and dark mode
- [x] Manually verify at a narrow/mobile viewport (~400px) that long lines wrap instead of overflowing the modal

`pnpm check` passes locally.

## UI Evidence

image

image

## Summary by CodeRabbit

* **Style**
  * Updated the Markdown preview appearance in the file editor to use consistent message content styling while preserving wrapping, text sizing, and color behavior.

PR #3243, opened 2026-07-05
fix: parse markdown headings in files with CRLF line endings

### LINKED ISSUE #3242: Markdown headings render as literal # text when a file has Windows (CRLF) line endings
**What happens**

When the markdown renderer is given text with Windows-style line endings (CRLF, `\r\n`), headings are not converted — the preview shows the literal text `# Game Assets Credits` at body-text size instead of a styled heading. Tables, bold, and inline code in the same file render fine, which makes the broken headings stand out.

**Try it yourself** (on a Windows install)

1. Open **Settings → Game Assets → Asset Browser**.
2. Open the seeded `CREDITS.md` (every install has one) — it opens straight into Preview.
3. The tables render with borders, but every heading shows as a literal `# ...` / `## ...` line.
4. For contrast: create a new markdown file in the same browser, type `# Hello`, and preview it — that heading converts fine, because files created in-app use plain `\n` line endings.

**Why it happens**

`renderMarkdownBlocks` in `packages/client/src/lib/markdown.tsx` splits content on `"\n"` only (line 378). For a CRLF file, every line keeps a trailing `\r`. The heading regex `/^(#{1,6})\s+(.+)$/` can't match across that carriage return (in JavaScript regexes, `.` and `$` stop at `\r`), so heading lines fall through and render as plain paragraph text. The table parser happens to call `.trim()` on each line first — which strips the `\r` — so tables work, headings don't.

The seeded `CREDITS.md` gets CRLF endings on Windows because git converts text files at checkout and the seeder copies the file byte-for-byte. Any markdown file written by a Windows editor and dropped into game assets hits the same problem. Chat messages are unaffected (typed/streamed content uses `\n`).

**Fix**

One-line change: split on `/\r?\n/` instead of `"\n"` so CRLF and LF input parse identically. This also protects the other line-anchored patterns in the block parser that don't trim. Related: #3240 fixed the preview's missing styles; this bug is independent of that one (the heading `#` marks survive even with styling fixed). I'm taking this one (kolacheee) — PR to `staging` incoming.

kolacheee:
Fixes #3242

## Summary

`renderMarkdownBlocks` split content on `"\n"` only, so text with Windows CRLF line endings left a trailing `\r` on every line. The heading regex (`/^(#{1,6})\s+(.+)$/`) cannot match across a carriage return, so headings in CRLF files rendered as literal `# ...` plain text. Now the renderer splits on `/\r?\n/`, making CRLF and LF input parse identically.

## Why

The seeded `CREDITS.md` in Game Assets arrives with CRLF endings on Windows (git converts it at checkout, and the seeder copies it byte-for-byte), so every Windows install shows broken headings when previewing it — while tables in the same file render fine, because the table parser happens to `.trim()` each line. Any markdown file written by a Windows editor and imported into game assets hits the same problem. Fixing it at the line-split also protects the parser's other line-anchored patterns that don't trim, instead of patching regexes one by one.

## Known limitations

The separate `applyInlineMarkdownHTML` path (sanitized-HTML pipeline where newlines are already `<br>` tags) is untouched — whether `\r` can survive into that pipeline depends on the upstream sanitizer and there is no observed repro there.

Note: the full styled result in the file preview also needs #3241 (the preview's CSS scoping fix, separate PR). This fix is observable independently: with it, the `#` marks are consumed and real `<h1>`–`<h6>` elements are emitted.

## Test plan

- [x] Manually verify in browser (Windows install): Settings → Game Assets → Asset Browser → open the seeded `CREDITS.md` → Preview shows "Game Assets Credits", "Music", etc. as headings with no literal `#` marks
- [x] Manually verify no regression for LF content: create a new markdown file in-app with `# Heading`, a list, and a table — everything still renders as before
- [x] Manually verify a code fence in a CRLF file still renders as a code block
- [x] Capture before/after screenshots for this PR

`pnpm check` passes locally.

## UI Evidence

How `credits.md` appears now:
image

With both this and #3241 applied:
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved markdown rendering to handle both Windows and Unix line endings consistently.
  * Fixed an issue where some headings and other line-based markdown patterns could fail to render correctly when content included carriage returns.

### DIFF
--- a/packages/client/src/lib/markdown.tsx
+++ b/packages/client/src/lib/markdown.tsx
@@ -375,7 +375,10 @@ export function renderMarkdownBlocks(
   renderInline: (text: string, keyPrefix: string) => ReactNode[] = applyInlineMarkdown,
   keyBase = "md",
 ): ReactNode {
-  const lines = normalizeCardAssetImageSyntax(text).split("\n");
+  // Split on CRLF or LF — CRLF input (e.g. .md files written on Windows) must
+  // not leave a trailing \r on lines, or line-anchored patterns like
+  // HEADING_RE fail to match.
+  const lines = normalizeCardAssetImageSyntax(text).split(/\r?\n/);
   const segments: ReactNode[] = [];
   let key = 0;
 


kolacheee:
@coderabbitai review

PR #3388, opened 2026-07-08
Fix v2.1.2 issue sweep

### LINKED ISSUE #3375: [Issue]: 'Remove from chat' on a lorebook that's also active via persona/character/global only unpins it — it keeps injecting, and 'Disable in this chat' is hidden while pinned
**What's wrong**

In Chat Settings → Lorebooks, each active lorebook row has a **single** action button with three mutually-exclusive states. For a book that is both **pinned to the chat** and **also active via a persona (or character/global) link**, the row shows only the 🗑️ **"Remove from chat"** button — not the 👁️ **"Disable in this chat"** button. Clicking "Remove from chat" only removes the chat pin; the book keeps injecting because the persona link still activates it. There is no single action that fully turns the book off, and "Remove from chat" reads as "this book is gone" when it isn't.

This is the likely cause of the user report *"I unlinked lorebook A from the chat and it still behaved like it had read A."* Unlinking (Remove from chat) does not stop a persona-linked book — only the separate "Disable in this chat" (👁️, which writes the per-chat exclusion) does. And that eye is hidden while the book is pinned, so the user has to: click trash → not realize the book is still active → later spot the "Persona" pill → click the now-appeared eye.

(To be clear about scope: the row already shows *why* a book is active via small pills — "Chat", "Persona", "Character", "Global" — so the persona activation is not invisible; and the injection gate itself works correctly. The gap is purely this confusing two-step control and the misleading "Remove from chat" label for a book kept alive by another source.)

1. Assign a lorebook to a persona (Persona editor → Lorebooks → Assign).
2. Start a chat with that persona, and also pin the same lorebook to the chat, so the row shows both a **"Chat"** and a **"Persona"** pill and a 🗑️ trash button.
3. Click the trash ("Remove from chat").

**You'll see:** the pin is removed, but the book is still listed as active (now just the "Persona" pill) and still injects — and the button silently changes to a 👁️ "Disable in this chat". Nothing told you it was still active.
**Expected:** one clear way to fully turn a book off for this chat — or, at minimum, after "Remove from chat" on a still-active book, a signal that it's still active via the persona and how to disable it.

**Where it is**

`packages/client/src/features/chat-settings/sections/LorebooksSection.tsx:127-151` — the per-row control:
- `isExcluded` → 👁️ "Enable in this chat" (`onSetLorebookExcluded(id, false)`), lines 127-134
- else `isPinned` → 🗑️ "Remove from chat" (`onToggleLorebook(id)`, unpin only), lines 135-142
- else → EyeOff "Disable in this chat" (`onSetLorebookExcluded(id, true)`), lines 143-150

Because the `isPinned` branch wins over the exclude branch, a pinned + persona-active book never shows the disable control. The "why" pills are rendered at 110-125 (`lorebook.activeReasons`). Unpin writes `activeLorebookIds`; disable writes `excludedLorebookIds` (`ChatSettingsDrawer.tsx` `setLorebookExcluded` ~2387). The server injection gate honors `excludedLorebookIds` correctly (`packages/server/src/services/lorebook/index.ts:246`, dropped before the persona check), so "Disable in this chat" genuinely stops injection in every mode — it's just hard to reach for a pinned book.

**What's happening**

The single-button, three-way design assumes a book is either pinned, excluded, or auto-active — not pinned **and** auto-active-via-another-source at once. So it offers "Remove from chat" (unpin) but hides "Disable in this chat" (exclude) for exactly the case where unpinning alone doesn't stop injection.

**Suggested fix**

For a book active via a non-chat source (persona/character/global), make the full-disable reachable in one step: show the 👁️ "Disable in this chat" control even when the book is pinned (e.g. render both a "Remove pin" and a "Disable" action, or make a single "Disable in this chat" that also clears the pin by adding to `excludedLorebookIds`). At minimum, after "Remove from chat" on a book still active via another source, surface an inline hint — "Still active via persona <name> — Disable in this chat?" — wired to `setLorebookExcluded`. No server change needed; the exclusion gate already works.

---
*AI-investigated against the current tree (v2.0.9); the control branches, the reason pills, and the server exclusion gate all read and confirmed. Please verify before fixing.*

### LINKED ISSUE #3376: [Issue]: Game-mode world generation ignores the per-chat 👁️ lorebook disable (excludedLorebookIds not passed to processLorebooks)
**What's wrong**

Every lorebook resolution in the app honors the per-chat 👁️ "Disable in this chat" toggle (`excludedLorebookIds`) — except one: **Game-mode world generation**. A lorebook you disabled with the eye can still be scanned into game world-gen.

1. In a Game-mode chat, have a persona/character/global lorebook active, and click 👁️ "Disable in this chat" on it in Chat Settings → Lorebooks.
2. Trigger game setup / world generation.

**You'll see:** the disabled lorebook's content is still pulled into world generation (verify via debug / Peek Prompt). Everywhere else (normal turns, previews, voice calls) it's correctly excluded.
**Expected:** the 👁️ disable applies to game world-gen too.

**Where it is**

`packages/server/src/routes/game.routes.ts:5594-5600` — the world-gen call:
```
const lorebookResult = await processLorebooks(app.db, [], null, {
  characterIds: setupConfig.partyCharacterIds,
  personaId: setupPersonaId,
  activeLorebookIds: setupConfig.activeLorebookIds,
  generationTriggers: ["game_setup", "game"],
  resolveContent: resolveSetupLorebookMacrosForFinal,
});
```
passes no `excludedLorebookIds`, so `filterRelevantLorebooks` (`packages/server/src/services/lorebook/index.ts:246`) receives an empty exclusion set and drops nothing. Contrast the roleplay path (`generate.routes.ts:1229` builds `lorebookScopeExclusions = resolveLorebookScopeExclusions(chatMode, chatMeta)` and passes `excludedLorebookIds` at every resolution call site, e.g. 1512).

**What's happening**

This is the one lorebook-resolution call site that was never wired to the per-chat exclusion set.

**Suggested fix**

Near the game setup config assembly, call `resolveLorebookScopeExclusions(chatMode, chatMeta)` (as `generate.routes.ts:1229` does) and add `excludedLorebookIds: lorebookScopeExclusions.excludedLorebookIds` to the `processLorebooks` options at `game.routes.ts:5594-5600`. Check for any other game-mode `processLorebooks` / `listActiveEntries` call sites with the same omission.

*Note: this is a distinct bug found while diagnosing a persona-lorebook report — it is NOT the cause of that report (that turned out to be a UX issue in roleplay, filed separately). This game-mode gap is real on its own.*

---
*AI-investigated against the current tree (v2.0.9); the world-gen call and the exclusion gate read and confirmed. Please verify before fixing.*

### LINKED ISSUE #3377: [Issue]: Persona/character card info is duplicated in the prompt when a description starts with a macro (conversation / preset-less modes)
**What's wrong**

Persona (and character) card info can be injected into the system prompt **twice** — once from your own prompt/macro, and once from an automatic "identity fallback" that fails to notice the info is already there. You get a duplicated description/personality block.

The report is right that this happens, but two of its details don't hold up in the code, so here's the corrected version:
- **It is not persona-specific.** Persona and character run the *identical* code path, so a character whose description starts with a macro duplicates the same way. There is no "character macros suppress, persona doesn't" asymmetry.
- **The persona section *marker* in a preset does not trigger it.** In roleplay/VN with a preset, the marker (preset assembler) path and the auto-inserter are mutually exclusive. The real trigger is inline text/field macros — in **conversation mode** (where the fallback always runs), or in a non-game mode with **no preset** selected.

The everyday trigger: personas very commonly start their description with `{{user}}` (or `{{char}}`), which is exactly what defeats the duplicate-detection.

1. Create a persona whose **Description starts with a macro**, e.g. `{{user}} is a rival journalist who …`.
2. Start a **conversation-mode** chat (or any non-game chat with **no** prompt preset selected).
3. Put the persona text macro (`{{persona}}`, or just `{{personaDescription}}`) into your system/prompt text.
4. Send a message and inspect the assembled prompt (Peek Prompt / debug).

**You'll see:** the persona description appears **twice** — once where your macro resolved it, and once in an auto-injected system block.
**Expected:** exactly once.
**Control that proves the cause:** change the description to plain text with no leading macro → it appears once. (The same duplication reproduces for a *character* whose description starts with a macro.)

**Where it is**

`packages/server/src/services/generation/character-prompt-context.ts` — `injectIdentityFallbackMessages` (133-217) re-injects card info unless `hasProfileBlock` (123-130) says it's already present. The character gate is line 151 and the persona gate line 188 — identical logic. `hasProfileBlock`'s only mode-independent check compares the **unresolved** description (`args.personaDescription`, macros intact) against the **resolved** prompt text:
```
content.includes(description.split("\n")[0].trim().slice(0, 80))   // line 126
```
so a macro in the description's first line makes the two strings differ and the check misses. The other checks — `<xmlTag>` (127), `<name>` (128), `# name` (129) — only match when the info was wrapped with the entity name, but inline text macros never emit a name wrapper and `wrapFormat="none"` emits none. Detection returns false and the block is spliced in a second time (persona 213-216, character 182-185).

This fallback runs (so the bug is reachable) only where `shouldInjectIdentityFallback` returns true (`packages/server/src/routes/generate/generate-route-utils.ts:1160-1169`: game → never, conversation → always, other modes → only when no preset). A parallel copy of the gate lives in `packages/server/src/routes/chats.routes.ts` (~2393-2415).

**What's happening**

The fallback decides "is this identity already in the prompt?" by fuzzy, after-the-fact text matching — comparing an *unresolved* description against *resolved* prompt text. When the description leads with a macro (or was surfaced without a name wrapper), the match fails and the card info is added again.

**Suggested fix**

Make detection macro-aware — but do it carefully so it doesn't reintroduce the macro-resolution cost problem from PR #3118 (see the note below). Two options, preferred first:

- **Preferred — deterministic "already presented" flag (zero extra macro resolution).** Instead of re-deriving presence by substring, track whether the identity was surfaced and skip the fallback when it was. This mirrors what the assembler already does for `chat_summary` (`assembler.ts:362 / 380-387 / 524`). In conversation mode (no marker pipeline) this can be a check for the persona/char macro token in the **pre-resolution** prompt text — still no macro resolution. This is the most robust and the least risky performance-wise.
- **Otherwise — resolve once and reuse.** If you resolve the description for the substring check, resolve it a **single time** with the resolver the block path already uses (`args.resolvePromptMacros(personaDescription)` for persona; the loop's `resolveCharacterMacros(character.description)` for characters — note the character check at line 151 must move to after `characterMacroContext` is built at 153-166) and reuse that same resolved string for **both** the `hasProfileBlock` detection **and** the injected block (persona line 192, character 170). Do **not** resolve twice, and do **not** touch `resolveMacros` in `packages/shared` or any client display path.

Apply the same fix to the parallel gate in `chats.routes.ts` (~2393-2415), and extend `scripts/regressions/prompt.regression.ts` (it already imports this fallback and guards the #3118 lazy-persona behavior) with: a macro-leading description → identity appears exactly once; a plain description → unchanged; and byte-identical macro output preserved.

**Performance & regression note (re: [PR #3118](https://github.com/Pasta-Devs/Marinara-Engine/pull/3118))**

#3118 fixed a super-linear freeze caused by `resolveMacros` eagerly rebuilding persona text **per message on every display render** (client-side, `ChatMessage`/`ConversationMessage`/`GameNarration`). This fix is **not** on that surface: `injectIdentityFallbackMessages` runs **server-side, once per generation** during prompt assembly (`generate.routes.ts:2177`), so at most it adds **one** description-resolve per identity per generation — negligible next to generation itself, and post-#3118 a single resolve is ~1 ms even for a pathological self-referential persona. The guardrails to keep it that way: (1) resolve at most once and reuse (or use the deterministic flag and resolve nothing extra); (2) use the resolver the block path already uses, so deferred-character-macro (`deferCharacterMacros`) and budget-cap semantics are unchanged and macro output stays byte-identical; (3) make no change to `packages/shared/src/utils/macro-engine.ts` or any per-render display path, so #3118's early-out and lazy-persona optimizations are untouched.

---
*AI-investigated against the current tree (v2.0.9); the fallback gate, `hasProfileBlock`, `shouldInjectIdentityFallback`, the persona/character symmetry, the server-once-per-generation call site, and the existing regression harness all read and confirmed. The original report's two inaccuracies (persona-only; marker-triggered) are corrected above. Please verify before fixing.*

### LINKED ISSUE #3378: [Issue]: Command list is wrapped in a literal <commands> XML tag even when section wrapping is set to None
**What's wrong**

When "prompt section wrapping" is set to **None**, every context block renders without XML tags — except the character-command list, which is still hard-wrapped in a literal `<commands>` … `</commands>` element. So with None you get bare, unwrapped status/memory/context blocks plus one incongruously XML-tagged command block.

1. In a conversation chat, enable Character Commands (at least one command available).
2. Set prompt section wrapping to **None**.
3. Generate and inspect the assembled prompt (Peek Prompt / debug).

**You'll see:** the command reminder is still wrapped in `<commands>` … `</commands>`, while the status/context and memory blocks render unwrapped.
**Expected:** with None selected, the command list renders without the XML tags, like every other block.

**Where it is**

`packages/server/src/services/generation/conversation-command-runtime.ts` — the command-reminder builder hardcodes the tags: `commandLines` starts with a literal `` `<commands>` `` (line 204), and the matching `</commands>` is appended near line 321. The builder never receives `wrapFormat` and never calls `wrapContent`, so it ignores the None setting that every sibling block honors.

**What's happening**

Unlike the status/context and memory blocks — which go through `wrapContent` and therefore respect `wrapFormat="none"` — the command list literally string-concatenates the `<commands>` open/close tags.

**Suggested fix**

Thread `wrapFormat` into the command-reminder builder and wrap the body via `wrapContent(body, "commands", wrapFormat)` instead of hardcoding the tags, so None strips them and xml/markdown stay consistent with the rest of the prompt. (Reporter flagged this in passing while requesting relocate/suppress macros — filed as its own bug; see the related enhancement.)

---
*AI-investigated against the current tree (v2.0.9); the hardcoded opening tag at line 204 read and confirmed. Please verify before fixing.*

### LINKED ISSUE #3379: [Feature]: Add relocate/suppress macros for auto-inserted conversation contexts (memories, status/time, emoji-react, command list)
**What's being requested**

Persona and character info can be placed/suppressed with macros — use the macro and the automatic insertion is skipped (no duplication). Several other auto-inserted **conversation-mode** context blocks have **no** such macro, so you can't relocate them or turn them off — and with section wrapping set to **None** they dump into the raw system prompt with nothing separating them. Add macros with the same "using the macro suppresses the auto-insert" contract for:

- **Conversation memories** — "Things you must remember from past conversations: [Week of 08.06.2026 – 14.06.2026] …"
- **Chat status / context** — "Your current status: online and active."; "{{user}}'s status: active."; "The current time and date: 10:24, 08.07.2026."
- **Emoji-react rules**
- **Command list** (this one also has a standalone bug where it's wrapped in `<commands>` even when wrapping is None — filed separately)

(Lorebook already has a marker for preset mode; it's only bypassed in conversation mode, so a conversation-mode lorebook macro would round this out.)

**Why**

When section wrapping is None, a lot of the automatic conversation context is dumped into the raw system prompt with no wrapping and blends together. Blocks that have macros can be controlled (placed/suppressed); these blocks don't have macros, so they can't be — you can neither separate them, move them, nor remove them.

**Where it is**

These blocks are built **outside** the preset/marker pipeline, so they can't reuse the persona/character suppression directly:
- memories — `packages/server/src/services/generation/conversation-history-runtime.ts`
- status/time context — `packages/server/src/services/generation/conversation-context-block.ts`
- emoji-react rules — `packages/server/src/routes/generate.routes.ts` (~1753-1762)
- command list — `packages/server/src/services/generation/conversation-command-runtime.ts`

The marker registry (for reference on how existing markers like persona/lorebook are defined) is `packages/server/src/services/prompt/marker-expander.ts:113-132`; lorebook's marker is at `marker-expander.ts:311-376`.

**Suggested approach**

Add optional conversation-mode macros (e.g. `{{memories}}`, `{{context}}`/`{{status}}`, `{{commands}}`, `{{reactRules}}`) that, when present in the prompt text, expand in place and **suppress** the corresponding auto-insert. Since conversation mode has no marker-expansion pipeline, implement suppression by scanning the resolved prompt for the token (a deterministic "present" flag) and skipping the auto-append — the same deterministic-flag approach the assembler already uses for `chat_summary` (`assembler.ts:362/380-387/524`), and the same fix direction recommended for the persona/character duplication bug. Pair the `{{commands}}` macro with the "command list ignores wrapFormat=none" fix.

---
*AI-investigated against the current tree (v2.0.9). Builder locations identified from a code sweep — treat the exact line numbers as pointers to confirm. Filed as an enhancement per the reporter's framing.*

### LINKED ISSUE #3380: [Feature]: Way to exclude a persona/character field from auto-insertion without rendering it in the prompt
**What's being requested**

A way to tell the engine "do **not** auto-add this persona/character field" **without** having to render it in the system prompt. Today, the only way to stop a field from being auto-inserted is to put its macro in the prompt so it appears in the final output — there's no way to say "don't auto-add it *and* don't show it," and no way to exclude a single field while letting the rest auto-add.

**The user's attempted workaround (and why it fails)**

They tried referencing the fields inside an always-false conditional, hoping the "using the macro suppresses the auto-insert" behavior would kick in without rendering anything:
```
{{#if user contains "DONTINCLUDETHIS"}}
{{description}}

{{personality}}

{{backstory}}

{{appearance}}

{{scenario}}
{{/if}}
```
It doesn't work: whether the auto-insert is suppressed is decided by whether the fields appear in the **final, resolved** prompt — not by whether the macros are present in the template. The `{{#if}}` is false, so the fields never reach the output, so the detector treats them as "missing" and auto-adds the whole block anyway.

**Why (root cause — shared with the duplication bug #3377)**

The identity auto-insert (`injectIdentityFallbackMessages`) decides whether to inject by scanning the **resolved** prompt text for the info — `hasProfileBlock` (`packages/server/src/services/generation/character-prompt-context.ts:123-130`) over `allContent = args.messages.map(m => m.content).join("\n")`. Because it's output-based, it cannot distinguish "the author deliberately excluded this" from "the author forgot to include it." It's also **entity-level** (all-or-nothing per persona/character), so there's no per-field control.

**Suggested approach**

Express suppression from author intent in the **template**, not from the resolved output — the same template-based / deterministic-flag direction already recommended for the #3377 duplication fix:
- If the author references a field's macro anywhere in the template (including an unreachable `{{#if}}` branch), treat that field as "handled" → suppress its auto-add whether or not it renders. This makes the user's false-conditional trick work and delivers exclude-without-render.
- For finer control, make the identity auto-insert **field-granular** (respect per-field references) so one field (e.g. `{{backstory}}`) can be excluded while the rest still auto-add.
- Alternatively/additionally, a dedicated explicit opt-out would be clearer and more discoverable — e.g. an exclude token like `{{exclude:backstory}}` / `{{persona:none}}`, or per-field "auto-include" toggles in the persona/character editor.

Note: implementing #3377's fix via template-based detection would grant the **entity-level** version of this for free; the extra ask here is **field-level** granularity and/or an explicit opt-out affordance. Related: #3377 (identity duplication — same detection root cause), #3379 (macros for other auto-inserted conversation contexts).

---
*AI-investigated against the current tree (v2.0.9); the output-based detection (`hasProfileBlock` scans resolved messages) confirmed, which is why an always-false `{{#if}}` can't suppress the auto-add. Filed as an enhancement per the reporter's request.*

### LINKED ISSUE #3381: [Feature]: Expose/export generated TTS clips — already persisted (IndexedDB) but unreachable
**What's being requested**

Marinara already saves every generated TTS clip to disk (a persistent IndexedDB cache), but there's no UI to retrieve, replay, or export them — unlike generated **images**, which are both persisted *and* surfaced in a gallery. The request is a small export/download affordance (or an audio gallery) that reads the existing cache. No new persistence layer is needed — the storage is already there; only the surface is missing.

**What's already in place (verified against source)**

- `packages/client/src/lib/tts-audio-cache.ts` maintains a **persistent IndexedDB cache**: DB `marinara-tts-audio-cache`, store `voiceLines` (+ a `voiceLineMeta` store), capped at **750 clips / 100 MB** (lines 5-11), with LRU pruning (114-149). Every generated clip is written via `putPersistentBlob` (170-198) inside `getOrCreateCachedTTSAudioBlob` (212-265).
- Both playback surfaces populate it — Conversation (`ChatArea.tsx`) and Game (`GameNarration.tsx`) — plus `ChatMessage.tsx`, `GameCombatUI.tsx`, and `tts-dialogue.ts`.
- The server route `packages/server/src/routes/tts.routes.ts` is a pure streaming proxy — it stores only the TTS config / API key (`parseStoredConfig`), **not** audio. So persistence is entirely the client-side cache.
- The cache exposes only **key-based getters** (`getCachedTTSAudioBlob`) — there is no enumeration/list/export API, and no consumer other than playback. So the clips sit on disk but are unreachable by the user.

**Why**

Generated media should be retrievable. Images already are (persisted + gallery). TTS clips are persisted the same way but have no surface — the gap is the missing UI, not missing storage.

**Suggested approach**

Add a list/enumeration helper to `tts-audio-cache.ts` (iterate the `voiceLines` / `voiceLineMeta` store) and a small UI that plays and exports from the existing cache — a download/export button on a played clip, and/or an audio gallery mirroring the image gallery. Optionally associate clips with their message/speaker for labeling. No changes to how clips are generated or stored are required.

---
*AI-investigated against the current tree; the IndexedDB cache, its write sites, the key-only getters, and the proxy-only server route were all read and confirmed. Filed as an enhancement per the reporter's framing.*

### LINKED ISSUE #3382: [Issue]: TTS clip is dropped (not played) when the tab is unfocused as generation finishes
**What's wrong**

If you trigger TTS and switch away from the tab before the audio finishes generating, the clip is dropped instead of played. TTS is fetched asynchronously and `Audio.play()` is called only *after* the fetch resolves — so if the tab is hidden at that moment, the browser's autoplay policy rejects `play()` and the clip is discarded (an error state is set, but nothing is played or retried).

1. Trigger TTS on a message (or let auto-TTS fire) with a voice that takes a moment to generate (e.g. ElevenLabs).
2. Immediately switch to another browser tab so Marinara's tab is hidden while the audio is still generating.
3. Switch back after it finishes.

**You'll see:** the clip never played (state went to an error / "Browser blocked audio playback"), and it isn't retried when you return.
**Expected:** the clip plays when you return to the tab — it should be queued and played on the next focus/visibility change instead of being dropped.

**Where it is**

`packages/client/src/lib/tts-service.ts` — the audio element is created *after* the async fetch: `blob = await this.getAudioBlob(...)` (192), then `const audio = new Audio(objectUrl)` (218), then `await audio.play()` (235). The catch (236-243) handles a rejected `play()` by calling `this.cleanup()` and `setState("error")` with "Browser blocked audio playback" — there's no queue or retry. A search for `visibilitychange` / `document.hidden` in this file finds nothing, so a hidden tab at play-time is unhandled.

**What's happening**

Browser autoplay policy blocks `play()` for audible media when the document is hidden and there's no fresh user gesture by the time the async fetch resolves. Because the play call happens post-fetch, a tab switch during generation lands exactly in that blocked window, and the code treats the rejection as a terminal error rather than a deferrable one.

**Suggested fix**

When `audio.play()` rejects (or `document.hidden` is true at play-time), don't discard the clip — hold it and retry playback on the next `visibilitychange`/focus, guarded so it only plays if it's still the current sequence. Revoke the object URL if the user has moved on. This keeps the existing "generate-then-play" flow but survives a backgrounded tab.

---
*AI-investigated against the current tree; the post-fetch `new Audio().play()` path, the terminal-error catch, and the absence of any visibility handling were all read and confirmed. Please verify before fixing.*

### LINKED ISSUE #3383: [Issue]: Character tab "Load more" bar overlaps character cards and blocks tapping them (mobile)
**What's wrong**

In the Characters tab, the "Load more" button is a bar pinned to the bottom of the list (it "stays still on one spot"). It sits **on top of** the character grid and swallows taps, so a character card that ends up underneath it can't be selected — most noticeable on a narrow/mobile viewport, where the pinned bar covers the bottom strip of the list right where you'd tap the lower characters.

1. Have enough characters that the Characters tab shows the "Load more" button (i.e. more than one page).
2. On a phone / narrow viewport (reported on the Android APK), scroll the character list so a character card sits behind the pinned "Load more" bar at the bottom.
3. Tap that character.

**You'll see:** the tap doesn't select the character — it hits the "Load more" bar instead; the bottom card is effectively unclickable.
**Expected:** every character card is tappable; the "Load more" bar shouldn't cover or intercept taps meant for a card.

**Where it is**

`packages/client/src/components/panels/CharactersPanel.tsx` — the "Load more" container at lines 1397-1408 is `sticky bottom-0 z-20` with an opaque background (`bg-[var(--sidebar)]/95` + `backdrop-blur-md`):
```
<div className="sticky bottom-0 z-20 -mx-3 mt-2 border-t ... bg-[var(--sidebar)]/95 px-3 pt-2 pb-3 backdrop-blur-md">
  <button ... onClick={() => void characterPages.fetchNextPage()}> Load more (...) </button>
</div>
```
The scroll container is the panel root (line 683, `... overflow-y-auto p-3`), and the sticky bar is a sibling **after** the card grid (grid closes at 1395). The card grid has **no bottom padding** reserving space for the bar, so cards can rest under it.

**What's happening**

A `position: sticky` element sits above the scrolling list in the stacking context (here `z-20`) and captures pointer events across its entire box — an opaque/semi-transparent background does not make it click-through (only `pointer-events: none` would). Because the grid reserves no space beneath the bar, whichever card is in the bar's footprint is both visually covered and untappable. On a short mobile viewport the bar occupies a proportionally large bottom strip, so the lower character(s) are routinely blocked.

**Suggested fix**

Reserve space so no card ever sits under the bar — e.g. add bottom padding to the scrollable card grid roughly equal to the bar's height (a `pb-*` on the grid, or a spacer), so the last row can scroll clear of the bar and the bottom-most character is always tappable. Alternatively, drop the overlap on mobile (render "Load more" inline at the end of the list rather than pinned), or shrink the pinned bar's footprint. Verify on a real phone / narrow viewport that the last character in the list can be tapped.

---
*AI-investigated against the current tree; the sticky bar styling, the scroll container, and the absence of reserved space below the grid were read and confirmed. The exact tap-blocking was reasoned from the layout (not reproduced on a device) — please verify on a phone/narrow viewport before fixing.*

### LINKED ISSUE #3384: [Issue]: Lorebook Keeper "Proposed Text" omits the proposed newFacts for updates to existing entries
*Reported against v2.1.0; confirmed still present on v2.1.1 (verified against source).*

**What's wrong**

When Lorebook Keeper proposes an **update** to an existing lorebook entry, the "Proposed Text" in the review modal shows the entry's existing/echoed content and leaves **out** the actual facts being proposed (`newFacts`). So you review roughly what the entry already says, not what's being added. **Brand-new entries preview correctly** — it's only updates to existing entries where the newFacts go missing from the preview. (You can confirm the newFacts exist by checking the agent's response in the console: `"newFacts": [...]` — those items aren't in the Proposed Text.)

1. In a chat with Lorebook Keeper active and "require approval for agent writes" turned on, get it to propose an **update** to an existing entry (e.g. a new durable fact about a character who already has an entry).
2. Open the "Review Lorebook Update" modal and read the "Proposed Text".
3. Compare with the agent's response in the browser console / network.

**You'll see:** the Proposed Text shows the existing/echoed entry content; the proposed `newFacts` (visible in the console response) are absent.
**Expected:** the Proposed Text shows what will actually be committed — the existing entry **plus** the new facts.

**Where it is**

`packages/server/src/routes/generate/agent-write-approval.ts` — the preview is built by `formatLorebookWriteApprovalText` (61-77) → `readUpdateContent` (24-33), which returns `update.content` (or `update.entry.content`) if present and only falls back to formatting `newFacts` when there is **no** content:
```
if (typeof update.content === "string" && update.content.trim()) return update.content.trim();   // returns here
...
if (Array.isArray(update.newFacts)) { ... }   // never reached when content is present
```
The Lorebook Keeper is instructed (`packages/shared/src/constants/agent-prompts.ts` ~165-174) to return **only** `newFacts` for updates and use `content` only for empty/malformed entries — but the model frequently returns a `content` too (echoing/rewriting the current entry). When it does, the preview shows that content and drops the `newFacts`.

By contrast, the **commit** path merges: `persistLorebookKeeperUpdates` (`lorebook-keeper-utils.ts:409-416`) uses `mergeLorebookKeeperUpdateContent` (271-305) with existing + replacement + newFacts. So the preview is built by a **different, incomplete** path than the commit — it doesn't reflect what will be saved. The proposal is assembled from the raw agent updates (`generate.routes.ts:3097, 3135`) with no existing-entry lookup.

**⚠️ Also worth verifying — possible newFacts loss at commit (not just preview):** when an update has `action:"update"` **and** a non-empty `content`, the commit takes the `getExplicitUpdateReplacementContent` branch (`lorebook-keeper-utils.ts:409-411` → `307-315`), which treats that content as a **wholesale replacement** and also drops `newFacts`. If the model's echoed `content` doesn't already contain the new facts, they may not just be hidden from the preview — they may not be saved at all. Please check whether newFacts survive to the committed entry in this case.

**What's happening**

The approval preview and the commit use different content logic. The preview is content-first and treats `newFacts` as a fallback, so any update carrying a `content` hides its `newFacts`. New entries put their real text in `content`, so they preview fine; updates (whose delta lives in `newFacts`) don't.

**Suggested fix**

Build the preview from the **same merge the commit uses** — resolve the existing entry and show `mergeLorebookKeeperUpdateContent(existing, replacement, newFacts)`, so "Proposed Text" is exactly what will be committed (existing + new facts). This means the proposal builder (`buildLorebookWriteApprovalProposal` / its call site at `generate.routes.ts:3135`) needs the existing entry content, which it doesn't currently fetch. Also reconsider whether an update carrying `content` should drop `newFacts` at commit at all (the instructions reserve `content` for empty/malformed entries).

**Related enhancement (also requested):** once the newFacts are shown, highlight/diff them against the original so it's clear what's **new** vs. original text (render the appended facts distinctly, or a small diff view). Even when displayed, it's hard to tell new from existing.

---
*AI-investigated against v2.1.1 (`git describe` = v2.1.1); the preview builder, the keeper output schema, and the commit merge/replace paths were all read and confirmed. The preview-drops-newFacts behavior is verified from the code; the commit-loss angle depends on the model including `content` — please verify on a real proposal before fixing.*

### LINKED ISSUE #3385: [Issue]: Branching a roleplay/VN chat clears all tracker state (world + character tracker) — snapshot copy is gated to Game mode only
**What's wrong**

When you branch a chat, the world/character tracker state is wiped in the new branch — even though the tracker agents are still on and you're continuing from the same point in the same conversation. You end up having to recreate everything by hand: the world tracker, and especially the character tracker (including the per-character variable fields, which aren't easy to re-add).

The cause: branching copies the tracker snapshots to the new branch **only for Game-mode chats**. Roleplay and Visual Novel chats use the exact same tracker snapshots, but the copy step skips them, so the branch starts with empty tracker state. This is a bug, not intended behavior — a branch continues the same conversation from the same point, so the tracker state should carry over just like it does in Game mode.

1. In a **roleplay** (or visual novel) chat with the world tracker and character tracker agents on, play until the trackers hold real state (location/time/weather, present characters, character variables, etc.).
2. Branch the chat ("fork from here").
3. Open the new branch and look at the trackers.

**You'll see:** the tracker panels are empty / reset — all world and character tracker info (including custom character variables) is gone, even though the agents are still enabled and the messages carried over.
**Expected:** the branch keeps the tracker state up to the branch point, the same way Game-mode branches do.

**Where it is**

`packages/server/src/routes/chats.routes.ts` — the branch route copies game-state/tracker snapshots to the new branch only when the source chat is Game mode:
```
if (sourceToBranchedMessageId.size > 0 && sourceChat.mode === "game") {   // line 3167
  ... copy gameStateSnapshots (date/time/location/weather/presentCharacters/
      playerStats/personaStats/fieldLocks/manualOverrides) ...
}
```
These `gameStateSnapshots` are keyed by `(chatId, messageId, swipeIndex)` and are mode-agnostic — the same snapshots roleplay/VN trackers write and that the generation path reads back as "committed tracker context" (`packages/server/src/routes/generate.routes.ts` ~2519 `getCommittedForMessages`, ~3068 `injectCommittedTrackerContext`). The branch route's own comment calls them "the correct tracker state for that specific message." So the `mode === "game"` condition is an over-restriction that drops roleplay/VN tracker state. (Messages and their swipes ARE copied for every mode — only the tracker snapshots are gated.)

**What's happening**

The tracker-snapshot copy on branch was scoped to Game mode, but trackers exist in roleplay and visual-novel modes too (same snapshot table). So a roleplay/VN branch copies the messages but none of the tracker snapshots, and the new branch reads back empty tracker state. The per-character variable fields the reporter mentions are the snapshot's `manualOverrides` (copied at `chats.routes.ts` ~3180/3198 inside the same gated block), so they're lost too.

**Suggested fix**

Copy the `gameStateSnapshots` on branch regardless of mode — change the guard from `sourceChat.mode === "game"` to just `sourceToBranchedMessageId.size > 0` (the copy is keyed per message and no-ops for a chat that has no snapshots). If desired, keep the turn-game engine-state (`gameEngineState`, UNO) copy inside a game check, but the tracker snapshots should be copied for roleplay and visual_novel too. This restores the world tracker, character tracker, and the per-character variables (`manualOverrides`) in the branch.

*On the reporter's "did this work before?" — I can't speak to history without git blame, but the current v2.1.1 code definitively gates tracker-snapshot copying to Game mode, so roleplay/VN branches lose tracker state today. It's a bug, not a feature request.*

---
*AI-investigated against v2.1.1 (`git describe` = v2.1.1); the branch mode gate, the snapshot fields, and the roleplay generation read/commit of these snapshots were all read and confirmed. Please verify before fixing.*

SpicyMarinara:
## Why

This PR batches the v2.1.2 issue sweep and the earlier requested fixes for agent output formatting, Grok 4.5, streaming stability, and recovery-page theming. It resolves the current open issue queue without targeting main directly.

Closes #3375, closes #3376, closes #3377, closes #3378, closes #3379, closes #3380, closes #3381, closes #3382, closes #3383, closes #3384, closes #3385.

## What Changed

- Fixes prompt assembly for active-agent output formats, Conversation context relocation macros, command wrapping, identity fallback suppression, and Lorebook Keeper approval/commit merging.
- Copies tracker snapshots when branching Roleplay/VN chats and applies game lorebook exclusions during world setup.
- Adds Grok 4.5/xAI reasoning handling and xAI connection defaults.
- Stabilizes roleplay streaming when parallel agents finish after the main stream.
- Fixes TTS hidden-tab playback, adds cached TTS clip export, and updates mobile/lorebook settings controls.
- Updates v2.1.2 changelog/release metadata and recovery-page chroma text color handling.

## Validation

- pnpm check
- pnpm version:check
- pnpm credits:check
- pnpm regression:prompt
- pnpm smoke:ui
- git diff --check

## Manual Verification Needed

- Manually verify Conversation prompt macros (`{{context}}`, `{{commands}}`, `{{reactRules}}`, `{{memories}}`, `{{lorebook}}`) in a browser chat preset.
- Manually verify hidden-tab TTS autoplay behavior in the target browser.
- Manually verify the lorebook active/disabled controls with a book that is both pinned and auto-active.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 
 ## [2.1.2]
 
+### Added
+
+- Added Grok 4.5 to the xAI / Grok model list and made new xAI connections default to it.
+- Added Conversation prompt relocation macros for auto-inserted context: `{{context}}`/`{{status}}`, `{{commands}}`, `{{reactRules}}`, `{{memories}}`, and `{{lorebook}}`.
+- Added a TTS cache export control in Text to Speech settings so generated cached voice clips can be downloaded from IndexedDB.
+
 ### Changed
 
 - Bumped release metadata to v2.1.2 across packages, the PWA manifest, README release pointer, Windows installer sources, Android APK metadata, and the home-page-visible app version.
@@ -16,6 +22,16 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 - Fixed manual agent retry/rerun requests so only agents currently added to the chat can be resolved, preventing removed agents from being prompted by stale retry requests.
 - Fixed agent prompt assembly so the required output format is appended to the terminal user message after chat history using the selected chat prompt preset wrapper (`<output_format>`, `## Output Format`, or raw text).
 - Fixed non-Quest agents receiving active quest progress in current game-state context while keeping compact quest state available to the Quest Tracker when it is active.
+- Fixed Roleplay/VN chat branching so tracker snapshots copy to the branch instead of only copying Game Mode snapshots (#3385).
+- Fixed Lorebook Keeper approval previews and commits so updates to existing entries keep the existing text and include `newFacts` alongside proposed replacement content (#3384).
+- Fixed the mobile Characters panel **Load more** footer so it no longer overlaps or blocks the last character cards (#3383).
+- Fixed generated TTS playback so clips created while the tab is hidden or unfocused wait for the tab to return instead of being dropped (#3382).
+- Fixed prompt identity fallback so card fields referenced by macros are not duplicated, and fields can be intentionally suppressed by placing their macro in the prompt template (#3380, #3377).
+- Fixed Game Mode world setup lorebook generation so per-chat disabled lorebooks and hidden Game Lorebook Keeper books are excluded (#3376).
+- Fixed active lorebook controls so a pinned lorebook that is also active via persona, character, or global scope can still be disabled in the current chat (#3375).
+- Fixed Conversation command reminders so preset wrap format `None` no longer emits a literal `<commands>` XML wrapper (#3378).
+- Fixed roleplay streaming recovery so parallel agent events are deferred until the main assistant stream finishes, preventing late agent updates from replacing the streamed message.
+- Fixed the recovery/error page so **Internal Server Error** uses the configured chat chrome text color instead of a hard-coded pink.
 
 ## [2.1.1]
 
--- a/packages/client/src/App.tsx
+++ b/packages/client/src/App.tsx
@@ -1,7 +1,7 @@
 // ──────────────────────────────────────────────
 // App: Root component with layout
 // ──────────────────────────────────────────────
-import { Component, lazy, Suspense, useEffect, useMemo, type ErrorInfo, type ReactNode } from "react";
+import { Component, lazy, Suspense, useEffect, useMemo, type CSSProperties, type ErrorInfo, type ReactNode } from "react";
 import { useQuery } from "@tanstack/react-query";
 import { APP_VERSION } from "@marinara-engine/shared";
 import { CustomThemeInjector } from "./components/layout/CustomThemeInjector";
@@ -91,6 +91,16 @@ function formatRecoveryError(error: unknown) {
   }
 }
 
+function getRecoveryChromeTextStyle(): CSSProperties | undefined {
+  const { chatChromeTextColor, theme } = useUIStore.getState();
+  const textColor = chatChromeTextColor.trim();
+  if (!textColor) return undefined;
+
+  return {
+    "--marinara-chat-chrome-text": getCssColorFallback(textColor, getDefaultChatChromeTextColor(theme)),
+  } as CSSProperties;
+}
+
 export class AppRecoveryBoundary extends Component<{ children: ReactNode }, { error: unknown; hasError: boolean }> {
   state: { error: unknown; hasError: boolean } = { error: null, hasError: false };
 
@@ -117,9 +127,13 @@ export class AppRecoveryBoundary extends Component<{ children: ReactNode }, { er
   render() {
     if (!this.state.hasError) return this.props.children;
     const errorMessage = formatRecoveryError(this.state.error);
+    const recoveryChromeTextStyle = getRecoveryChromeTextStyle();
 
     return (
-      <div className="mari-chrome-token-scope flex min-h-screen items-center justify-center bg-[var(--background)] px-4 text-[var(--marinara-chat-chrome-panel-text)]">
+      <div
+        className="mari-chrome-token-scope flex min-h-screen items-center justify-center bg-[var(--background)] px-4 text-[var(--marinara-chat-chrome-panel-text)]"
+        style={recoveryChromeTextStyle}
+      >
         <div className="w-full max-w-lg rounded-xl border border-[var(--marinara-chat-chrome-accent)] bg-[var(--marinara-chat-chrome-panel-bg)] p-5 shadow-2xl ring-1 ring-[var(--marinara-chat-chrome-focus-ring)]">
           <h1 className="text-lg font-semibold text-[var(--marinara-chat-chrome-panel-title)]">
             Marinara hit a recoverable UI error.
--- a/packages/client/src/components/connections/ConnectionEditor.tsx
+++ b/packages/client/src/components/connections/ConnectionEditor.tsx
@@ -1253,7 +1253,7 @@ export function ConnectionEditor() {
                     // `grok models`. Other providers keep their usual seeded
                     // default model when we know one.
                     setLocalModel(
-                      key === "grok_subscription" ? "" : (defaultModel?.id ?? (key === "xai" ? "grok-4.3" : "")),
+                      key === "grok_subscription" ? "" : (defaultModel?.id ?? (key === "xai" ? "grok-4.5" : "")),
                     );
                     setLocalMaxContext(
                       key === "grok_subscription"
--- a/packages/client/src/components/modals/CreateConnectionModal.tsx
+++ b/packages/client/src/components/modals/CreateConnectionModal.tsx
@@ -96,7 +96,7 @@ export function CreateConnectionModal({ open, onClose }: Props) {
           </div>
           <p className="text-[0.625rem] text-[var(--muted-foreground)]">
             {provider === "xai"
-              ? "Creates an xAI connection prefilled with Grok 4.3 and https://api.x.ai/v1."
+              ? "Creates an xAI connection prefilled with Grok 4.5 and https://api.x.ai/v1."
               : "You can adjust the endpoint, key, and model after creating the connection."}
           </p>
         </div>
--- a/packages/client/src/components/panels/CharactersPanel.tsx
+++ b/packages/client/src/components/panels/CharactersPanel.tsx
@@ -309,14 +309,7 @@ export function CharactersPanel() {
         toast.error("Failed to remove tag from some characters");
       }
     },
-    [
-      sort,
-      updateCharacter,
-      includedTags,
-      excludedTags,
-      setCharacterPanelIncludedTags,
-      setCharacterPanelExcludedTags,
-    ],
+    [sort, updateCharacter, includedTags, excludedTags, setCharacterPanelIncludedTags, setCharacterPanelExcludedTags],
   );
 
   const toggleIncludedTag = useCallback(
@@ -766,7 +759,11 @@ export function CharactersPanel() {
             New Folder
           </button>
         </div>
-        {parsedGroups.length > 0 && <p className="mari-folder-helper">Drag and drop characters to folders, double-click or double-tap to rename</p>}
+        {parsedGroups.length > 0 && (
+          <p className="mari-folder-helper">
+            Drag and drop characters to folders, double-click or double-tap to rename
+          </p>
+        )}
       </div>
 
       {/* Filters */}
@@ -1395,7 +1392,7 @@ export function CharactersPanel() {
       </div>
 
       {characterPages.hasNextPage && (
-        <div className="sticky bottom-0 z-20 -mx-3 mt-2 border-t border-[var(--marinara-chat-chrome-panel-divider)] bg-[var(--sidebar)]/95 px-3 pt-2 pb-3 backdrop-blur-md">
+        <div className="relative -mx-3 mt-2 border-t border-[var(--marinara-chat-chrome-panel-divider)] bg-[var(--sidebar)]/95 px-3 pt-2 pb-3 backdrop-blur-md md:sticky md:bottom-0 md:z-20">
           <button
             type="button"
             onClick={() => void characterPages.fetchNextPage()}
--- a/packages/client/src/components/panels/settings/TTSConfigCard.tsx
+++ b/packages/client/src/components/panels/settings/TTSConfigCard.tsx
@@ -15,12 +15,18 @@ import {
   ChevronUp,
   Plus,
   X,
+  Download,
 } from "lucide-react";
 import { cn } from "../../../lib/utils";
 import { toast } from "sonner";
 import { useTTSConfig, useUpdateTTSConfig, useTTSVoices } from "../../../hooks/use-tts";
 import { useCharacters } from "../../../hooks/use-characters";
 import { ttsService } from "../../../lib/tts-service";
+import {
+  listCachedTTSAudioEntries,
+  listCachedTTSAudioMeta,
+  type CachedTTSAudioExportEntry,
+} from "../../../lib/tts-audio-cache";
 import { parseCharacterDisplayData } from "../../../lib/character-display";
 import type {
   TTSConfig,
@@ -135,6 +141,49 @@ function addSavedVoiceOption(options: VoiceOption[], voiceId: string): VoiceOpti
 function formatVoiceOptionLabel(option: VoiceOption): string {
   if (option.category === "saved") return `${option.id} (saved; not in current voice list)`;
   return option.name === option.id ? option.id : `${option.name} (${option.id})`;
+}
+
+function formatCacheBytes(bytes: number): string {
+  if (!Number.isFinite(bytes) || bytes <= 0) return "0 B";
+  const units = ["B", "KB", "MB", "GB"];
+  let value = bytes;
+  let unitIndex = 0;
+  while (value >= 1024 && unitIndex < units.length - 1) {
+    value /= 1024;
+    unitIndex += 1;
+  }
+  return `${value >= 10 || unitIndex === 0 ? value.toFixed(0) : value.toFixed(1)} ${units[unitIndex]}`;
+}
+
+function extensionForTTSBlob(blob: Blob): string {
+  const type = blob.type.toLowerCase();
+  if (type.includes("mpeg") || type.includes("mp3")) return "mp3";
+  if (type.includes("wav")) return "wav";
+  if (type.includes("ogg")) return "ogg";
+  if (type.includes("webm")) return "webm";
+  if (type.includes("mp4") || type.includes("m4a")) return "m4a";
+  return "audio";
+}
+
+function safeTTSFileStem(value: string): string {
+  return (
+    value
+      .replace(/[^a-z0-9._-]+/gi, "-")
+      .replace(/-+/g, "-")
+      .replace(/^-|-$/g, "")
+      .slice(0, 80) || "tts-clip"
+  );
+}
+
+function downloadTTSClip(entry: CachedTTSAudioExportEntry, index: number): void {
+  const url = URL.createObjectURL(entry.blob);
+  const anchor = document.createElement("a");
+  anchor.href = url;
+  anchor.download = `${String(index + 1).padStart(3, "0")}-${safeTTSFileStem(entry.key)}.${extensionForTTSBlob(entry.blob)}`;
+  document.body.appendChild(anchor);
+  anchor.click();
+  anchor.remove();
+  setTimeout(() => URL.revokeObjectURL(url), 30_000);
 }
 
 const ELEVENLABS_DEFAULT_MALE_VOICE_NAMES = new Set([
@@ -364,6 +413,8 @@ export function TTSConfigCard() {
   const statusTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const [ttsState, setTTSState] = useState(ttsService.getState());
   const [previewError, setPreviewError] = useState<string | null>(null);
+  const [ttsCacheSummary, setTtsCacheSummary] = useState({ count: 0, bytes: 0 });
+  const [exportingTtsCache, setExportingTtsCache] = useState(false);
 
   // Voice fetch — keyed on the *saved* baseUrl so it only refetches when saved
   const savedSource = savedConfig?.source ?? "openai";
@@ -432,6 +483,21 @@ export function TTSConfigCard() {
     },
     [],
   );
+
+  useEffect(() => {
+    if (!expanded) return;
+    let cancelled = false;
+    void listCachedTTSAudioMeta().then((entries) => {
+      if (cancelled) return;
+      setTtsCacheSummary({
+        count: entries.length,
+        bytes: entries.reduce((total, entry) => total + Math.max(0, entry.size || 0), 0),
+      });
+    });
+    return () => {
+      cancelled = true;
+    };
+  }, [expanded, ttsState]);
 
   const buildPayload = (overrides?: Partial<TTSConfig>): TTSConfig => ({
     enabled,
@@ -571,6 +637,29 @@ export function TTSConfigCard() {
     })();
   };
 
+  const handleExportCachedClips = async () => {
+    setExportingTtsCache(true);
+    try {
+      const entries = await listCachedTTSAudioEntries();
+      if (entries.length === 0) {
+        toast.info("No cached TTS clips to export yet.");
+        setTtsCacheSummary({ count: 0, bytes: 0 });
+        return;
+      }
+
+      entries.forEach((entry, index) => downloadTTSClip(entry, index));
+      setTtsCacheSummary({
+        count: entries.length,
+        bytes: entries.reduce((total, entry) => total + Math.max(0, entry.size || entry.blob.size), 0),
+      });
+      toast.success(`Exported ${entries.length} cached TTS clip${entries.length === 1 ? "" : "s"}.`);
+    } catch {
+      toast.error("Failed to export cached TTS clips.");
+    } finally {
+      setExportingTtsCache(false);
+    }
+  };
+
   const voices = voicesData?.voices ?? [];
   const fetchedVoiceOptions = voicesData?.voiceOptions ?? voices.map((v) => ({ id: v, name: v }));
   const voiceOptions = useMemo(() => {
@@ -588,7 +677,15 @@ export function TTSConfigCard() {
       nextOptions = addSavedVoiceOption(nextOptions, savedVoice);
     }
     return nextOptions;
-  }, [fetchedVoiceOptions, narratorVoice, npcDefaultFemaleVoices, npcDefaultMaleVoices, source, voice, voiceAssignments]);
+  }, [
+    fetchedVoiceOptions,
+    narratorVoice,
+    npcDefaultFemaleVoices,
+    npcDefaultMaleVoices,
+    source,
+    voice,
+    voiceAssignments,
+  ]);
   const voicesFromProvider = voicesData?.fromProvider ?? false;
   const elevenLabsMatchedMaleVoiceOptions = useMemo(
     () =>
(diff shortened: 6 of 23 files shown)


SpicyMarinara:
@coderabbitai review

PR #3396, opened 2026-07-08
Fix v2.1.2 roleplay and call followups

### LINKED ISSUE #3392: [Issue]: #3322 still present on main — RP wizard "Use Settings Presets" skips the first message (greeting guard counts system "joined" notices)
Follow-up to #3322 (closed / marked fixed for v2.1.1). A user reports it's **still happening on current `main`**, and it is — the fix was applied but doesn't work because of the guard it uses.

**What's wrong**

Creating a Roleplay chat via the wizard's **"Use Settings Presets"** shortcut still opens the chat with no first message — only the "X has joined the chat" system notice — exactly like the original #3322 report.

**Why the fix doesn't work**

The #3322 fix added `seedInitialGreetingsIfEmpty()` to `handleShortcutApply`, guarded to seed the greeting only when the chat is "empty." But the guard uses the **total** message count, and by the time it runs the chat already contains the system **"joined the chat"** notices — so `count > 0` is always true, the guard returns early, and the greeting is never posted.

The original issue's proposed fix specifically said to guard on "no **non-system** messages yet." The implementation guards on total messages instead, which includes those system join notices.

1. Sidebar → RP tab → **New Roleplay**.
2. In the wizard, click **Use Settings Presets**.
3. Pick a chat preset and select a character that has a `first_mes`.
4. Click **Apply & Start**.

**You'll see:** the chat opens with no first message — only the "X has joined the chat" notice (same as #3322).
**Expected:** the character's `first_mes` is posted as the opening assistant message (with alternate greetings as swipes), like the full step-by-step wizard path.

**Where it is**

`packages/client/src/components/chat/ChatSetupWizard.tsx` — `seedInitialGreetingsIfEmpty` (lines 1998-2009):
```
const { count } = await api.get<{ count: number }>(`/chats/${chat.id}/message-count`);
if (count > 0) return;   // counts the system "joined" notices, so it always skips
...
for (const charId of chatCharIds) await createInitialGreetingForCharacter(charId);
```
called from `handleShortcutApply` (lines 2013 and 2023).

- The count comes from `countMessages` (`packages/server/src/services/storage/chats.storage.ts:791-794`), which counts **all** rows with **no role filter**:
  ```
  async countMessages(chatId) { const rows = await db.select({ id: messages.id }).from(messages).where(eq(messages.chatId, chatId)); return rows.length; }
  ```
- The "joined the chat" notices are `role: "system"` messages, created when characters are added to the chat (`packages/server/src/routes/chats.routes.ts:715-721`), which happens during the wizard — before **Apply & Start**.

So by the time `seedInitialGreetingsIfEmpty` runs, `count` already includes those system messages → `count > 0` → early return → no greeting. (For contrast, the full step-by-step path `finishWizard`, lines 1987-1996, calls `createInitialGreetingForCharacter` **unconditionally**, which is why it still works.)

**What's happening**

The guard was meant to avoid duplicating the greeting in the character-card quick-start flow (where an assistant greeting is pre-injected). But it keys on total message count including system join notices, so in the "New Roleplay wizard" flow — where only system notices exist, no greeting — it wrongly concludes the chat is non-empty and skips seeding.

**Suggested fix**

Guard on the presence of a **non-system** message (an actual greeting / assistant / user message), not the total count — matching the original issue's stated intent ("no non-system messages yet"). Options:
- Skip only when a non-system message already exists — e.g. add a count/lookup that excludes `role === "system"` (and likely `"narrator"`), or fetch the messages and check.
- Or specifically check whether an assistant greeting (`first_mes`) already exists before seeding, so the character-card quick-start dedup still holds while the wizard flow seeds correctly.

---
*AI-investigated against current `main` (v2.1.1 tree; fix from commit `2f3bef600` "[codex] v2.1.1 issue sweep #3331" is present but ineffective). The guard, `countMessages`, and the system-join-notice creation were all read and confirmed. Please verify in-app before fixing.*

SpicyMarinara:
## Linked issue

Closes #3392

## Why this change

- Fixes the Roleplay setup wizard shortcut path where system "joined the chat" notices made the chat look non-empty, so first messages were skipped.
- Bundles the v2.1.2 follow-up fixes requested in this working branch: summary output sizing, Illustrator prompt guidance, Conversation Calls permissions policy, and Mini Mari toast polish.

## What changed

- Checks for non-system chat messages before skipping Roleplay initial greeting seeding in the **Use Settings Presets** shortcut.
- Adds a Roleplay Chat Summary maximum output size setting and uses it for manual and automatic summaries.
- Updates the default Illustrator prompt rules to include character build, clothing, and appearance details.
- Allows same-origin microphone, camera, and screen capture for Conversation Calls via the server permissions policy.
- Removes the extra framed border around the Mini Mari surprise toast image.
- Updates CHANGELOG and Conversation Calls troubleshooting docs.

## Validation

### Manual verification notes

- Automated validation run: `pnpm check`.
- Browser/manual verification still needed for the RP wizard shortcut, Edge microphone prompt, and Mini Mari toast appearance.

## Docs and release impact

- [ ] No docs changes needed

- Not captured.

## Summary by CodeRabbit

* **New Features**
  * Added a new “Maximum output size” setting for chat summaries, available in the Summary Connection area for both manual and automatic summaries.
  * Roleplay setup now seeds initial greetings more reliably when a chat is still empty.

* **Bug Fixes**
  * Fixed microphone, camera, and screen-capture access in Chromium-based browsers for conversation calls.
  * Fixed the Roleplay setup shortcut so earlier notices no longer block the first seeded character message.

* **Documentation**
  * Updated troubleshooting guidance for browser microphone permission issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,12 +15,14 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 - Added Grok 4.5 to the xAI / Grok model list and made new xAI connections default to it.
 - Added Conversation prompt relocation macros for auto-inserted context: `{{context}}`/`{{status}}`, `{{commands}}`, `{{reactRules}}`, `{{memories}}`, and `{{lorebook}}`.
 - Added a TTS cache export control in Text to Speech settings so generated cached voice clips can be downloaded from IndexedDB.
+- Added a Roleplay Chat Summary maximum output size setting under Summary Connection, defaulting to 4096 tokens for manual and automatic summaries.
 
 ### Changed
 
 - Bumped release metadata to v2.1.2 across packages, the PWA manifest, README release pointer, Windows installer sources, Android APK metadata, and the home-page-visible app version.
 - Android `versionName` is `2.1.2` with `versionCode 31`.
 - Reworked Settings with search-first navigation, compact pinned controls, fixed top-level categories, and finer section shortcuts for faster navigation.
+- Updated the default Illustrator prompt rules so generated image prompts carry available character builds, clothing/outfits, and appearance details instead of relying on the image model to infer them.
 
 ### Fixed
 
@@ -37,6 +39,8 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 - Fixed Conversation command reminders so preset wrap format `None` no longer emits a literal `<commands>` XML wrapper (#3378).
 - Fixed roleplay streaming recovery so parallel agent events are deferred until the main assistant stream finishes, preventing late agent updates from replacing the streamed message.
 - Fixed the recovery/error page so **Internal Server Error** uses the configured chat chrome text color instead of a hard-coded pink.
+- Fixed Conversation Calls on Chromium browsers by allowing same-origin microphone, camera, and screen-capture access in the server permissions policy instead of blocking the browser permission prompt.
+- Fixed the Roleplay setup wizard's **Use Settings Presets** shortcut so system "joined the chat" notices no longer block seeding the selected character's first message (#3392).
 
 ## [2.1.1]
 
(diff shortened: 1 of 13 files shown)


SpicyMarinara:
@coderabbitai review

PR #3419, opened 2026-07-10
feat(turn-games): add 8-Ball Pool as the fourth conversation-mode table game

Romuromylus:
## Linked issue

Closes #3418
Closes #3427

## Why this change

- Conversation mode has UNO, chess, and poker as table games; 8-ball pool is the most-requested "bar game" shape — one-on-one, spectator-friendly, and a natural fit for character banter and an announcer seat.
- Selecting shots from a menu (the design this PR originally shipped) didn't feel like playing pool. This revision resolves every shot through a real 2D physics simulation: the human aims and shoots directly, and outcomes are whatever the balls actually do.

## What changed

- **New shared physics module** (`packages/shared/.../eightball/physics.ts`): pure, deterministic 2D simulation — swept-circle continuous collision detection inside fixed 1/240s steps (no tunneling at break speeds), equal-mass restitution impulses, cushion bounces, friction, pocket capture with a jaw model (only genuinely pocket-bound balls pass the cushion line inside a pocket mouth; grazers reflect off the jaw; a ball that stops inside the jaws drops). No RNG, no wall clock — rewind/replay reproduces shots exactly.
- **Engine** (`eightball/engine.ts`, `types.ts`, `tools.ts`): move union `aimed` (human: angle + power) / `menu` (character pick) / `place` (ball in hand) / `next_rack`. Characters still choose from the engine-computed candidate menu (pots/banks/safeties with advisory make-% estimates), converted to an aim vector with skill/style-based jitter and executed through the same physics. Rules derive from actual sim events: wrong-ball-first and scratches are fouls → ball in hand (kitchen-restricted after break fouls), slop counts, open table after every break, first legal pot assigns groups, early 8 loses, 8-on-the-break wins (bar rules, documented), 8+scratch loses. Race to 1/3/5, alternating breaks, player-paced racks. The seeded RNG's only remaining consumer is bot aim jitter.
- **Client** (`EightBallBoard.tsx`, `eightball-game.store.ts`): drag-to-aim with an obstruction-clipped guide line, ghost-ball contact preview, and object-ball deflection tick; vertical power slider; tap-to-place ball-in-hand with kitchen overlay and invalid-spot tinting; every shot animates from server-produced 30Hz frames (the client never simulates; `prefers-reduced-motion` and hidden tabs snap instantly). Pointer events throughout — touch works.
[...]

PR #3428, opened 2026-07-10
Fix turn-game narration leaking model chain-of-thought into chat

### LINKED ISSUE #3427: [Issue]: Turn-game bot narration leaks model chain-of-thought as the character's chat message
## Summary

Turn-game bot narration (character table talk and dealer/announcer lines) saves the model's chain-of-thought as the character's chat message when the connected model emits inline reasoning (e.g. a leading `<think>` block). With a reasoning-heavy model, the short `maxTokens` budget is often consumed entirely by the reasoning, so the persisted message is a truncated meta monologue instead of a spoken line.

## Expected behavior

After a bot move (UNO, chess, poker, 8-ball), the character speaks one short in-character line; the announcer voices one short announcement. Inline reasoning from the model is never persisted or shown, matching how the main generation pipeline strips leading thinking blocks before saving.

## Actual behavior

The character's message in chat reads like this (real example, truncated mid-sentence by the 100-token cap):

> The user wants me to say one short, natural line as Rias Gremory after making my move in 8-ball pool. I should react naturally - maybe a confident quip, a flourish, something in character. I should NOT explain strategy, reveal hidden info, or recite board state. Just a natural pool table reaction.
>
> Let me think of something Rias would say - she's confident, regal, a bit playful, competitive. She just made a good move after getting ball in

## Steps to reproduce

1. Connect a model that emits inline reasoning in `content` (a leading `<think>`/`<thinking>` block) rather than a provider-native thinking channel.
2. Start any turn game against a character (e.g. `/poker` or another turn-game command) and make a move so the bot takes its turn.
3. The bot's post-move line and/or the announcer's line contains the raw reasoning, often cut off mid-sentence.

## Environment

- Marinara Engine version: staging (a624b957)
- Install type: docker
- OS + version: Linux host
- Browser or app shell: Chrome

See the quoted chat message above (bot seat "Rias Gremory" during an 8-ball game; identical mechanism applies to poker/chess/UNO narration and dealer announcements).

## Additional context

Root cause: `narrateOutcome()` and `narrateAnnouncements()` in `packages/server/src/services/turn-games/turn-game-bot-runner.service.ts` return `res.content` verbatim, while the generation pipeline (e.g. `generate.routes.ts`, `game.routes.ts`) runs model output through `extractLeadingThinkingBlocks()` before persisting. Two compounding factors:

- No inline-thinking strip at all on the narration paths.
- `maxTokens: 100` / `120` is routinely exhausted by the reasoning itself, so the closing tag never arrives and the message is truncated mid-thought — even a strip of *closed* blocks would not recover a spoken line in that case (the correct behavior is falling back to the factual event summary, which both call sites already support for empty narration).

The bot **move** call is unaffected (it only consumes `res.toolCalls`).

## Template check

Romuromylus:
## Linked issue

Closes #3427

## Why this change

When the connected model is a reasoning model that emits its chain-of-thought inline in `content` (a leading `<think>`/`<thinking>` block) instead of a provider-native thinking channel, turn-game narration saves that reasoning verbatim as the character's chat message. With the previous 100/120-token narration budgets the reasoning routinely ate the whole budget, so what landed in chat was a truncated meta monologue ("The user wants me to say one short, natural line as … Let me think of something she would say -") instead of a spoken line. This affects all four turn games (UNO, chess, poker, 8-ball): both the bot's post-move table talk and the dealer/announcer lines.

The main generation pipeline already strips these blocks (`extractLeadingThinkingBlocks` in `generate.routes.ts` / `game.routes.ts`); the turn-game bot runner predates none of that but never adopted it.

## What changed

All changes are in `packages/server/src/services/turn-games/turn-game-bot-runner.service.ts` (plus a CHANGELOG entry):

- New `stripInlineThinking()` helper applied to both narration paths (`narrateOutcome`, `narrateAnnouncements`): strips leading inline thinking blocks using the same shared extractor as the generation pipeline, and — the case a plain strip cannot recover — returns an empty line when an **unclosed** opening tag remains at the front (maxTokens ran out mid-thought, so everything present is reasoning). Both call sites already fall back to the factual event summary on an empty line, so the worst case is now a dry-but-correct line instead of leaked chain-of-thought.
- Narration `maxTokens` raised from 100/120 to 300 so reasoning models have headroom to finish thinking and still produce the spoken line.
- Prompt hardening on both narration prompts: "Reply with the spoken line ONLY — no reasoning, no preamble."
- The bot **move** call is untouched — it only consumes `res.toolCalls`, so inline reasoning there was never persisted.

## Validation

### Manual verification notes

Done so far:

- `pnpm check` (typecheck + lint + build) passes locally.
- Local unit tests on `stripInlineThinking` (11 cases): plain lines pass through; closed `<think>`/`<thinking>`/`[think]` blocks stripped (case-insensitive, multiple consecutive blocks); the exact truncated-unclosed-block shape from the bug report returns empty (fallback path); tags mentioned mid-line are not treated as thinking.

Still to verify manually before promoting:

- Live turn-game session with a reasoning model: bot table talk and announcer lines contain only the spoken line.
- Live session with a non-reasoning model: narration behavior unchanged.

## Docs and release impact

- [ ] No docs changes needed

CHANGELOG entry added under [Unreleased] → Fixed. No version bump; no other docs reference turn-game narration internals.

Server-side narration fix — the user-visible effect is the *absence* of leaked reasoning in chat. The linked issue quotes the leaked message this fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 
 ### Fixed
 
+- Fixed turn-game bot table talk and dealer announcements leaking the model's chain-of-thought into chat when the connected model emits inline reasoning (e.g. a leading `<think>` block): narration now strips inline reasoning like the main generation pipeline, falls back to the factual event line when the output was all reasoning, and gets a larger output budget so reasoning models can still land the spoken line (#3427).
 - Completed Noodle's automatic timeline refresh scheduling: daily refresh times are now distributed across persisted local-day windows, survive restarts, collapse overdue slots into one catch-up refresh, retry safely after failures, and update an open Noodle timeline automatically.
 - Fixed PocketTTS voice refresh so built-in and custom voices returned by the provider `/v1/voices` endpoint are listed, including custom voices identified by URL/path fields (#3410).
 - Fixed Conversation Call live-mic input so Local Whisper/provider media submissions cannot queue unbounded speech segments and lock the call UI behind repeated "too many requests" failures (#3411).
--- a/packages/server/src/services/turn-games/turn-game-bot-runner.service.ts
+++ b/packages/server/src/services/turn-games/turn-game-bot-runner.service.ts
@@ -14,11 +14,13 @@
 // Invoked from the /api/generate handler ONLY when input.turnGameBots is set,
 // so it can never affect a normal conversation/roleplay generation.
 import type { FastifyReply } from "fastify";
+import { BUILT_IN_THINKING_TAG_PAIRS } from "@marinara-engine/shared";
 import type { DB } from "../../db/connection.js";
 import { logDebugOverride, logger } from "../../lib/logger.js";
 import { isDebugAgentsEnabled } from "../../config/runtime-config.js";
 import type { BaseLLMProvider, ChatMessage, LLMToolDefinition } from "../llm/base-provider.js";
 import { createLLMProvider } from "../llm/provider-registry.js";
+import { extractLeadingThinkingBlocks } from "../llm/inline-thinking.js";
 import { trySendSseEvent } from "../../routes/generate/sse.js";
 import { createCharactersStorage } from "../storage/characters.storage.js";
 import { createChatsStorage } from "../storage/chats.storage.js";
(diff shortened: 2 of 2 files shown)


Romuromylus:
Folding this into #3419 instead — the narration fix ships with the 8-ball PR.

Issue #3448, opened 2026-07-10
[Feature]: Support relocation macros ({{memories}}, {{lorebook}}, {{commands}}, ...) in {{#if}} conditions


kolacheee:
## What problem does this solve?

The Conversation-mode **relocation macros** — `{{context}}`/`{{status}}`, `{{commands}}`, `{{reactRules}}`, `{{memories}}`, `{{lorebook}}` — cannot be tested inside `{{#if}}` conditions. A preset like:

```
{{#if memories != ""}}
## Memories
{{memories}}
{{/if}}
```

**always renders the block**, even when there are no memories, so the `## Memories` header shows with nothing under it.

**Why:** relocation macros are not resolved by the shared macro engine at all — their content is injected by the route (`replaceConversationContextMacro`) *after* `resolveMacros` has already processed the `{{#if}}`. At condition-evaluation time the operand resolves to the literal token string (`"memories"`), so `"memories" != ""` is trivially true. This is unlike every value macro (`{{char_about}}`, `{{date}}`, `{{agent::TYPE}}`, …), which the engine can resolve. (Filed as the follow-up to the `{{#if}}` macro-resolution fixes in #3435.)

Users building modular Conversation presets want conditional headers/wrappers around these blocks, and want `{{#if}}` to behave consistently across all macros.

## Proposed solution

Add a **generic, opt-in deferral hook** to the shared macro engine so a caller can mark certain `{{#if}}` operands as "resolved later," reusing the exact mechanism the engine already uses to defer per-character macros in group chats:

- `resolveMacros` gains an optional `deferConditionalOperand?: (operand: string) => boolean`. When a `{{#if}}` condition depends on such an operand, the engine encodes the block as a deferred token (leaving inner macros raw) instead of evaluating it. The engine stays mode-agnostic — it never hardcodes conversation macro names.
- The conversation route supplies the predicate (the relocation-macro aliases), and after all relocation content is known (right after the memory step) runs a decode pass that evaluates each deferred conditional against the real values and substitutes the inner relocation macros.

This makes `{{#if memories != ""}}`, `{{#if lorebook contains "x"}}`, etc. work with the full operator set, consistent with all other macros. Conditions that mix a *character* macro and a *relocation* macro in the same comparison are out of scope (documented).

Also required: the slot detector must recognize a relocation alias used only as an `{{#if}}` operand (a header-only block with no bare `{{memories}}`), so retrieval still runs.

## Alternatives considered

[...]

PR #3451, opened 2026-07-10
fix(macros): resolve #3449 relocation-{{#if}} regressions (context dup, slot over-match, else spacing)

### LINKED ISSUE #3450: [Issue]: #3449 relocation-{{#if}} regressions — context double-inject, slot over-match drops retrieval, else-branch blank line
## Summary

A post-merge review of #3449 (relocation macros in `{{#if}}`) surfaced three regressions in `staging`: the current-context block is injected twice, a `{{#if}}` comparing to a relocation-alias string literal silently drops retrieval, and a deferred `{{else}}` branch keeps a stray blank line.

## Expected behavior

1. The current-context block is sent to the model **once**.
2. `{{#if topic == "memories"}}` (comparing to the literal string `"memories"`) does not affect memory recall.
3. A deferred standalone `{{#if}}…{{else}}…{{/if}}` whose else branch is selected has no leading blank line — same as an evaluated block.

## Actual behavior

1. **Context double-injection.** A preset that references `{{context}}` only inside a deferred conditional (e.g. `{{#if context != ""}}…{{context}}…{{/if}}`) leaves no literal `{{context}}` after deferral. The context tail-fallback keys on `replaceConversationContextMacro`'s return value, which is `false` for a deferred token, so it **pushes the context block as a trailing user message** — and the decode pass **also inlines it** into the system prompt. Result: the context block (latest user message, status, activity, timestamp, mentions) is sent twice, every turn. `generate.routes.ts` ~2110.

2. **Slot-detection over-match drops retrieval.** Slot detection uses a loose regex over `{{#if …}}` bodies (`conversationContextMacroConditionPattern`, ~499/528). For `{{#if topic == "memories"}}` or `{{#if user.status == "x"}}`, `\bmemories\b`/`\bstatus\b` matches inside the quoted literal / dotted operand, flipping `slots.memories` (or `.status`/`.commands`/`.reactRules`/`.lorebook`) true — but the parser-based defer predicate correctly declines. With the slot wrongly true, the code takes the "macro placed" branch (replace a bare `{{memories}}` that isn't there) instead of the normal recall injection, so **memory recall / react rules / commands / lorebook silently vanish**.

3. **Deferred else-branch blank line.** `trimDeferredStandaloneBranches` (`macro-engine.ts` ~998) trims the leading newline only for branch index 0, but the evaluate-now path strips whichever branch is selected — so a selected `{{else}}`/`{{else if}}` branch keeps a stray blank line the inline path removes.

## Steps to reproduce

1. Conversation preset with `{{#if context != ""}}Current situation:\n{{context}}{{/if}}` and no bare `{{context}}` → context block appears twice.
2. Preset with `{{#if topic == "memories"}}…{{/if}}` and no bare `{{memories}}` → memory recall no longer injected.
3. Preset with a standalone `{{#if memories != ""}}…{{else}}…{{/if}}` and empty memories → the else branch renders with a leading blank line.

## Environment

- Marinara Engine version: staging (post-#3449)
- Install type: source
- OS + version: Windows 11
- Browser or app shell: N/A (prompt-assembly)

N/A — traced in code; fix PR incoming.

## Additional context

Root cause of (1) and (2) is a slot-detection path that disagrees with the parser-based defer decision. Fix: drive slot detection from the actual deferred tokens (same parser) and key the context fallback on the slot, not the replace() return. Follow-up to #3449 / #3448.

## Template check

kolacheee:
## Linked issue

Closes #3450 (fast-follow to the merged #3449 / #3448)

## Why this change

A post-merge adversarial review of #3449 (relocation macros in `{{#if}}`) found three regressions that shipped to `staging`:

1. **Context double-injection (medium).** A preset referencing `{{context}}` only inside a deferred conditional leaves no literal `{{context}}` after deferral. The context tail-fallback keyed on `replaceConversationContextMacro`'s return — which is `false` for a deferred token — so it pushed the context block as a trailing user message **and** the decode pass inlined it. The current-context block (latest user message, status, activity, timestamp, mentions) was sent **twice, every turn**.
2. **Slot-detection over-match drops retrieval (medium).** Slot detection used a loose regex over `{{#if …}}` bodies. `{{#if topic == "memories"}}` (string literal) or `{{#if user.status == "x"}}` (dotted operand) matched the alias word and flipped the slot true, while the parser-based defer predicate correctly declined. With the slot wrongly set, the code took the "macro placed" branch (replacing a bare `{{memories}}` that isn't there) instead of the normal recall injection — **silently dropping memory recall / react rules / commands / lorebook**.
3. **Deferred else-branch blank line (low).** `trimDeferredStandaloneBranches` trimmed the leading newline only for branch 0, so a selected `{{else}}`/`{{else if}}` branch kept a stray blank line the evaluate-now path strips.

## What changed

Root cause of #1 and #2 is a slot-detection path that disagreed with the parser-based defer decision. Fix makes them share one parser:

- `packages/shared/src/utils/macro-engine.ts`:
  - New `collectDeferredRelocationConditionOperands(text)` — reads the **actual** deferred tokens and returns their condition operands (same `parseConditionExpression` the deferral uses).
  - `trimDeferredStandaloneBranches`: trim the leading newline of **whichever** branch is selected, not just branch 0.
- `packages/server/src/routes/generate.routes.ts`:
  - Replaced the loose `conversationContextMacroConditionPattern` regex with token-driven slot detection: after the conversation macro pass, map each deferred-token operand back to its key via `relocationKeyForOperand` (now shared with the defer predicate) and OR into the slots. Slot detection and deferral can no longer disagree — fixing both the over-match (dropped retrieval) and the inverse braced-operand miss. `relocationKeyForOperand` stays scoped to `CONVERSATION_RELOCATION_MACRO_KEYS`, preserving the About-me exclusion and covering `replyRules`.
  - The context fallback now keys on `conversationContextMacroSlots.context` instead of the `replace()` return, so a deferred `{{#if context}}` no longer double-injects.

## Validation

### Manual verification notes

Automated checks I ran locally (green):

- `pnpm --filter @marinara-engine/shared build` + `pnpm --filter @marinara-engine/server lint` (tsc) — clean
- `pnpm regression:prompt` — full suite passes (61 checks, incl. the deferred-relocation cases added with #3449)
- A throwaway targeted script (not committed) confirming: deferred else/else-if branch collapses with no leading blank line; `{{#if topic == "memories"}}` and `{{#if user.status …}}` do **not** defer and yield no relocation operand (so retrieval is unaffected); `{{#if char_about}}` (about-me field) is not deferred; real bare / braced-mixed / `{{#if replyRules}}` references defer and collect their key operand.

Still to verify manually in the app (please tick above once done):

1. Preset with `{{#if context != ""}}…{{context}}…{{/if}}` and no bare `{{context}}` → context block appears **once**.
2. Preset with `{{#if topic == "memories"}}…{{/if}}` and no bare `{{memories}}` → memory recall still injected.
3. Deferred `{{#if memories != ""}}…{{else}}…{{/if}}` with empty memories → else branch renders with no leading blank line.

## Docs and release impact

- [ ] No docs changes needed

No docs/version changes; behavior-only bug fixes.

N/A (prompt-assembly logic).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversation prompt handling for relocation macros used inside deferred conditional blocks.
  * Prevented duplicate context insertion when context is referenced conditionally.
  * Improved whitespace and newline handling across deferred conditional branches.

### DIFF
--- a/packages/server/src/routes/generate.routes.ts
+++ b/packages/server/src/routes/generate.routes.ts
@@ -12,6 +12,7 @@ import {
   resolveDeferredCharacterMacros,
   hasDeferredCharacterMacros,
   hasDeferredRelocationConditionals,
+  collectDeferredRelocationConditionOperands,
   DEFERRED_RELOCATION_CONDITIONAL_TOKEN_RE,
   parseDeferredConditionalPayload,
   selectConditionalPayloadBranch,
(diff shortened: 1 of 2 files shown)


kolacheee:
@coderabbitai review

Issue #3491, opened 2026-07-11
[Issue]: Code sweep: misleading names and duplicated implementations that invite wrong edits


kolacheee:
## Summary

Names that say the wrong thing and duplicated implementations that have already started to drift.

**Why this is a problem:** agents and contributors act on names. An alias whose name implies a narrower scope than it gates, refs whose names are swapped, and a load-bearing variable dressed as unused all invite confident wrong edits. Duplicated logic means a fix to one copy silently misses the other.
**What it negatively affects:** refactor safety (including automated "remove unused" passes), and behavior consistency between duplicated code paths.

## Expected behavior

Names describe scope and usage truthfully; shared logic has one implementation.

## Actual behavior

Misleading names:

- [ ] `server/.../services/generation/prose-guardian-settings.ts:108` — `shouldHoldForProseGuardianRewrite` is a verbatim alias of `shouldHoldForTextRewrite`, which fires for ANY of prose-guardian, continuity, or html rewrites. Its result gates 7 generation branches far from the definition. The name invites scoping a hold change to Prose Guardian and breaking the other two.
- [ ] `client/.../chat/HomeProfessorMariChat.tsx:1894` — `embeddedTextareaRef` is attached inside the floating body (:3073) and `floatingTextareaRef` inside the embedded body (:3693). The names are inverted relative to usage; editing "the floating textarea" by name grabs the wrong ref with no type error.
- [ ] `client/.../game/GameCombatUI.tsx:441` — `_isStreaming` carries the unused-underscore convention but is load-bearing: read in the custom-action reset effect (:1544-1556), fed live by `GameSurface.tsx:10984`. An automated unused-prop cleanup would break combat UI resets.
- [ ] `client/.../features/tracker-panel/components/sections/PersonaTrackerPanel.tsx:102` — filename does not match the export (`PersonaInventoryPanel`), breaking the sibling convention (low; TS catches misuse).

Duplicated logic already drifting:

- [ ] `client/.../chat/HomeProfessorMariChat.tsx:2916` — `renderFloatingChatBody` (~200 lines) is re-implemented inline in the non-floating return (:3519-3725) and has already drifted (different text-size classes; `MariSuggestionChips compact` prop in one copy only). Fixes to one surface miss the other.
[...]

PR #3503, opened 2026-07-11
refactor: fix misleading names and a duplicate Seedance upload (part of #3491)

kolacheee:
## Linked issue

Part of #3491 (the mechanically-verifiable subset; skipped items are re-dispositioned below and on the issue).

## Why this change

Names that say the wrong thing invite confident wrong edits — especially from automated refactors and agents acting on the name alone. This PR fixes the #3491 items whose correctness TypeScript can verify exhaustively, plus one duplicate call that caused real double work.

## What changed

- **`shouldHoldForProseGuardianRewrite` removed** (`prose-guardian-settings.ts`): it was a verbatim alias of `shouldHoldForTextRewrite`. The name implied a prose-guardian-only gate, but the predicate fires for prose-guardian, continuity, AND html rewrites, and its result gates seven branches in `generate.routes.ts`. The call site and its local variable now use the truthful name.
- **`_isStreaming` → `isStreaming`** (`GameCombatUI.tsx` + the `GameSurface.tsx` call site): the underscore convention marks unused values, but this prop is load-bearing (it drives the custom-action reset effect at ~1546). An automated "remove unused underscore props" pass would have silently broken combat UI resets.
- **`PersonaTrackerPanel.tsx` → `PersonaInventoryPanel.tsx`**: filename now matches its export, restoring the sibling convention; sole import updated.
- **Seedance duplicate upload removed** (`video-generation.ts`): the raw-URL path called `maybeUploadSeedanceReferenceImage` and, when it returned nothing, fell through to an identical second call. Both paths now converge on the single fallback — same behavior for every path, and the raw-URL path no longer attempts the same upload twice.

Re-verified against current staging and deliberately NOT changed:

- The `HomeProfessorMariChat` textarea-ref inversion **no longer exists**: the component was restructured upstream, and I verified today that `floatingTextareaRef` attaches inside the floating surface (`floatingSurfaceRef` container) and `embeddedTextareaRef` outside it. Nothing to fix.
- The `retry-agents-route` `chatMeta` shadow stays as-is: there are four parallel local parses in that route, the sweep verdict was "no active bug, latent only", and re-scoping variables in a mega-route without runtime coverage is riskier than the landmine it would defuse.
- The two duplication findings (Mari chat body, sprite crop helpers) still need behavior-equivalence verification and remain open on #3491.

## Validation

### Manual verification notes

Renames are exhaustively verified by the TypeScript strict build (`pnpm check` passes; zero references to any old name remain, checked by grep). The Seedance change is behavior-preserving by control-flow analysis: identical arguments, and every path still performs exactly one upload attempt (previously the raw-URL path could perform two). Optional live check: generate a Seedance first/last-frame clip with reference upload enabled and confirm a single upload per reference in the server log.

## Docs and release impact

- [x] No docs changes needed

N/A — internal renames; no UI surface changes.

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated Persona panel with portrait display, editable status, stats, inventory management, and inline editing controls.
  - Portrait focus and zoom updates are saved automatically while editing.

- **Bug Fixes**
  - Improved streamed text-rewrite handling for more consistent generated responses.
  - Fixed a video generation issue that could upload the same reference image more than once.
  - Corrected combat streaming state handling and Persona panel loading.


kolacheee:
@coderabbitai review

kolacheee:
@coderabbitai Verified the keepalive suggestion on `PersonaInventoryPanel.tsx` (your inline comment failed to post, but the review summary preserved it). Verdict after tracing the branch: **partially right, but not actionable in this PR** — no change made. Reasoning:

1. **This PR doesn't touch that code.** `PersonaInventoryPanel.tsx` is a 100%-similarity pure rename of `PersonaTrackerPanel.tsx` (`git diff -M` from the merge base shows `similarity index 100%`). The flush logic is pre-existing staging behavior; this PR is a naming cleanup.
2. **keepalive alone would fix almost nothing.** The flush is an unmount-only React effect cleanup. In the SPA, nothing aborts an in-flight fetch on component unmount — there is no lifecycle-tied `AbortController` anywhere in `api-client.ts` — so the normal panel-close flush already completes fine without `keepalive`. `keepalive` only matters when the *document* unloads, and no code path flushes this pending save on document unload at all. The real gap is a missing `pagehide` flush (compare `use-game-state-patcher.ts` — `beforeunload` listener + `{ keepalive: true }` — and `ChatInput.tsx`'s `pagehide` draft flush), not a missing flag on the unmount path.
3. **Exposure is small:** the at-risk data is cosmetic portrait focus/zoom inside `trackerCardColors`, behind a 180ms debounce.

If this is judged worth fixing, the correct shape is a small separate change shipping both halves together: thread a narrow `keepalive?: boolean` option through `useUpdatePersona` (destructured out of the PATCH body, following the `use-achievements.ts` precedent rather than raw `RequestInit`), plus a `pagehide` listener performing the same flush. Bolting that behavior change onto this rename PR would muddy both.

kolacheee:
Follow-up issue filed for the missing pagehide flush: #3506 — includes the agreed fix shape (narrow `keepalive?: boolean` through `useUpdatePersona` per the `use-achievements.ts` precedent, plus a `pagehide` listener) and a note to land it after this PR's rename merges.

PR #3499, opened 2026-07-11
fix: warn on restart-required SPRITE_ANIMATED_FFMPEG_TIMEOUT_MS change (part of #3487)

kolacheee:
## Linked issue

Part of #3487 (first checklist item; the other entries need maintainer wire-up-or-remove decisions and are not touched here).

## Why this change

`SPRITE_ANIMATED_FFMPEG_TIMEOUT_MS` is read once at module load (`sprites.routes.ts:121`), so editing it in a live `.env` silently does nothing until a restart, with no `[env-watcher]` warning. Its sibling `SPRITE_GENERATION_TIMEOUT_MS` had the identical gap and was fixed the same way in 9a22f1af3.

## What changed

- `packages/server/src/config/env-watcher.ts`: add the key to `RESTART_REQUIRED_KEYS` (one line, mirrors the sibling entry).
- `docs/CONFIGURATION.md`: add the key to the documented restart-required list.

## Validation

- [x] `pnpm check` passes locally
- [x] Container (Docker / Podman) built and ran without issue
- [x] Ran the app, clicked through the changes manually
- [x] Checked edge cases (light + dark mode, mobile viewport, empty states, error paths)
- [x] Above manual verification completed (describe below)
- [x] Read and followed `CONTRIBUTING.md`

### Manual verification notes

Static change to a string set consumed by one loop (`env-watcher.ts:96`); behavior is identical to the already-merged sibling key. Optional live check: edit the key in `.env` while running and confirm the restart-required warning appears in the server log.

## Docs and release impact

- [ ] No docs changes needed
- [x] Updated docs (README / CONTRIBUTING / android/README / CHANGELOG) as needed
- [ ] Version/release files updated (only if this PR includes a version bump)

## UI evidence (if applicable)

N/A - server log warning only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated configuration handling so changes to the animated media processing timeout now correctly trigger a server restart warning.

* **Documentation**
  * Clarified the list of configuration changes that require a full server restart.

Issue #3488, opened 2026-07-11
[Issue]: Code sweep: dead code catalogue (unreferenced components, exports, unreachable branches)


kolacheee:
## Summary

A verified catalogue of dead code: whole components with zero importers, exported functions nothing calls, and branches that can never run.

**Why this is a problem:** dead code is where agents go wrong quietly. They edit files that never run (a concrete example below), copy superseded patterns, and in two cases here the dead name collides with a live one, inviting the wrong import.
**What it negatively affects:** agent and contributor productivity, code search quality, and refactor safety. No user-facing behavior — that is the point.

Verification method: every entry was checked with repo-wide reference searches including barrel re-exports, lazy/dynamic imports, codegen registries, and string-keyed dispatch. Visual-Novel-related dead code (including `shared/src/types/vn.ts`, fully dead) is deliberately excluded here — it is catalogued under #3456.

## Expected behavior

Code in the repo is reachable, or it is deleted.

### Dead files (zero importers; delete candidates)

- [ ] `client/.../components/game/GameSessionBanner.tsx` — superseded by `GameSurface.handleConcludeSession`/`useConcludeSession`
- [ ] `client/.../components/game/GameStateIndicator.tsx` — its `STATE_CONFIG` is a byte-duplicate of the live one in `GameMap.tsx`
- [ ] `client/.../components/chat/ExpressionPanel.tsx` (and its `ExpressionSprite`)
- [ ] `client/.../components/chat/SpriteSidebar.tsx` — live path is `SpriteOverlay.tsx`
- [ ] `client/.../components/modals/BotBrowserModal.tsx` — Chub-only modal superseded by `BotBrowserPanel` (wired in `RightPanel.tsx:26,50`)
- [ ] `client/.../components/game/GamePartySidebar.tsx` — `GameSurface.tsx:11386` says "REMOVED: Old sidebar replaced by compact GamePartyBar"
- [ ] `client/.../components/game/GameCheckpoints.tsx` — checkpoint browse/restore UI; server checkpoints have no UI entry point
- [ ] `client/.../components/game/GameDialogueOverlay.tsx`
- [ ] `client/.../components/game/GameNpcTracker.tsx`
- [ ] `client/.../features/chat-settings/sections/ManualRepliesSection.tsx` — `ChatSettingsDrawer` imports every sibling section except this one
- [ ] `client/.../hooks/use-party-turn.ts` — party turns flow through the main SSE `/generate` pipeline
[...]

PR #3508, opened 2026-07-12
fix: remove settings, fields, and switches that silently do nothing

### LINKED ISSUE #3487: [Issue]: Code sweep: settings, fields, and switches that silently do nothing
## Summary

A batch of settings, request fields, function parameters, and switches that look functional but have no effect.

**Why this is a problem:** controls that silently do nothing break user trust ("I changed the setting and nothing happened, no error") and mislead agents into building on toggles that are inert, or into "cleaning up" values that actually matter elsewhere.
**What it negatively affects:** user configuration (changes silently ignored), debugging time, and the safety of automated refactors.

## Expected behavior

Every exposed setting, accepted request field, and function parameter either does what its name says or does not exist.

## Actual behavior

- [ ] `packages/server/src/config/env-watcher.ts:45` — `RESTART_REQUIRED_KEYS` omits `SPRITE_ANIMATED_FFMPEG_TIMEOUT_MS` (read once at boot, `sprites.routes.ts:121`). Editing it in a live `.env` silently no-ops with no `[env-watcher]` warning. Identical pattern to the sibling key already fixed in #3457 / 9a22f1af3.
- [ ] `packages/server/src/routes/updates.routes.ts:492` — `/apply` accepts `currentBuild` (and the client sends it) but the handler validates only `currentVersion`/`currentCommit`; the field does nothing.
- [ ] `packages/client/src/lib/tts-dialogue.ts:39` — `ttsConfigMatchesSpeaker` unconditionally returns `true`; the `dialogueScope`/`dialogueCharacterName` config it appears to honor has no effect (no UI ever sets scope to "character"). It gates 4 live call sites.
- [ ] `packages/client/src/features/tracker-panel/lib/world-state-display.ts:414` — `getWorldAmbienceStyle(_state)` ignores its argument and returns a static CSS variable (real implementation introduced in f9fccda75, gutted in 454cfcfb5). The call site still passes full world state, reinforcing the illusion.
- [ ] `packages/server/src/services/game/element-reactions.service.ts:447` — the `genshin` and `hsr` reaction presets are unreachable: nothing sets `meta.gameSetupConfig.elementPreset`, so only `default` ever applies.
- [ ] `packages/client/src/components/presets/PresetEditor.tsx:339` — the preset's own `parameters` field round-trips on save, but no tab or field ever edits it.
- [ ] `packages/server/src/routes/game.routes.ts:5928` — the schema accepts `characterConnectionId`, it is destructured at 5818, and then never reused (the write hardcodes null) while sibling fields are threaded correctly. Wire it up or remove the schema field.
- [ ] `packages/server/src/utils/security.ts:623` — `resolveValidatedImage`'s `contentTypeHeader` parameter is never read (the body sniffs magic bytes only); all 6 callers pass a real header, implying it validates. The AVIF path is unreachable (needs `expectedExt`, never forwarded).
- [ ] `packages/client/src/components/chat/ConversationCallSurface.tsx:1245` — `soundboardEnabled = true` constant leaves 4 permanently one-sided guards; `canSpeak` (:110) is a write-only field.
- [ ] `packages/server/src/routes/game.routes.ts:9624` — `generateSceneWrapAssetsInline = false` kill-switch permanently disables ~170 lines across 3 branches (self-documented; keep or delete, but decide).

## Steps to reproduce

Each entry lists the exact file/line; the fastest check is reading the reference and searching for consumers of the value in question.

## Environment

- Marinara Engine version: 2.2.0 (staging @ 45b650de9)
- Install type: source
- OS + version: N/A — verified by reading the code
- Browser or app shell: N/A

N/A.

## Additional context

Found by a repo-wide read-only sweep; each entry adversarially re-verified (consumers searched repo-wide, including dynamic dispatch) before filing. No fixes started. Part of a coordinated set of sweep catalogue issues filed together.

## Template check

kolacheee:
## Linked issue

Closes #3487 (together with already-open PR #3499, which covers the env-watcher `SPRITE_ANIMATED_FFMPEG_TIMEOUT_MS` item — the only catalogue entry not in this PR).

## Why this change

Controls that silently do nothing break user trust and mislead agents into building on inert toggles. This PR removes the remaining nine no-op surfaces from the #3487 catalogue. Every item was re-verified against current staging (the catalogue was verified 129 commits ago) and traced through git history before removal — each one is a deliberately-dismantled or born-dead surface, not a dropped wire-up, and the reasoning per item is in the commit message.

## What changed

- **updates `/apply`**: the `currentBuild` body field was never read by the handler (the build label is exactly `currentVersion`+`currentCommit`, both already validated); removed from the body type and the client sender. Old clients still sending it remain compatible (no runtime schema rejects unknown keys).
- **TTS dialogue scope**: `ttsConfigMatchesSpeaker` returned `true` unconditionally since its first commit, so the `dialogueScope`/`dialogueCharacterName` config fields gated nothing in any shipped version; removed the stub, its four call-site gates, and the schema fields. Stored configs carrying the stale keys parse fine (zod strips unknown keys). Game voice-line cache keys are preserved byte-for-byte via fixed placeholders in `buildVoiceConfigSignature`, so no user pays TTS regeneration cost.
- **world-state display**: `getWorldAmbienceStyle` ignored its state argument since 454cfcfb5 deliberately replaced dynamic ambience with the user-controlled tracker-panel background theming; the signature now says so. Two keyword branches that returned their own fallback value collapsed (the cherry-blossom/aurora pair differs from its fallback and is untouched). This also covers the two world-state-display entries from #3488's no-op-branch list.
- **element reactions**: the `genshin`/`hsr` presets were unreachable — nothing ever writes `gameSetupConfig.elementPreset` (the setup schema strips it), and the two `/game/elements` discovery endpoints had zero callers. Removed presets, endpoints, and the orphaned shared `ElementPresetName` type. **Maintainer note:** this was flagged as a plausibly-staged feature (the client renderer has display support for these element ids, which is kept — the GM can emit arbitrary element strings); removal over wire-up was an explicit ruling by @kolacheee. The live default-preset combat threading is deliberately untouched to avoid modifying live damage-resolution code.
- **PresetEditor**: the `parameters` round-trip is a leftover from the Parameters tab removed in v1.4.0. The server PATCH is field-conditional and merging, so omitting the field preserves imported ST/Marinara parameters exactly — and removal closes a real stale-snapshot clobber vector between two concurrently-open editors. The server-side `parameters` field is live generation data and is untouched.
- **game `/create`**: `characterConnectionId` was accepted, destructured, and discarded (the write hardcoded null) — the character-connection tier was deliberately dismantled in v1.5.5. Removed the schema field, destructure, null metadata write, and client type. Third-party senders get silent zod stripping, identical net effect to today.
- **`resolveValidatedImage`**: the `contentTypeHeader` param was never read (a 2f3bef600 refactor remnant); all six bot-browser callers passed it under the impression it validates. Magic-byte sniffing is the authoritative check and the signature now admits it. AVIF behavior unchanged (these endpoints never accepted AVIF, before or after the helper existed).
- **ConversationCallSurface**: `soundboardEnabled = true` (upstream #3310 committed to always-on and removed one guard itself — this finishes that cleanup), the write-only `Participant.canSpeak` field, and the command-message formatting switch that `visibleCallMessages` filters out before the only render site. Live command execution (the turn handler) is a separate, untouched path. Covers the #3488 command-switch entry.
- **scene-wrap inline asset generation** (~185 lines): `generateSceneWrapAssetsInline = false` was born false in 0b7e03c68 — the same commit that built the richer deferred `/game/generate-assets` replacement — and the route's own comment documents why inline generation is architecturally wrong here (client scene-analysis timeout). The unreachable branches had been hand-mirrored through 6+ refactors to keep compiling. Also deletes the orphaned `applyGeneratedIllustration` helper, and stops per-character sprite-file disk reads on every scene-wrap request that only fed never-read maps (pure perf win).

Two adversarial-review follow-ups included: the orphaned `ElementPresetName` shared type, and two `docs/conversation/calls.md` lines that described a soundboard enable condition that doesn't exist.

## Validation

### Manual verification notes

Validation performed: every item re-verified against current staging by a dedicated read-only verification pass (repo-wide consumer searches including dynamic dispatch, plus git-history intent tracing), exact-match scripted edits with assertions, `pnpm check` after each batch, and a 6-lens adversarial review of the final diff (regression / orphan / scope lenses per branch) whose findings were fixed in the second commit.

Manual checks still worth doing in a browser:

- Manually verify a conversation call: soundboard button opens/plays sounds, call messages render, TTS speaks dialogue with "Only read dialogues" on and off.
- Manually verify game mode: scene-wrap still resolves backgrounds/illustrations via the deferred `/game/generate-assets` flow, and combat rounds still apply element reactions (default preset).
- Manually verify a preset edit + save preserves generation parameters on a preset imported from SillyTavern.
- Manually verify the updates panel Apply flow still works (server ignores the removed field either way).

## Docs and release impact

- [x] No docs changes needed

(Two in-app guide lines in `docs/conversation/calls.md` were corrected as part of the change itself; no README/CHANGELOG impact.)

No visual changes — all removals are of surfaces that never rendered or never had any effect. The soundboard button renders identically (it was gated by a hardcoded `true`).

## Summary by CodeRabbit

* **New Features**
  * Soundboard controls are now consistently available during calls, including sound uploads with supported formats and file-size guidance.
  * Missing scene backgrounds and illustrations can be generated after scene preparation, improving scene-loading flow.
* **Bug Fixes**
  * Improved avatar image validation for proxied character images.
  * Updated voice playback and caching to provide more consistent dialogue narration.
* **Changes**
  * Removed legacy element-reaction presets and preset parameter editing.
  * Simplified text-to-speech settings by removing dialogue-scope options.


kolacheee:
@coderabbitai review

SpicyMarinara:
Integrated current staging and resolved the TTS conflict while preserving the newer HTML-safe dialogue extraction.

CodeRabbit AVIF note was verified and skipped: before this PR, resolveValidatedImage(buf, contentTypeHeader) never read contentTypeHeader and already called isAllowedImageBuffer(buf) without an extension hint. Removing the unused parameter does not change AVIF acceptance or any other validation behavior.

Validation: pnpm regression:tts, pnpm regression:roleplay, and pnpm check all pass.

PR #3507, opened 2026-07-11
Fix generation feedback, reasoning, Noodle, and portrait persistence

SpicyMarinara:
## Why

Several user-facing generation and persistence paths were inconsistent: Game setup offered almost no progress feedback, the Conversation runtime silently discarded the UI default reasoning effort, Noodle could repeat characters and interactions while text-only models lacked image context, validation errors rendered as `[object Object]`, and pending portrait edits could be lost on page exit.

This PR also carries the previously completed Roleplay streaming, rewrite comparison, TTS markup filtering, shell motion, and Conversation response-envelope cleanup requested by the maintainer.

## What changed

- Added a visible, accessible Game world-generation progress bar with elapsed time and first-turn phase labels.
- Aligned the shared and server reasoning default with the UI `Maximum` default while preserving an explicit disabled value.
- Added optional Noodle image captioning through a selectable vision-capable connection.
- Deprioritized accounts used in the preceding Noodle refresh unless they are tagged or directly involved.
- Rejected repeated same-post activity and self-replies unless a tag or direct reply creates legitimate new activity.
- Flattened structured API validation failures into readable messages.
- Flushed pending persona portrait focus/zoom saves on `pagehide` using a narrow keepalive request.
- Updated the v2.2.0 changelog and deterministic regressions.

Closes #3495
Closes #3498
Closes #3505
Closes #3506

## Validation

- `pnpm check`
- `pnpm version:check`
- `pnpm regression:noodle`
- `pnpm regression:issues`
- `pnpm regression:prompt`
- `pnpm regression:providers`
- `pnpm exec playwright test -c playwright.config.ts --project=desktop-chromium --workers=1 --grep "Noodle interface icons consistently use Noodle blue|Noodle settings persist through refetch and reload"` (2 passed)

The full parallel `pnpm smoke:ui` run reached 16 passing tests before the shared test server began returning HTTP 429 across unrelated cases; the affected Noodle checks pass when rerun serially.

## Manual verification needed

- [ ] Manually verify the Game setup progress animation on desktop and mobile.
- [ ] Manually verify GLM reasoning through NanoGPT with default and explicitly disabled effort.
- [ ] Manually verify Noodle captioning with a vision caption model and a text-only timeline model.
- [ ] Manually verify portrait crop persistence after immediate refresh/tab exit.
- [ ] Manually verify the carried Roleplay streaming, rewrite switching, TTS, and shell motion changes.

## Summary by CodeRabbit

* **New Features**
[...]

PR #3510, opened 2026-07-12
test: align TTS regression with simplified API

SpicyMarinara:
## Why

Follow-up integration fix for #3507 and #3508. #3508 removed the obsolete TTS dialogue-scope argument, while #3507 added two regression calls using the previous signature. Once both PRs were merged, the combined prompt regression passed the removed config object as a speaker and failed.

## What changed

- update the two HTML-safe TTS regression calls to use the current (text, fallbackSpeaker) signature
- production TTS behavior is unchanged

## Validation

- [ ] pnpm regression:prompt
- [ ] pnpm regression:roleplay
- [ ] pnpm check

All commands above were run locally and passed; checkboxes remain for maintainer verification.

## Summary by CodeRabbit

* **Tests**
  * Updated regression tests to reflect the current dialogue extraction interface.
  * Preserved existing expected dialogue extraction results for speaker-scoped and narrator cases.


SpicyMarinara:
@coderabbitai review

PR #3605, opened 2026-07-14
Fix shutdown write drain and remote admin folder picker

### LINKED ISSUE #3602: [Issue]: Graceful shutdown can drop writes queued during an active file-store flush
## Summary

Severity: **Medium — rare data-loss race.** Graceful shutdown can finish while an asynchronous file-store flush is still running and a newer dirty write is waiting behind it. The process then exits successfully, but the latest mutation may not have reached disk.

## Expected behavior

`FileBackedStore.close()` should wait for any active flush to finish, then flush changes that accumulated during that write before reporting that shutdown is complete.

## Actual behavior

When `flush()` is called while `this.saving` is true, it marks the store dirty and returns immediately. `close()` clears the timers and calls `await this.flush(true)`, but that await also returns immediately if a flush is already active. The active flush does not schedule or drain a follow-up pass when it finishes, and the normal SIGINT/SIGTERM path then exits the process.

This leaves a narrow sequence where a mutation made after the active flush captured its snapshot is absent from the file even though graceful shutdown completed without an error.

## Steps to reproduce

1. Use the file-native database backend and delay or block `atomicWriteFile` so a flush remains active after the target table has been serialized.
2. Mutate the table and allow the 750 ms debounce to start that flush.
3. Mutate the same table again while the first write is still active.
4. Send SIGINT/SIGTERM or otherwise trigger the graceful `app.close()` path.
5. Reopen the store and inspect the table. The second mutation is absent.

## Environment

- Marinara Engine version: 2.2.2 (`bcd4e836`)
- Install type: source
- OS + version: Windows 11 Pro 10.0.26200 (build 26200)
- Browser or app shell: N/A (storage-layer deterministic reproduction)

No server error is emitted because `close()` resolves successfully.

Deterministic ephemeral reproduction output:

```json
{"closeMs":0,"diskIds":["before-active-flush"]}
```

The second insert occurred during the active flush before `close()`, but was still absent from disk after the original flush was allowed to finish. A negative control with the same two inserts followed by an ordinary close, with no active flush, persisted both rows.

## Additional context

- Related: #2527 changed the in-app updater to use graceful `app.close()` so the file-backed store would be flushed before exit.
- Related: #2528 moved file-store persistence off the event loop. The current residual race exists in the asynchronous flush lifecycle.
- The relevant async flush behavior was introduced in `58ffdfd7` and adjusted in `55aac5cb`.
- There are no known user reports for this exact case. It requires a mutation during a narrow active-write window followed immediately by shutdown, so it is likely uncommon despite the data-loss impact.
- A focused fix would track the active flush promise, have `close()` await it, then drain any dirty tables accumulated during that flush. Persistent I/O errors should still propagate rather than causing an endless retry loop. No file-format or public API change should be required.
- Verified with an ephemeral deterministic reproduction; no test file needs to be committed.

## Template check

Please **uncheck (untick)** the checkbox below before submitting this issue.

- [ ] I have removed any checkmark from this box, and I understand that leaving it checked may cause the issue to be closed automatically.

### LINKED ISSUE #3603: [Issue]: Remote admin folder picker omits the configured admin header
## Summary

Remote administrators cannot use Agent Editor's custom-music folder picker even after correctly configuring and saving `ADMIN_SECRET`. The picker calls the privileged endpoint with direct `fetch()`, so it omits the `X-Admin-Secret` header that Marinara's shared API client normally supplies.

This is one confirmed user-facing instance of a wider client architecture problem: the current client contains 52 direct internal `/api` fetches across 12 modules despite the documented requirement to use `api.*` from `@/lib/api-client`.

## Expected behavior

After a remote user saves the matching secret under **Settings → Advanced → Admin Access**, selecting a custom music folder should send that secret through Marinara's shared API client and open the server-side folder picker.

## Actual behavior

`AgentEditor` calls `POST /api/game-assets/pick-local-music-folder` with direct `fetch()` and only a `Content-Type` header. The route intentionally calls `requirePrivilegedAccess(...)`, so a non-loopback request is rejected with HTTP 403 and `Invalid or missing X-Admin-Secret header` even though the browser has a valid saved admin secret.

The server is failing closed as intended; the client simply does not attach the configured credentials for this call.

## Steps to reproduce

1. Serve Marinara for access from another device, configure Basic Auth and `ADMIN_SECRET`, and open it from a non-loopback browser.
2. In **Settings → Advanced → Admin Access**, save the same admin secret.
3. Open Agent Editor, configure custom music to use an external folder, and click the folder-selection action.
4. Observe that `POST /api/game-assets/pick-local-music-folder` returns HTTP 403 with `Invalid or missing X-Admin-Secret header` instead of opening the folder picker.

## Environment

- Marinara Engine version: 2.2.2 (`bcd4e836`)
- Install type: source
- OS + version: Windows 11 Pro 10.0.26200 (build 26200)
- Browser or app shell: remote browser/non-loopback client

Relevant client call:

```ts
const res = await fetch("/api/game-assets/pick-local-music-folder", {
  method: "POST",
  headers: { "Content-Type": "application/json" },
});
```

Relevant server response path:

```text
403 Invalid or missing X-Admin-Secret header
```

This was confirmed from the request and authorization paths; it has not been manually reproduced against a separate remote device.

### Focused fix

Route this request through the existing shared client, for example `api.post<{ success: boolean; path: string }>("/game-assets/pick-local-music-folder")`. `apiFetch()` already attaches the saved `X-Admin-Secret`, adds the CSRF header for unsafe methods, disables caching, and normalizes error handling.

Related closed issue: #3055 addressed server-side access around local music files. This report is different: the picker is correctly privileged, but its legitimate remote-admin client does not send the configured credential.

### Wider direct-fetch inventory

The same audit found **52 direct internal `/api` fetches across 12 modules**:

| Module | Calls |
| --- | ---: |
| `components/bot-browser/BotBrowserView.tsx` | 25 |
| `components/agents/AgentEditor.tsx` | 11 |
| `components/modals/STBulkImportModal.tsx` | 4 |
| `components/panels/SettingsPanel.tsx` | 3 |
| `hooks/use-scene-analysis.ts` | 2 |
| `App.tsx` | 1 |
| `components/chat/ChatBranchSelector.tsx` | 1 |
| `components/layout/ChatSidebar.tsx` | 1 |
| `components/ui/GifPicker.tsx` | 1 |
| `hooks/use-knowledge-sources.ts` | 1 |
| `lib/tts-service.ts` | 1 |
| `stores/chat.store.ts` | 1 |

Not every direct call is necessarily producing a visible bug today, and raw/streaming responses may need `api.raw()` rather than the normal JSON helpers. However, all direct internal calls bypass some combination of shared admin headers, CSRF headers, `cache: "no-store"`, fallback-header handling, and normalized errors.

The client instructions already say: **“Don't use `fetch()` directly. Always use `api.*` from `@/lib/api-client`.”** The smallest sustainable follow-up is incremental migration using `api.raw`, `api.upload`, or the normal helpers, followed by an ESLint restriction preventing direct internal `/api` fetches outside `api-client.ts`.

## Template check

SpicyMarinara:
## Why

Graceful shutdown could report success before a newer file-native mutation reached disk, and the Agent Editor folder picker bypassed the shared API client, causing legitimate remote administrators to receive a 403 despite saving the correct admin secret.

Closes #3602
Closes #3603

## What changed

- track and await the active file-native flush
- drain writes queued during that flush before shutdown completes
- propagate a persistent close-time write failure instead of looping or silently succeeding
- route the privileged custom-music folder picker through `api.post` so admin and CSRF headers are attached
- add deterministic active-flush shutdown coverage and a regression guard for the privileged client call
- update the v2.2.2 changelog

## Risk boundary

- Entrypoints: file-native `flush()` / `close()`, Agent Editor custom-music folder selection
- File format: unchanged
- Normal debounced flush behavior: retained
- Negative controls: close cannot resolve while the captured write is blocked; persistent close-time I/O failure rejects
- Positive proof: a mutation queued after snapshot capture is present on disk after close
- Remote-device manual reproduction: still required

## Test plan

- [ ] Manually delay an active file-native write, mutate during it, close Marinara, and verify both mutations persist after restart
- [ ] Manually simulate a persistent close-time I/O failure and verify shutdown reports failure
- [ ] Manually use the custom-music folder picker from a non-loopback browser with matching `ADMIN_SECRET`
- [ ] Manually verify the picker still works from localhost
- [ ] Run `pnpm regression:issues`
- [ ] Run `pnpm check`
- [ ] Run `pnpm smoke:ui`

## Automated validation performed

- `pnpm regression:issues`
- `pnpm check`
- `pnpm smoke:ui` (54 passed, 26 intentionally skipped)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 
 ### Fixed
 
+- Fixed graceful shutdown completing before a file-native storage flush could drain newer writes queued during its active disk write, and made persistent close-time I/O failures reject shutdown instead of silently reporting success (#3602).
+- Fixed the Agent Editor custom-music folder picker omitting saved remote-admin and CSRF credentials by routing its privileged request through the shared API client (#3603).
 - Fixed dense Connections editor text flickering or shifting in Chromium browsers at the accent pulse's 500 ms update cadence. Editor reading surfaces now retain the selected base accent while headers and explicit accent chrome continue to pulse (#3599).
 - Fixed recalled-memory truncation splitting supplementary Unicode characters such as emoji or Lydian text into invalid UTF-16 surrogate halves, which caused strict provider JSON parsers to reject otherwise valid Conversation generations (#3596).
 - Fixed selected background-library images immediately disappearing in Roleplay mode because the renderer probed the GET-only image route with an unsupported HEAD request (#3592).
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "regression:providers": "pnpm build:shared && pnpm --filter @marinara-engine/server exec tsx ../../scripts/regressions/provider-compat.regression.ts",
     "regression:jsonish": "pnpm --filter @marinara-engine/server exec tsx ../../scripts/regressions/jsonish-output.regression.ts",
     "regression:tts": "pnpm build:shared && pnpm --filter @marinara-engine/server exec tsx ../../scripts/regressions/tts-source-persistence.regression.ts",
-    "regression:issues": "pnpm --filter @marinara-engine/server exec tsx ../../scripts/regressions/open-issues.regression.ts",
+    "regression:issues": "pnpm --filter @marinara-engine/server exec tsx ../../scripts/regressions/open-issues.regression.ts && pnpm --filter @marinara-engine/server exec tsx ../../scripts/regressions/file-backed-shutdown.regression.ts",
     "regression:roleplay": "pnpm build:shared && pnpm --filter @marinara-engine/server exec tsx ../../scripts/regressions/roleplay-streaming.regression.ts",
     "smoke:ui": "playwright test -c playwright.config.ts",
     "test": "node ./scripts/check-windows-installer-layout.mjs && pnpm -r run test",
--- a/packages/client/src/components/agents/AgentEditor.tsx
+++ b/packages/client/src/components/agents/AgentEditor.tsx
@@ -7,6 +7,7 @@ import { useQueryClient } from "@tanstack/react-query";
 import { toast } from "sonner";
 import { useUIStore } from "../../stores/ui.store";
 import { showConfirmDialog } from "../../lib/app-dialogs";
+import { api } from "../../lib/api-client";
 import {
   agentKeys,
   useAgentConfigs,
@@ -1343,14 +1344,8 @@ export function AgentEditor() {
 
   const handleSelectCustomMusicFolder = useCallback(async () => {
     try {
-      const res = await fetch("/api/game-assets/pick-local-music-folder", {
-        method: "POST",
-        headers: { "Content-Type": "application/json" },
-      });
-      const data = (await res.json().catch(() => ({}))) as { success?: boolean; path?: string; error?: string };
-      if (!res.ok || data.success !== true || !data.path) {
-        throw new Error(data.error ?? "No folder selected.");
-      }
+      const data = await api.post<{ success: boolean; path: string }>("/game-assets/pick-local-music-folder");
+      if (data.success !== true || !data.path) throw new Error("No folder selected.");
       setLocalCustomMusicExternalFolder(data.path);
       setLocalCustomMusicSource("folder");
       setDirty(true);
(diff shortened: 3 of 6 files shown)


SpicyMarinara:
@coderabbitai review

PR #3761, opened 2026-07-17
Fix NovelAI resolved-size reporting

### LINKED ISSUE #3758: [Issue]: v2.3.3 — Dynamic resolution works, but the image generation UI displays the pre-resolution dimensions
Follow-up to #3726 / #3732.

**Correction to my original report:** I initially filed this as "dynamic resolution has no effect." That was wrong — the feature works. The generated images **are** the correct dimensions. The problem is that the size shown in the image generation UI reports the *requested/configured* dimensions rather than the resolved size that was actually sent, which is what made it look broken.

## Actual

- Generated files are correctly sized: portrait (832×1216) for a single subject, landscape (1216×832) for three or more — `resolveNovelAiSize` is doing its job.
- The image generation page displays the connection's configured/requested dimensions instead, so the reported size doesn't match the file that was produced.

## Expected

The displayed size should reflect the dimensions actually requested from NovelAI (i.e. the value returned by `resolveNovelAiSize`), so it matches the generated image.

## Why it's worth fixing

It's cosmetic, but actively misleading: it makes a working feature look like it's doing nothing. I only realised the feature worked after checking the generated files' actual pixel dimensions — the UI had me convinced it was broken, and I filed it as such.

## Notes

- Reproduced on a **brand-new** NovelAI connection with **Choose resolution from character count** enabled and an empty Prompt Prefix.
- Suggest the label read the resolved size rather than `request.width` / `request.height`.

---

## Separate latent issue in the same code path

Unrelated to the display bug above, but noticed while looking into this — `resolveNovelAiSize` is called with `mergedPrompt`:

```js
// ~L1396
const mergedPrompt = mergePromptPrefix(defaults.promptPrefix, request.prompt);
const prompt = prepareNovelAiPrompt(mergedPrompt, "prompt", model);
// ...
// ~L1414
const size = resolveNovelAiSize(request, mergedPrompt, defaults);
```

`mergedPrompt` includes the connection's **Prompt Prefix**. Since `detectNovelAiSubjectCount` scans everything before the first `|` and **sums every count token it finds**:

```js
for (const match of subjectTokens)
    tokenCount += Number.parseInt(match[1] ?? "0", 10);
```

…any count token in a Prompt Prefix would be added to the scene's subject count. E.g. a prefix containing `2girls` plus a `1boy` scene resolves to 3 → landscape instead of portrait.

This doesn't affect me (my Prompt Prefix is empty), but it looks unintended. Passing `prompt` (the prepared scene prompt) rather than `mergedPrompt` would keep prefix text out of subject detection.

## Also worth a look — upgrade path for existing profiles

```js
function resolveNovelAiDefaults(request) {
    if (request.imageDefaults?.service === "novelai" && request.imageDefaults.novelai) {
        return request.imageDefaults.novelai;
    }
    return DEFAULT_NOVELAI_DEFAULTS;
}
```

This returns the stored profile as-is rather than merging it over the defaults. NovelAI profiles saved **before** v2.3.3 won't contain `dynamicResolutionBySubjectCount`, `styleReferenceStrength`, or `styleReferenceFidelity`, so those features would silently no-op for existing users on upgrade.

The editor masks this, since it fills gaps via `value.novelai ?? createDefaultImageGenerationProfile("novelai").novelai!` — the checkbox still appears ticked even if the key never reaches the server.

Something like:

```js
return { ...DEFAULT_NOVELAI_DEFAULTS, ...request.imageDefaults.novelai };
```

## Environment

- Marinara Engine **v2.3.3**, Windows, launched via `MarinaraLauncher.exe`
- Stock install — no local modifications
- NovelAI native API (`https://image.novelai.net`), model `nai-diffusion-4-5-full`, Opus tier

---

Apologies for the inaccurate original title — happy to test any fix.

SpicyMarinara:
## Why

NovelAI dynamic resolution correctly changed the generated image size, but image prompt review still displayed the configured dimensions. Subject-count tokens in a connection Prompt Prefix could also change scene orientation, and partial pre-v2.3.3 profiles could omit newer defaults.

Closes #3758.

## What changed

- report subject-count-resolved dimensions from the final scene prompt in every image prompt review surface
- exclude Prompt Prefix from subject-count detection while still sending the prefix to NovelAI
- merge legacy partial NovelAI profiles over current defaults
- preserve requested dimensions for non-native NovelAI proxies, other providers, and disabled dynamic sizing
- add positive and negative regression rows for native NovelAI, prefixes, proxies, other providers, and legacy profiles
- update the changelog

## Validation

Automated evidence:

- PASS: `pnpm regression:issues`
- PASS: `pnpm check`
- PASS after CodeRabbit cleanup: focused open-issues regression

Manual verification remaining:

- [ ] Review a one-subject NovelAI prompt and confirm the UI reports 832x1216
- [ ] Review a three-subject NovelAI prompt and confirm the UI reports 1216x832
- [ ] Confirm a Prompt Prefix count token does not change scene orientation

## Summary by CodeRabbit

* **Bug Fixes**
  * Image prompt previews now display dimensions that match the actual image request, including NovelAI subject-based sizing.
  * Prompt prefixes no longer affect scene-based image dimensions.
  * Legacy and partial NovelAI settings now receive appropriate default values.
  * Preview sizing is corrected across backgrounds, characters, galleries, storyboards, sprites, illustrations, and retry flows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 
 ### Fixed
 
+- Made image prompt review display the subject-count-resolved dimensions actually sent to native NovelAI, kept Prompt Prefix count tokens out of scene sizing, and filled new NovelAI settings for legacy partial profiles (#3758).
 - Kept existing cropped Character avatars contained inside the Metadata upload preview instead of allowing the image to cover the card editor (#3741).
 
 ## [2.3.3]
--- a/packages/server/src/routes/backgrounds.routes.ts
+++ b/packages/server/src/routes/backgrounds.routes.ts
@@ -26,6 +26,7 @@ import { createPromptOverridesStorage } from "../services/storage/prompt-overrid
 import { buildBackgroundProviderPrompt, generateChatBackground } from "../services/game/game-asset-generation.js";
 import { resolveConnectionImageDefaults } from "../services/image/image-generation-defaults.js";
 import { loadImageGenerationUserSettings } from "../services/image/image-generation-settings.js";
+import { resolveImagePromptReviewSize } from "../services/image/image-prompt-review.js";
 import { resolveImageConnectionFallback } from "../services/generation/media-connection-fallback.js";
 import { resolveGameSetupArtStylePrompt } from "@marinara-engine/shared";
 
@@ -472,6 +473,13 @@ export async function backgroundsRoutes(app: FastifyInstance) {
       promptOverride: context.promptOverride?.prompt,
       negativePromptOverride: context.promptOverride?.negativePrompt,
     });
+    const previewSize = resolveImagePromptReviewSize({
+      connection: context.imgConn,
+      prompt: compiled.prompt,
+      width: context.imageSettings.background.width,
+      height: context.imageSettings.background.height,
+      imageDefaults: resolveConnectionImageDefaults(context.imgConn),
+    });
 
     return {
       items: [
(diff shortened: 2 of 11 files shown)


SpicyMarinara:
@coderabbitai review

Issue #3623, opened 2026-07-14
[Issue]: [Bug] Escaping <> mangles standards <START> example dialog format


SpicyMarinara:
## Summary
Imported from the Discord thread "[Bug] Escaping <> mangles standards <START> example dialog format".

## Expected behavior
Not provided in the Discord context.

## Actual behavior
See the Discord context transcript below.

## Steps to reproduce
1. Not provided in the Discord context.

## Environment
- Marinara Engine version: Not provided
- Install type: Not provided
- OS + version: Not provided
- Browser or app shell: Not provided

## Logs, screenshots, or video
See any attachment links and screenshots in the context below.

## Additional context
This issue was created automatically from Discord support context.

Discord thread: [Bug] Escaping <> mangles standards <START> example dialog format
Discord thread link: https://discord.com/channels/1417099416812392641/1525229638115135561
Requested by: Marinara (marinara_spaghetti, Discord ID 169561011928432653)

<details>
<summary>Discord thread transcript (9 messages)</summary>

```text
[2026-07-10T19:58:03.123Z] Kat: Version: Staging, may affect other versions

Standard `<START>` marker in example dialog gets escaped to `&lt;START>`

[attached image 1: https://cdn.discordapp.com/attachments/1525229638115135561/1525229638429966456/image.png?ex=6a57e64b&is=6a5694cb&hm=daad75ec8935544db47f14c04ea1c40009d1e9941688077a6ca5cd69fabace75&]
[Discord message: https://discord.com/channels/1417099416812392641/1525229638115135561/1525229638115135561]

[2026-07-10T20:04:31.488Z] Marinara: Wtf, I already fixed this.
[Discord message: https://discord.com/channels/1417099416812392641/1525229638115135561/1525231267036266567]

[2026-07-10T20:04:35.459Z] Marinara: Like twice???
[Discord message: https://discord.com/channels/1417099416812392641/1525229638115135561/1525231283691720805]

[2026-07-10T20:04:49.846Z] Marinara: What is causing this to regress constantly?
[Discord message: https://discord.com/channels/1417099416812392641/1525229638115135561/1525231344035168316]

[2026-07-10T20:05:05.592Z] Marinara: @Professor Mari GitHub Issue
[Discord message: https://discord.com/channels/1417099416812392641/1525229638115135561/1525231410078810162]

[...]

Issue #3441, opened 2026-07-10
[Issue]: <START> example dialog marker is escaped to &lt;START&gt;


SpicyMarinara:
## Summary
In Staging, the standard `<START>` marker in example dialog format is being escaped to `&lt;START&gt;`. This appears to be a regression that has reportedly been fixed before.

## Expected behavior
The example dialog marker should remain as the literal `<START>` text.

## Actual behavior
The marker is converted/escaped to `&lt;START&gt;`, mangling the standard example dialog format.

## Steps to reproduce
1. Use a Staging build of Marinara Engine.
2. Create or view example dialog containing the standard `<START>` marker.
3. Observe that `<START>` is rendered or saved as `&lt;START&gt;`.

## Environment
- Marinara Engine version: Staging
- Install type: Not provided
- OS + version: Not provided
- Browser or app shell: Not provided

## Logs, screenshots, or video
Attached image showing `<START>` escaped to `&lt;START&gt;`: https://cdn.discordapp.com/attachments/1525229638115135561/1525229638429966456/image.png?ex=6a52a04b&is=6a514ecb&hm=6c494941423f5ad6580deb5c60a2cc6e6a2c6377776e983bddd8e239840b0215&

## Additional context
Reporter notes this may affect versions other than Staging. Maintainer indicated this is a recurring regression that has been fixed multiple times before.

Discord thread: [Bug] Escaping <> mangles standards <START> example dialog format
Discord thread link: https://discord.com/channels/1417099416812392641/1525229638115135561
Requested by: Marinara (marinara_spaghetti, Discord ID 169561011928432653)

<details>
<summary>Discord thread transcript (4 messages)</summary>

```text
[2026-07-10T19:58:03.123Z] Kat: Version: Staging, may affect other versions

Standard `<START>` marker in example dialog gets escaped to `&lt;START>`

[attached image 1: https://cdn.discordapp.com/attachments/1525229638115135561/1525229638429966456/image.png?ex=6a52a04b&is=6a514ecb&hm=6c494941423f5ad6580deb5c60a2cc6e6a2c6377776e983bddd8e239840b0215&]
[Discord message: https://discord.com/channels/1417099416812392641/1525229638115135561/1525229638115135561]

[2026-07-10T20:04:31.488Z] Marinara: Wtf, I already fixed this.
[Discord message: https://discord.com/channels/1417099416812392641/1525229638115135561/1525231267036266567]

[...]

Issue #3206, opened 2026-07-04
[Issue]: html sanitizing converting < to &lt;


Winterfell8840:
## Summary

Html coding constabtly being overly sanitized, even for font color
## Expected behavior
Display roleplay text appropriatly

Showing regular text encoded property 
## Actual behavior
Html being converted to &lt;, the response showing the coding 

## Steps to reproduce

1. Regenerate response
2.new response
3.any responses with html

## Environment

- Marinara Engine version: 2.0.o
- Install type: (source / installer / docker / termux / other) termuz
- OS + version:
- Browser or app shell:

## Template check

Please **uncheck (untick)** the box below before submitting so we know you read the template. It is intentionally pre-checked:

SpicyMarinara:
Addressed in ready PR #3218 targeting staging.

Issue #3108, opened 2026-06-30
[Issue]: In XML wrap format, user-typed < and > are escaped to &lt;/&gt; before the model sees them (breaks blockquotes, wastes tokens)


kolacheee:
**What's wrong**

When the prompt **wrap format is set to XML**, any `<`, `>`, and `&` you typed yourself — in a message, a lorebook entry, a character field, a summary, etc. — is converted to HTML entities (`&lt;`, `&gt;`, `&amp;`) *before the model sees it*. So a lorebook heading `<Setting>` is sent as `&lt;Setting&gt;`, and a markdown blockquote `> ...` is sent as `&gt; ...`. Marinara's own structural tags (`<lorebook_all>`, `<last_message>`, …) are **not** escaped, which is why those look fine while your typed text doesn't.

This has two real costs the reporting user called out:
- **It confuses the model.** The model receives literal `&gt;` / `&lt;Setting&gt;` rather than `>` / `<Setting>`. (Confirmed live: a user asked the model to echo their message back verbatim and the reply contained `&gt;`, so the model genuinely received the entity, not the character.) Markdown blockquotes (`>`), a staple of roleplay formatting, are broken this way on every turn.
- **It wastes tokens.** `&gt;` / `&lt;` tokenize to several tokens each versus one for `>` / `<`, so heavy `<...>`/blockquote users pay an ongoing token tax.

The biggest single offender is `>`: **XML does not require `>` to be escaped in text content at all** (only `<` and `&` are mandatory; `>` only needs escaping inside the literal `]]>` sequence). So escaping every `>` is unnecessary *and* it's the part that breaks the very common markdown blockquote.

**How to reproduce (try it yourself)**

1. Use a connection/preset whose **prompt wrap format is "XML"** (the assembled prompt shows `<lorebook_all>`, `<last_message>` style tags).
2. Either: add a lorebook entry whose text contains `<Setting>`, **or** send a chat message that starts with a markdown blockquote, e.g. `> hello`.
3. Open **Peek Prompt** (or have the model echo your message back verbatim).

**You'll see:** the typed `<Setting>` appears as `&lt;Setting&gt;`, and the `>` appears as `&gt;` — and an echo test shows the model received `&gt;`, not `>`.
**Expected:** your `<`/`>` reach the model as the characters you typed (at minimum `>` should never be escaped), so blockquotes work and no extra tokens are spent. Marinara's own wrapper tags stay literal as today.

[...]

PR #3765, opened 2026-07-18
fix(prompt): pass prompt leaf content to the model verbatim (stop escaping angle brackets)

### LINKED ISSUE #3766: Angle brackets in cards, personas, and lorebooks get mangled to &lt; before reaching the model
If you type an angle bracket into a character card field, a persona/user description, a lorebook entry, or example dialogue, the model doesn't get what you typed. A `<` is turned into `&lt;` and a `&` into `&amp;` before the prompt is sent.

That matters because a lot of people organize their cards and lorebooks with angle-bracket / HTML-style tags like `<thinking>`, `<scenario>`, or `<div>`, and some cards use inline HTML like `<font color=...>` for styling. All of that reaches the model garbled — it sees `&lt;thinking>` and `&lt;font color=...&gt;` — so tag-based structure and HTML formatting break in the reply. The prompt viewer also shows the literal `&lt;`, and the escaped entities waste tokens.

This happens whenever the Conversation preset uses the default XML wrap format, which is the out-of-the-box setup.

## Try it yourself (on released v2.3.3)

1. Open any persona (user) and put `<test>` in its description. (A character card field or a lorebook entry works too.)
2. Open or start a chat that uses the default XML-wrapped Conversation preset.
3. Open the Peek Prompt / Assembled Prompt view.
4. You'll see `&lt;test>` where you typed `<test>`. That's exactly what the model receives.

This affects "leaf" content generally: character card fields (description, personality, scenario, etc.), the persona / user description, lorebook entries, recalled memories, scene text, and example dialogue.

## Why it happens (short note)

In `packages/server/src/services/prompt/prompt-escaping.ts`, the function `escapeXmlText` runs `value.replace(/&/g,'&amp;').replace(/</g,'&lt;')` on this leaf content and is applied to the XML-wrapped path via `sanitizePromptLeaf`. It was originally added as a prompt-injection guard against untrusted imported cards — but for a local single-user tool that "injection" is at most a rare, self-fixable annoyance (you see the weird output and edit your own card), while the escaping corrupts the very common, legitimate use of tags in cards and lorebooks. Leaf content should reach the model verbatim.

This exact behavior has been broken and re-fixed several times, because each pass only partially carved out the guard (dropping `>`, exempting the `<START>` example-dialogue marker, decoding on the display layer in #3206) instead of removing it — so the `<` escaping kept surviving.

## Not part of this bug

The agent value/attribute escapers that feed strictly machine-parsed XML — agent world-state, knowledge/entry catalogs, storyboard sections, attachment attributes — are correct to keep escaping and should stay as they are. This is only about the human-authored prompt leaf path.

## Fix

A fix is already open in #3765: it makes leaf content verbatim and hardens the code so this stops recurring — https://github.com/Pasta-Devs/Marinara-Engine/pull/3765

kolacheee:
## Linked issue

Closes #3766

## What this does

Prompt **leaf** content — character card fields, persona, lorebook entries, memories, scene text, and example dialogue — now reaches the model **verbatim**. Angle-bracket and HTML-style tags (`<thinking>`, `<scenario>`, `<div style="...">`) are passed through exactly as authored instead of arriving at the model as `&lt;thinking&gt;`.

This fixes a recurring user-visible bug: `<test>` typed in a persona or card showed up as `&lt;test>` in Peek Prompt / the assembled prompt, roleplay HTML formatting broke, the editor displayed raw `&lt;`, and tokens were wasted on entities.

The change itself is one line of behavior (the XML leaf path stops HTML-escaping). The bulk of this PR is **hardening** so it stops getting silently re-broken.

---

## Why this regression keeps happening

Leaf-content escaping has flip-flopped across many versions. The reason it keeps coming back is that **the guard was never actually deleted before this branch** — every prior "fix" only carved out an *exception* rather than removing the escaping:

- the escaper was born escaping all of `&`, `<`, `>`;
- one change dropped only `>`, a later one re-added `>`, another dropped it again;
- `<START>` example-dialogue carve-outs were added (#3441, #3623);
- a display-layer band-aid decoded `&lt;font…&gt;` before rendering (#3206) so the escaping merely *looked* harmless while the model still received the mangled `&lt;`.

The `<` escaping survived all of that. Because the underlying injection-guard framing was never removed, three renewable sources kept regenerating the belief that escaping is correct and kept re-attracting it:

1. **The function name.** `escapeXmlText` reads as "an XML escaper," so making its body escape again looks like a bug fix, not a regression. (There is even a second, unrelated local `escapeXmlText` in `knowledge-router.ts` that legitimately *does* escape — an identical name with opposite behavior.)
2. **The CHANGELOG.** It still declares `<`/`&` escaping deliberate "prompt-boundary safety" (#3108) and "Hardened prompt assembly … by escaping untrusted character/persona card text," with no countervailing entry. Release notes are the first thing a contributor reads, and they argued *for* escaping.
3. **No recorded decision.** `CONTRIBUTING.md` never mentioned this; the only rationale lived in a file header a re-escape PR's author might never open.

This PR is the first *true* removal, and it closes all three sources (see Hardening below).

---

## Why verbatim is the correct final state

**Threat model.** Marinara Engine is local, single-user roleplay. The person who authors or imports a card is the same person the card would "attack." The worst realistic outcome of an unescaped `<system>`-looking tag is the model role-playing something odd — which the user sees immediately in their own chat and fixes by editing their own card. That is a rare, self-inflicted, self-correcting annoyance, not a security boundary. Escaping traded that non-threat for a *constant, real* harm: the very common, legitimate use of angle-bracket/HTML tags in cards and lorebooks arrived mangled.

**The model is not an XML parser.** Escaping treated the LLM as if a stray `<` could "break out" of a structural element the way it would in a real XML document. An LLM reads tokens; there is no parser to break out of. Turning `<` into `&lt;` closes no injection surface — it only corrupts text.

**Structure is not at risk.** The framework's own section wrappers (`<description>…</description>`, `<last_message>…`, etc.) are emitted by `wrapContent` *around* leaf content, from fixed section names, independent of what the content contains. Verbatim leaf content therefore cannot alter the structural tags — it only changes what sits inside them, as intended.

**Honest residual limitation.** Verbatim does mean a self-authored/imported card can paste a convincing framework-style closing tag plus fake instructions into context. In a local single-user tool that is self-inflicted and self-fixable, and it is no stronger than writing those instructions in plain prose (which the model reads either way). If a future multi-tenant or shared-marketplace deployment ever changes the threat model, the right response is a validation/consent step at the **import boundary**, not silent per-token escaping of everyone's content at prompt-assembly time.

---

## What deliberately stays escaped (out of scope)

The agent value/attribute escapers are **unchanged** and must stay:

- `escapeXml` / `escapeXmlAttribute` in `agent-executor.ts`, the local escaper in `knowledge-router.ts`, and the storyboard/attachment escapers escape dynamic values into XML **attributes** and into strict, **machine-parsed** agent documents (world-state, entry catalogs, `<turn_sections>`, `<attached_file name="…">`). Those are parsed downstream, where a stray `"` or `<` genuinely breaks an attribute or element.

Note that `agent-executor.ts` also escapes some of the *same* card fields into element content of that parsed document. That is a different consumer with different rules — it is explicitly **not** a consistency argument for re-escaping the main prompt path (the SCOPE comment now says so).

---

## Hardening (so this stops recurring)

- **Renamed the identity function** `escapeXmlText` → `passThroughLeaf` (+ its callers in `scene-context-runtime.ts` and the regression). A function named "escapeXmlText" whose body is `return value` invites the one-line "obviously it should escape" change.
- **CHANGELOG [Unreleased] entry** marks the #3108 and "Hardened prompt assembly" escaping entries as **superseded** for the local single-user model, so the release notes no longer read as evidence that escaping is correct. (Historical entries left intact.)
- **CONTRIBUTING.md decision record** ("Prompt Leaf Content Is Verbatim") states the invariant, the threat model, and what stays escaped — something a reviewer can cite when rejecting a re-add.
- **Assembled-prompt canary test** feeds a persona **and** a character description containing `<test>` + inline HTML through the real assembly path and asserts they arrive verbatim with no `&lt;`/`&amp;`/`&gt;` anywhere. The prior lock test only proved the leaf helper is identity, not the whole path.
- **Renamed the misleading** "identity fallback escapes … delimiters" test to "… passes … verbatim" and extended the file-header SCOPE comment.

---

## Validation

- `pnpm check` (TypeScript + ESLint + build): passes.
- `pnpm regression:prompt`: passes for all prompt-escaping tests, including the new canary and the renamed verbatim tests. (One unrelated, pre-existing failure exists on this tree — the "Professor Mari … public reference … Prose Guardian" agent-catalog completeness check — and reproduces without any change in this PR.)
- `pnpm --filter @marinara-engine/server exec tsx ../../scripts/regressions/open-issues.regression.ts`: passes.

## Manual verification (please confirm)

- [x] In Peek Prompt / Assembled Prompt, a persona description containing `<test>` shows `<test>` verbatim (not `&lt;test>`).
- [x] A character card / lorebook entry using `<thinking>` or inline HTML reaches the model and renders as authored (roleplay HTML no longer breaks).
- [x] Example dialogue `<START>` separators still work, including cards imported with the marker already entity-encoded.
- [x] Agent-driven features that emit machine-parsed XML (world-state trackers, knowledge/entry catalogs, storyboard sections) still parse correctly — those paths intentionally remain escaped.

## Notes

- Targets `staging` per the contributor workflow.
- Tracked by #3766 (from a direct user report with screenshots). Because this PR targets `staging` rather than the default branch, the `Fixes` keyword won't auto-close #3766 on merge — the `in-pr` → `fixed-in-staging` label flow covers closure when staging merges to main.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prompt content now preserves angle brackets and inline HTML verbatim, including character and persona details, lorebook entries, memories, scene text, and chat context.
  * Tag-like content such as `<thinking>` and `<scenario>` is no longer converted into escaped entities like `&lt;...&gt;`.
  * Example dialogue markers continue to be recognized correctly while surrounding content remains unchanged.


kolacheee:
@coderabbitai review

PR #3835, opened 2026-07-19
Fix assigned image, agent, and package lifecycle issues

### LINKED ISSUE #3828: [Issue]: ComfyUI Engine loop skips indexed reference placeholders — Tauri fix from PR #1319 was never ported to packages/server
## Summary

`packages/server/src/services/image/image-generation.ts` still exhibits the exact bug described in #1275 — indexed ComfyUI reference placeholders (`%reference_image_name_02%`, `%reference_image_name_03%`, …) are never substituted when the scene has fewer reference images than the workflow declares slots for. They leak to ComfyUI as literal text and the queue rejects the workflow with `Invalid image file: %reference_image_name_02%` (or similar).

That bug was already fixed on the **Tauri side** in PR #1319 (`src-tauri/src/commands/storage/images/providers.rs` — `collect_comfy_reference_name_tokens` + missing-slot placeholder padding) and #1275 was closed accordingly. However the fix was **never mirrored to the Marinara Engine's Node/TypeScript codebase** (`packages/server/src/services/image/image-generation.ts` and `runpod-comfyui.service.ts`), where the same loop still stops at `references.length` without padding remaining indexed placeholders.

## Expected behavior

For any ComfyUI workflow that uses indexed reference placeholders, every `%reference_image_NN%` / `%reference_image_name_NN%` should be substituted:
- Slots 1..N where refs are available → filled with the corresponding uploaded reference.
- Slots N+1..COMFYUI_MAX_REFERENCE_IMAGES that the workflow still references → filled with the built-in `COMFYUI_PLACEHOLDER_REFERENCE_BASE64` (subject to `defaults.uploadPlaceholderOnMissingReference`).

Matches the Tauri behavior from PR #1319.

## Actual behavior

Current `image-generation.ts` (Engine v2.3.3, `staging` branch as of $(date -u +%Y-%m-%d)):

\`\`\`ts
const references = collectComfyReferenceImages(request, defaults);
for (let i = 0; i < references.length; i++) {
  // fills only slots 1..references.length
}
const resolvedWorkflow = replaceComfyUiPlaceholders(workflow, replacements);
\`\`\`

The loop is bound by `references.length`. If the workflow uses `%reference_image_name_02%` and the scene supplies 1 reference, that placeholder stays literal and ComfyUI rejects the payload.

`runpod-comfyui.service.ts` has the same shape (line ~111) and the same defect.

`collectComfyReferenceImages`'s placeholder fallback (line ~2422) fires only when `references.length === 0` and then only fills slot 1 — it never pads slots 2..N.

## Steps to reproduce

1. Configure a ComfyUI connection in Engine v2.3.3 with a workflow that declares three `LoadImage` nodes using `%reference_image_name_01%`, `_02%`, `_03%`.
2. In a chat with one active character, run the Illustrator `[selfie]` command (only one reference image is available).
3. ComfyUI queue call returns `400 prompt_outputs_failed_validation` with `details: "image - Invalid image file: %reference_image_name_02%"`.

## Prior art

- Original bug: #1275 (Beholder, May 2026)
- Rust/Tauri fix: PR #1319 (`cha1latte`, merged) — `src-tauri/src/commands/storage/images/providers.rs`
- Closely related: #3769 (1x1 placeholder crashing ComfyUI `fillborders` — separate concern; that fix has its own PR pending)

## Suggested fix

Port the Tauri logic to the Engine's Node/TS side. Concretely, after the existing `for (let i = 0; i < references.length; i++)` loop in `generateComfyUI`:

- Iterate `i` from `references.length` up to `COMFYUI_MAX_REFERENCE_IMAGES`.
- For each `i`, if `workflowJson.includes("%reference_image_${padded}%")` set `replacements[...] = COMFYUI_PLACEHOLDER_REFERENCE_BASE64`.
- For each `i`, if `workflowJson.includes("%reference_image_name_${padded}%")`, upload the placeholder image **once** (memoize) via `uploadComfyReferenceImage` and use the returned filename for every unfilled name slot.
- Gate the whole padding block on `defaults.uploadPlaceholderOnMissingReference` so the opt-out semantics stay consistent.

Mirror the same in `runpod-comfyui.service.ts` (string-substitution flavor, no separate upload step).

Happy to author the port if useful — flagging first so we don't double-work if a mirror-port PR is already in flight.

## Environment

- Marinara Engine version: 2.3.3 (and current `staging` HEAD)
- Install type: local dev (pnpm)
- OS + version: Linux (CachyOS, kernel 7.0.10)
- Backend: ComfyUI-ImageEditing on port 8189, Qwen-Image-Edit workflow with 3 LoadImage slots

### LINKED ISSUE #3829: [Issue]: Reject turn-game packages without a server entrypoint before installation
## Summary

The shared capability manifest contract accepts a package whose `kind` includes `turn-game` while `entrypoints.server` is absent. Installation persists that package as active/ready before the route attempts mandatory server activation, so activation fails after mutation and leaves an inconsistent installed record.

## Expected behavior

A turn-game package without a server entrypoint should be rejected before its files or installed-registry entry are mutated.

## Actual behavior

1. `capabilityPackageManifestSchema` permits `kind: ["turn-game"]` with no `entrypoints.server`.
2. `installCatalogPackage()` persists it with `status: "active"` and `readiness: "ready"` because there is no server entrypoint.
3. The install route then calls `capabilityModuleRuntime.activatePackage()` for every turn-game package.
4. `runtimePackages()` excludes the package because it has no server entrypoint, so activation throws `Installed capability package <id> has no server runtime` before status/error handling or rollback runs.
5. The route's `finally` refreshes the agent registry but does not remove or mark the installed record as failed.

Relevant source at audited `staging` revision:

- [Manifest fields are independently optional](https://github.com/Pasta-Devs/Marinara-Engine/blob/e2aaec9bf327a686f4471deac96c1076632994d2/packages/shared/src/schemas/capability-package.schema.ts#L22-L30)
- [Installation persists active/ready state](https://github.com/Pasta-Devs/Marinara-Engine/blob/e2aaec9bf327a686f4471deac96c1076632994d2/packages/server/src/services/capability-packages/package-manager.service.ts#L288-L301)
- [Turn-game installation always requests activation](https://github.com/Pasta-Devs/Marinara-Engine/blob/e2aaec9bf327a686f4471deac96c1076632994d2/packages/server/src/routes/capability-packages.routes.ts#L27-L37)
- [Activation rejects packages missing a server runtime](https://github.com/Pasta-Devs/Marinara-Engine/blob/e2aaec9bf327a686f4471deac96c1076632994d2/packages/server/src/services/capability-packages/capability-module-runtime.service.ts#L200-L208)

## Steps to reproduce

1. Create a valid capability-package fixture with `kind: ["turn-game"]`, a client entrypoint/contribution, and no server entrypoint.
2. Parse it with `capabilityPackageManifestSchema`; validation succeeds.
3. Install it through the capability-package route using a catalog fixture.
4. Observe the install request fails with `has no server runtime`.
5. Read `/api/capability-packages/installed`; the package remains present as active/ready.

## Environment

- Marinara Engine version: 2.3.3 (`staging` revision `e2aaec9bf327a686f4471deac96c1076632994d2`)
- Install type: source audit
- OS + version: Not platform-specific; this is a source-level contract issue
- Browser or app shell: Not applicable

A direct schema/readiness check produced:

```json
{"parsedKind":["turn-game"],"server":null,"consideredReady":true}
```

The complete catalog-download route was not executed during the audit; the persistence and activation sequence above was verified statically.

## Additional context

The package documentation states that turn-game installation registers its server engine immediately, so a server entrypoint is a runtime invariant.

The smallest safe fix is an install-time/catalog invariant check before artifact or registry mutation. Adding the rule directly to the base manifest schema needs compatibility consideration because the same schema parses existing installed-registry records; an already persisted malformed record should not prevent startup without a migration path.

## Template check

### LINKED ISSUE #3831: [Issue]: Agent-default changes (prompt template, per-agent toggles) don't propagate to existing chats — chatMeta overrides pin forever
## Summary

When a user changes an agent's default (e.g. the Illustrator agent's default prompt template, or its default Attach-Card-Appearance / Send-Avatar-References toggles), **new chats** correctly pick up the new default, but **existing chats** continue to use the old value indefinitely. There is no UI affordance to inherit-from-default; the only workaround is opening every existing chat and manually re-selecting the current default.

Confirmed by inspection: `chatMeta` is authoritative once it holds any value for a given field, and it's written on the user's first interaction with that field (even implicit interaction, in some cases). After that, the "default" the user changes in agent settings applies only to future chats.

## Expected behavior

A change to an agent's default should be reflected in all chats that have not explicitly overridden that field. Concretely, one of:
1. Existing chats without an explicit user pick track the current default (sentinel-based / null-means-inherit).
2. Users can "reset to agent default" per chat via a small UI control.
3. A one-click "push current defaults to all existing chats" bulk action from the agent editor.

## Actual behavior

Pattern in the server code (Node/TS Engine):

`packages/server/src/routes/generate/retry-agents-route.ts:3057-3063` (mirror at `generate.routes.ts:7872`):

\`\`\`ts
const useAvatarRefs =
  typeof chatMeta.illustratorUseAvatarReferences === "boolean"
    ? chatMeta.illustratorUseAvatarReferences
    : illustratorAgent?.resolved.settings?.useAvatarReferences === true;
\`\`\`

Once `chatMeta.illustratorUseAvatarReferences` is set (`true` or `false`), the ternary always resolves to `chatMeta`, and the agent's default is never consulted again. Same shape for `illustratorIncludeCharacterAppearance` and `illustratorPromptConnectionId`.

For prompt template selection, `chatMeta.agentPromptTemplateIds` is a `Record<agentType, templateId>`:

`packages/server/src/routes/generate/retry-agents-route.ts:1046`:

\`\`\`ts
const agentPromptTemplateSelections = normalizeAgentPromptTemplateSelectionMap(chatMeta.agentPromptTemplateIds);
// … later …
selectedPromptTemplateId: agentPromptTemplateSelections[cfg.type as string] ?? null,
\`\`\`

Client-side fallback at `packages/client/src/components/chat/ChatSettingsDrawer.tsx:3821` also picks a stable string:

\`\`\`ts
selectedId={agentPromptTemplateSelections[agent.id] ?? getDefaultPromptTemplateIdForAgent(agent.id)}
\`\`\`

Whichever template ID lands in `chatMeta.agentPromptTemplateIds[agentType]` sticks. Changing `defaultPromptTemplateId` on the agent afterward has no effect on that chat.

Same-shape pattern is used across other per-chat agent overrides (Music DJ, Storyboard, Video, etc.).

## Steps to reproduce

1. Attach the Illustrator agent to two chats. Let both save their per-chat metadata (open the agent settings drawer to trigger it).
2. In the Illustrator agent editor, change the default prompt template selection from Template A to Template B.
3. Start a **new** chat and attach the Illustrator agent → Template B is used ✓.
4. Return to either existing chat → still uses Template A. Same result if you change the default value of Attach-Card-Appearance or Send-Avatar-References.

## Suggested fix (pick one)

1. **Sentinel-based reset (minimal change).** Treat `chatMeta.X === null` (or missing) as "inherit from agent default." Add a small "Reset to agent default" affordance per field in the chat's agent settings that writes `null`. Chats that don't have the field set at all keep working; users can opt back into inheritance any time.
2. **Bulk-migration action (matches the reporter's actual pain most directly).** From the agent editor: "Push current defaults to all existing chats using this agent." Explicit, destructive, single click.
3. **Effective-value + override badge (aligned UI).** Show the effective value in the chat drawer with a small "inherited" / "overridden" tag and a per-field revert control. Matches the code's real semantics and makes the model discoverable.

Options 1 and 3 are compatible; 2 stands alone as a maintenance action.

## Test plan (leave boxes unchecked until a human verifies)

- [ ] With two existing chats using the Illustrator agent, change the agent's default prompt template. A new chat picks up the new default. Existing chats behave per the chosen fix (inherit / show badge / receive the bulk push).
- [ ] Same for Attach-Card-Appearance toggle.
- [ ] Same for Send-Avatar-References toggle.
- [ ] Same for Illustrator prompt connection selection.
- [ ] Verify parallel behavior for Music DJ, Storyboard, Video agent per-chat overrides (same code shape, presumably same fix).

## Impact

Users who tune agent defaults after building up a chat backlog have to manually walk every chat to apply the change. Scales poorly, easy to miss chats, invisible to the user which chats are stale.

## Environment

- Marinara Engine version: 2.3.3 (and current `staging` HEAD)
- Install type: local dev (pnpm)
- OS + version: Linux (CachyOS)

### LINKED ISSUE #3832: [Issue]: Package uninstall leaves secondary agent state for multi-agent packages
## Summary

Package uninstall cleanup removes chat selections and stored agent configuration using only the package ID. The documented capability contract allows one package to contribute multiple agent definitions, so any secondary agent IDs remain as stale chat references and orphaned configuration after uninstall.

No current official package is known to trigger this: all 29 packages in the current Marinara-Agents catalog contain one same-named agent. This is a latent contract bug rather than an active production regression.

## Expected behavior

Uninstalling a package should remove every agent definition supplied by that package from active chat selections and delete each corresponding agent configuration/runtime record, while preserving historical chat content as documented.

## Actual behavior

1. A package may contain an array of agent definitions, and the runtime registry registers every `definition.id`.
2. Chats store those definition IDs in `metadata.activeAgentIds`; configurations are stored by definition type.
3. The uninstall route calls `capabilityPackageManager.uninstall(id)`, which removes the package registry entry and deletes its version directory.
4. Cleanup then filters `activeAgentIds` and deletes a configuration using only the package ID.
5. Secondary definition IDs therefore remain in chat metadata, and their configurations/memory remain stored even though their runtime definitions are gone.

Relevant source at the audited `staging` revision:

- [The package architecture permits one or more declarative agents](https://github.com/Pasta-Devs/Marinara-Engine/blob/e2aaec9bf327a686f4471deac96c1076632994d2/docs/development/optional-agent-packages.md#L9-L18)
- [The shared schema accepts an array of agent definitions](https://github.com/Pasta-Devs/Marinara-Engine/blob/e2aaec9bf327a686f4471deac96c1076632994d2/packages/shared/src/schemas/capability-package.schema.ts#L143-L164)
- [Runtime registration uses every definition ID](https://github.com/Pasta-Devs/Marinara-Engine/blob/e2aaec9bf327a686f4471deac96c1076632994d2/packages/server/src/services/capability-packages/package-manager.service.ts#L362-L378)
- [Uninstall deletes package files before route-level cleanup](https://github.com/Pasta-Devs/Marinara-Engine/blob/e2aaec9bf327a686f4471deac96c1076632994d2/packages/server/src/services/capability-packages/package-manager.service.ts#L529-L538)
- [Route cleanup uses only the package ID](https://github.com/Pasta-Devs/Marinara-Engine/blob/e2aaec9bf327a686f4471deac96c1076632994d2/packages/server/src/routes/capability-packages.routes.ts#L39-L65)

## Steps to reproduce

1. Create a valid catalog fixture whose package ID is `agent-suite` and whose `agents.json` defines both `agent-suite` and `suite-helper`.
2. Install the package, create/configure both agents, and add both definition IDs to a chat's `activeAgentIds`.
3. Uninstall `agent-suite` through `/api/capability-packages/agent-suite`.
4. Inspect the chat metadata and agent configuration storage.
5. Observe that `agent-suite` is removed, while `suite-helper` remains referenced/configured without a live package definition.

## Environment

- Marinara Engine version: 2.3.3 (`staging` revision `e2aaec9bf327a686f4471deac96c1076632994d2`)
- Install type: source audit
- OS + version: Not platform-specific; this is a package-lifecycle contract issue
- Browser or app shell: Not applicable

A direct schema/filter check produced:

```json
{"packageId":"agent-suite","definitionIds":["agent-suite","suite-helper"],"remainingAfterCurrentFilter":["suite-helper"]}
```

A complete database-backed uninstall was not executed during the audit; the storage and route sequence was verified statically.

## Additional context

At [Marinara-Agents commit `51eed186`](https://github.com/Pasta-Devs/Marinara-Agents/commit/51eed1860dd549f515e30857fb75434782a34a53), all 29 official `agents.json` files contain exactly one agent whose ID matches its package ID. The catalog validator requires a matching primary ID but permits additional definitions, so a future supported multi-agent package would expose this path.

The smallest fix is to resolve and retain all package-owned definition IDs before deleting the package version directory, then clean chat selections and configurations for that full set. Uninstall should still succeed with a safe package-ID fallback if the definition file is missing or corrupt.

Adjacent agent-keyed metadata such as prompt-template selections and overrides should be considered when defining the cleanup set, without deleting historical chat messages or completed feature data.

## Template check

SpicyMarinara:
## Linked issues

Closes #3828
Closes #3829
Closes #3831
Closes #3832

## Why this change

The assigned sweep exposed four host-integration gaps: indexed ComfyUI reference slots could leak into workflows, invalid turn-game packages could mutate install state before activation failed, existing chats had no clear way back to changing agent defaults, and multi-agent package removal cleaned only the primary package ID.

## What changed

- Pad only workflow-declared missing ComfyUI reference slots with the existing placeholder image, for local and RunPod workflows, while preserving the opt-out setting.
- Reject turn-game packages without a server entrypoint before registry or artifact mutation.
- Keep newly attached Illustrator settings inherited when they match the agent defaults, and show compact inherited/override status plus a reset action for prompt templates, the Illustrator prompt model, and Illustrator reference toggles.
- Resolve all agent definitions owned by a package before uninstall, then remove their chat selections, agent-keyed metadata, configurations, runs, and memory using the existing storage cleanup path.
- Add focused regression coverage for each lifecycle edge case.

## Validation performed

- `pnpm check`
- `pnpm regression:issues`
- `pnpm smoke:ui` — 72 passed, 46 intentional viewport skips
- `git diff --check`

## Validation checklist

- [ ] Relevant regression suites pass locally
- [ ] Ran affected UI paths manually
- [ ] Checked relevant error and compatibility paths

## Manual verification requested

- Manually verify a local ComfyUI workflow with one supplied reference and multiple indexed filename slots.
- Manually verify an existing Illustrator chat can reset its prompt template, prompt model, and reference toggles to the current agent defaults.
- Manually verify uninstalling a fixture package with two agent definitions removes both from chat settings and stored agent state.

## Summary by CodeRabbit

- **New Features**
  - Chat settings now show when Illustrator options and prompt templates override agent defaults.
  - Added “Use agent default” reset controls for Illustrator character appearance and avatar reference toggles.
  - Improved ComfyUI reference-image handling, automatically filling missing reference slots (including names).

- **Bug Fixes**
  - Capability package uninstall/removal now cleans up all associated agents and updates affected chat metadata accordingly.
  - Invalid “turn-game” packages missing a server entrypoint are rejected before installation proceeds.
  - Illustrator default vs override settings are preserved correctly; ComfyUI placeholder backfill now reuses uploaded placeholder filenames and uses safer timeout parsing.

### DIFF
--- a/packages/client/src/components/chat/AgentAddSetupFields.tsx
+++ b/packages/client/src/components/chat/AgentAddSetupFields.tsx
@@ -464,16 +464,24 @@ export function buildAgentAddMetadataPatch(
   }
   if (agentId === "illustrator") {
     const defaults = options?.illustratorDefaults;
-    if (!defaults || setup.includeCharacterAppearance !== defaults.includeCharacterAppearance) {
-      patch.illustratorIncludeCharacterAppearance = setup.includeCharacterAppearance;
-    } else if (hasOwn(metadata, "illustratorIncludeCharacterAppearance")) {
-      patch.illustratorIncludeCharacterAppearance = null;
-    }
-    if (!defaults || setup.useAvatarReferences !== defaults.useAvatarReferences) {
-      patch.illustratorUseAvatarReferences = setup.useAvatarReferences;
-    } else if (hasOwn(metadata, "illustratorUseAvatarReferences")) {
-      patch.illustratorUseAvatarReferences = null;
-    }
+    const applyIllustratorDefault = (
+      key: "illustratorIncludeCharacterAppearance" | "illustratorUseAvatarReferences",
+      value: boolean,
+      defaultValue: boolean | undefined,
+    ) => {
+      if (defaultValue === undefined || value !== defaultValue) patch[key] = value;
+      else if (hasOwn(metadata, key)) patch[key] = null;
+    };
+    applyIllustratorDefault(
+      "illustratorIncludeCharacterAppearance",
+      setup.includeCharacterAppearance,
+      defaults?.includeCharacterAppearance,
+    );
+    applyIllustratorDefault(
+      "illustratorUseAvatarReferences",
+      setup.useAvatarReferences,
+      defaults?.useAvatarReferences,
+    );
   }
   if (agentId === "haptic") {
     patch.enableHapticFeedback = setup.hapticFeedbackEnabled;
--- a/packages/client/src/components/chat/AgentAddSetupFields.tsx
+++ b/packages/client/src/components/chat/AgentAddSetupFields.tsx
@@ -394,7 +394,11 @@ export function buildAgentAddMetadataPatch(
   agentId: string,
   setup: AgentAddSetupState,
   metadata: Record<string, unknown>,
-  options?: { allowSecretPlot?: boolean; defaultPromptTemplateId?: string },
+  options?: {
+    allowSecretPlot?: boolean;
+    defaultPromptTemplateId?: string;
+    illustratorDefaults?: { includeCharacterAppearance: boolean; useAvatarReferences: boolean };
+  },
 ): Record<string, unknown> {
   const patch: Record<string, unknown> = {};
 
@@ -459,8 +463,17 @@ export function buildAgentAddMetadataPatch(
     patch.spotifyArtist = setup.spotifySourceType === "artist" ? setup.spotifyArtist.trim() || null : null;
   }
   if (agentId === "illustrator") {
-    patch.illustratorIncludeCharacterAppearance = setup.includeCharacterAppearance;
-    patch.illustratorUseAvatarReferences = setup.useAvatarReferences;
+    const defaults = options?.illustratorDefaults;
+    if (!defaults || setup.includeCharacterAppearance !== defaults.includeCharacterAppearance) {
+      patch.illustratorIncludeCharacterAppearance = setup.includeCharacterAppearance;
+    } else if (hasOwn(metadata, "illustratorIncludeCharacterAppearance")) {
+      patch.illustratorIncludeCharacterAppearance = null;
+    }
+    if (!defaults || setup.useAvatarReferences !== defaults.useAvatarReferences) {
+      patch.illustratorUseAvatarReferences = setup.useAvatarReferences;
+    } else if (hasOwn(metadata, "illustratorUseAvatarReferences")) {
+      patch.illustratorUseAvatarReferences = null;
+    }
   }
   if (agentId === "haptic") {
     patch.enableHapticFeedback = setup.hapticFeedbackEnabled;
--- a/packages/client/src/components/chat/ChatSettingsDrawer.tsx
+++ b/packages/client/src/components/chat/ChatSettingsDrawer.tsx
@@ -2121,14 +2121,14 @@ export function ChatSettingsDrawer({
     [chat.id, updateMeta],
   );
   const renderIllustratorPromptConnectionSelect = () => (
-    <label className="flex flex-col gap-1">
+    <div className="flex flex-col gap-1">
       <span className="text-[0.625rem] font-medium text-[var(--foreground)]">Prompt Model</span>
       <select
         value={illustratorPromptConnectionId}
         onChange={(event) => updateIllustratorPromptConnection(event.target.value)}
         className="w-full rounded-lg border border-[var(--border)] bg-[var(--background)] px-2.5 py-2 text-xs text-[var(--foreground)] outline-none transition-colors focus:border-[var(--primary)]/50"
       >
-        <option value="">Main chat model</option>
+        <option value="">Agent default</option>
         {selectedIllustratorPromptConnectionMissing && (
           <option value={illustratorPromptConnectionId}>Missing connection</option>
         )}
@@ -2143,7 +2143,11 @@ export function ChatSettingsDrawer({
         Chooses the text model that writes Illustrator/selfie prompts. Image rendering still uses the selected image
         connection for this feature or the agent setup.
       </span>
-    </label>
+      <AgentDefaultStatus
+        overridden={illustratorPromptConnectionId.length > 0}
+        onReset={() => updateIllustratorPromptConnection("")}
+      />
+    </div>
   );
   const toggleIllustratorCharacterAppearance = useCallback(() => {
     updateMeta.mutate({
@@ -2157,6 +2161,12 @@ export function ChatSettingsDrawer({
       illustratorUseAvatarReferences: !illustratorUseAvatarReferences,
     });
   }, [chat.id, illustratorUseAvatarReferences, updateMeta]);
+  const resetIllustratorCharacterAppearance = useCallback(() => {
+    updateMeta.mutate({ id: chat.id, illustratorIncludeCharacterAppearance: null });
+  }, [chat.id, updateMeta]);
+  const resetIllustratorAvatarReferences = useCallback(() => {
+    updateMeta.mutate({ id: chat.id, illustratorUseAvatarReferences: null });
+  }, [chat.id, updateMeta]);
   const proseGuardianBannedWords =
     typeof metadata.proseGuardianBannedWords === "string"
       ? metadata.proseGuardianBannedWords
@@ -3372,6 +3382,10 @@ export function ChatSettingsDrawer({
         ...buildAgentAddMetadataPatch(agent.id, setup, metadata, {
           allowSecretPlot: supportsNarrativeDirectorSecretPlot,
           defaultPromptTemplateId: resolveDefaultAgentPromptTemplateId(nextSettings),
+          illustratorDefaults: {
+            includeCharacterAppearance: nextSettings.includeCharacterAppearance === true,
+            useAvatarReferences: nextSettings.useAvatarReferences === true,
+          },
         }),
       });
       toast.success(`Added ${agent.name}! You can access its settings in Agents section in Chat Settings!`);
@@ -3938,6 +3952,7 @@ export function ChatSettingsDrawer({
                 <AgentPromptTemplateSelect
                   options={promptOptions}
                   selectedId={agentPromptTemplateSelections[agent.id] ?? getDefaultPromptTemplateIdForAgent(agent.id)}
+                  overridden={typeof agentPromptTemplateSelections[agent.id] === "string"}
                   onChange={(promptTemplateId) => updateAgentPromptTemplateSelection(agent.id, promptTemplateId)}
                 />
               </div>
@@ -7273,6 +7288,7 @@ export function ChatSettingsDrawer({
                               agentPromptTemplateSelections["echo-chamber"] ??
                               getDefaultPromptTemplateIdForAgent("echo-chamber")
                             }
+                            overridden={typeof agentPromptTemplateSelections["echo-chamber"] === "string"}
                             onChange={(promptTemplateId) =>
                               updateAgentPromptTemplateSelection("echo-chamber", promptTemplateId)
                             }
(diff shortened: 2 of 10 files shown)


SpicyMarinara:
@coderabbitai review

SpicyMarinara:
@coderabbitai review

Issue #2683, opened 2026-06-20
[Issue]: Agent connection resolution silently misroutes — deleted/empty-baseUrl falls back to main chat; empty model sends model=""


kolacheee:
## Summary

Agent connection resolution has two silent failure modes that misroute agent API calls: **(1)** a per-agent connection that was deleted, or has an empty/unresolvable base URL, silently falls back to the **main chat** connection/model with no warning; **(2)** a default-for-agents connection with an empty `model` is used anyway, sending agent requests with `model=""`.

## Expected behavior

- A per-agent connection that is deleted or unusable should produce a visible warning and either skip the agent or fall back to the *default-for-agents* connection (with the existing `default_agent_connection_active` warning) — not silently run on the user's main (often paid) chat connection/model.
- A connection with no model should be treated as unusable for agents, not dispatched with `model=""`.

## Actual behavior

**(1)** `resolveAgentConnectionProvider` loads the agent's connection via `connections.getWithKey(connectionId)`. Both a deleted connection (`getWithKey` returns null) and an empty resolved base URL fall through to `return { provider: fallbackProvider, model: fallbackModel, ... }` (the main chat provider/model) silently — the provider is only built inside an `if (agentBaseUrl)` guard, and the `default_agent_connection_active` warning is only emitted when the effective connection *is* the default-for-agents one. A dangling `connectionId` is trivial to produce (`agent.schema.ts` `connectionId` is `z.string().nullable()` with no existence validation).

**(2)** `getDefaultForAgents` returns the first non-image connection flagged `defaultForAgents` regardless of an empty `model` column (schema default `''`); `resolveAgentPipelineAgents` caches a provider with `model: defaultAgentConn.model` (no empty guard), and the executor passes `model=''` straight into `provider.chatComplete`.

## How it breaks the pipeline

Code trace (staging @ `b826f8f4`):

- (1) `packages/server/src/services/generation/agent-resolution.ts:130-155` (`resolveAgentConnectionProvider`; fallback at `:151-155`; provider built only inside `if (agentBaseUrl)` `:132-133`); warning gated at `:260-262`. `connections.getWithKey` null at `connections.storage.ts:25-29`. Empty baseUrl via `resolveBaseUrl` (`generate-route-utils.ts:769-778`). `connectionId` nullable at `agent.schema.ts:49`.
[...]

PR #3866, opened 2026-07-20
Sweep dangling connection references when a connection is deleted

### LINKED ISSUE #3865: Deleting a connection leaves dangling references across chat metadata and agent settings
## Summary

Deleting a connection (`DELETE /api/connections/:id`) removes only the connection row itself. Every stored reference to that id elsewhere is left untouched: `chats.connectionId`, `agentConfigs.connectionId`, `connections.embeddingConnectionId`, and the large family of `*ConnectionId` keys stored inside `chats.metadata` (`illustratorPromptConnectionId`, `gameImageConnectionId`, `gameSceneConnectionId`, `sceneVideoConnectionId`, `imageGenConnectionId`, `agentOverrides.<type>.imageConnectionId`, and others).

## Expected behavior

Deleting a connection should clear (null out) every stored reference to it, so no feature is left pointing at an id that no longer resolves to anything.

## Actual behavior

References survive the delete. What happens next depends on which code path reads the dangling id:

- Some paths handle it gracefully (this looks like the intent of #2683 / PR #2730 for the general per-agent connection path — confirmed present and working on current staging).
- `services/generation/illustrator-prompt-runtime.ts` (`resolveIllustratorPromptRuntime`, used by Gallery "Generate Selfie" and the Conversation `/selfie` command) throws synchronously when the id doesn't resolve — surfaced as a 400 to the caller, so at least fails fast.
- Other paths that read `chatMetadata.illustratorPromptConnectionId` (e.g. the `agentType === "illustrator"` special case in `agent-resolution.ts` / its near-duplicate in `retry-agents-route.ts`) do not fail fast in practice. Reproduced firsthand: a chat with a stale `illustratorPromptConnectionId` had one post-processing agent turn silently route through an unrelated, much larger local model instead of the intended (or default-for-agents) connection, taking over an hour instead of the usual ~3 minutes for that turn. Clearing the stale metadata field immediately fixed it; re-introducing the same stale id reproduced the multi-minute misroute again.

## Steps to reproduce

1. Create a connection, note its id.
2. Reference it from a chat: set it as the chat's main connection, and/or add it to chat metadata as `illustratorPromptConnectionId` or `gameImageConnectionId` (no dedicated UI control for these two currently exists; they're set via `PATCH /api/chats/:id/metadata`, but any chat that went through a setup wizard step touching these fields at the time can end up with one).
3. Delete the connection.
4. Observe every one of those references still holds the deleted id — `GET /api/chats/:id` still shows the stale ids in `connectionId` and `metadata`.
5. Depending on which agent/feature reads the stale id next turn, behavior ranges from a fast, clear error to an unbounded misroute (see above).

## Suggested fix

Sweep and null out every stored reference to a connection's id at delete time, in `connections.storage.ts`'s `remove()`. Given how many distinct `*ConnectionId` fields exist (15+ across chat metadata alone, per a repo-wide grep) and that new ones get added as features grow, a generic sweep — treat any JSON key matching `/ConnectionId$/` inside `chats.metadata` and `agentConfigs.settings`, at any nesting depth, as a reference and null it if it matches — seems more durable than hardcoding the current field list. Direct columns (`chats.connectionId`, `agentConfigs.connectionId`, `connections.embeddingConnectionId`) should be cleared the same way.

I have a working patch for this (generic recursive sweep, verified end-to-end against a running install with a connection referenced four different ways including a nested `agentOverrides.<type>.imageConnectionId`) and will open a PR against `staging`.

## Environment

- Marinara Engine version: staging @ current (confirmed PR #2730 / #2683 fix already present)
- Install type: source
- OS + version: Windows (code-level defect, likely platform-independent)

kh0p:
## Linked issue

Closes #3865

## Why this change

Deleting a connection leaves every stored reference to its id untouched — `chats.connectionId`, `agentConfigs.connectionId`, `connections.embeddingConnectionId`, and 15+ distinct `*ConnectionId` fields inside `chats.metadata` (`illustratorPromptConnectionId`, `gameImageConnectionId`, `gameSceneConnectionId`, `sceneVideoConnectionId`, `imageGenConnectionId`, `agentOverrides.<type>.imageConnectionId`, and more). #2683 / PR #2730 fixed the general per-agent connection resolution path to fail gracefully on a dangling id, but that fix didn't (and structurally couldn't, being a resolution-time guard) prevent the dangling reference from existing in the first place, and at least one other path (`illustratorPromptConnectionId` read directly in `agent-resolution.ts`'s illustrator special case and its near-duplicate in `retry-agents-route.ts`) doesn't fail gracefully at all — reproduced firsthand, a stale reference there silently misrouted a post-processing agent turn onto an unrelated, much larger model and turned a normal ~3 minute turn into over an hour.

Rather than adding another one-off guard to that specific path (and every future path that reads one of these fields), this removes the dangling reference at the source: the moment a connection is deleted.

## What changed

- Added `sweepDanglingConnectionReferences()` (`services/storage/connection-reference-cleanup.ts`), called from `connections.storage.ts`'s `remove()`.
- Clears `chats.connectionId` and `agentConfigs.connectionId` directly, and `connections.embeddingConnectionId` on any other connection.
- Sweeps `chats.metadata` and `agentConfigs.settings` recursively: any JSON key matching `/ConnectionId$/`, at any nesting depth, is treated as a connection reference and nulled if it matches the deleted id.
- Deliberately generic rather than a hardcoded field list — matches the naming convention already used consistently for every such field in this codebase, so it stays correct as new `*ConnectionId` settings get added without this file needing to track them.
- Logs a one-line summary (`chatsUpdated` / `agentsUpdated` / `connectionsUpdated`) when the sweep actually clears something, silent no-op otherwise.

## Validation

### Manual verification notes

This is a backend-only, non-UI change, so the container/UI-mode checklist items above don't apply (left unticked rather than claimed). What I did verify directly:

- `tsc --noEmit` and a full `tsc` build against a real, dependency-installed checkout — both pass with zero errors.
- End-to-end against a running instance: created a throwaway connection, referenced it four different ways on a chat (`chat.connectionId`, `metadata.illustratorPromptConnectionId`, `metadata.gameImageConnectionId`, and the nested `metadata.agentOverrides.custom-akistar-forge.imageConnectionId`), deleted the connection via the API, then re-fetched the chat and confirmed all four references were nulled. Repeated with a fresh connection/chat pair to rule out a fluke.
- Confirmed the pre-existing #2683/#2730 general-path fix is present and working on current staging before scoping this PR to the gap it left.

## Docs and release impact

- [x] No docs changes needed

N/A — backend-only change, no UI surface.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deleting a connection now automatically clears references to it across chats, agent configurations, and related connection settings.
  * Nested connection references in metadata and settings are safely removed, preventing stale or invalid configuration links.
  * Cleanup activity is tracked to improve visibility into affected records.

### DIFF
--- a/packages/server/src/services/storage/connection-reference-cleanup.ts
+++ b/packages/server/src/services/storage/connection-reference-cleanup.ts
@@ -29,6 +29,8 @@ import type { DB } from "../../db/connection.js";
 import { apiConnections, chats, agentConfigs } from "../../db/schema/index.js";
 import { now } from "../../utils/id-generator.js";
 
+const CLEANUP_BATCH_SIZE = 250;
+
 export type ConnectionReferenceCleanupResult = {
   chatsUpdated: number;
   agentsUpdated: number;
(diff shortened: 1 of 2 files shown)


SpicyMarinara:
Thank you for the help!

Xelvanis:
Concurrent connection deletion can restore dangling references

The bulk-delete UI sends deletions concurrently via `Promise.allSettled` (`ConnectionsPanel.tsx:1591`). Each `remove()` deletes its connection and runs the cleanup sweep without a transaction (`connections.storage.ts:436-439`).

The sweep reads chat rows, computes a complete replacement `metadata` string, then writes it back (`connection-reference-cleanup.ts:118-136`). Two concurrent sweeps can therefore read the same original metadata and overwrite each other:

 - Sweep A clears reference A.
 - Sweep B, using its earlier snapshot, clears B but writes A back.

 I reproduced this against the current PR head. Deleting A and B concurrently left:

 ```json
 { "firstConnectionId": "A", "secondConnectionId": null }
 ```

 Wrapping each connection deletion and its complete sweep in the existing `db.transaction(...)` produced both references as `null`.

 Please perform the connection delete and full cleanup inside one transaction, passing `tx` into `sweepDanglingConnectionReferences`. This also prevents a later cleanup failure from leaving the connection deleted with only partial reference cleanup.

PR #4001, opened 2026-07-23
(feat) Noodle:Slice 8b: access-protected generated creator images

Gunterlie:
## Summary
- Access-protected generated images for NoodleR creators (manual Guide and automatic-post paths): NoodleR-owned enablement and per-run quota on the `autoPosting` subtree, identity-protected image prompts, a private-media namespace served through an access-checked endpoint, a prompt-review lease path, and delete cleanup.
- Adds a per-creator manual "Run now" trigger under the automatic-posting scheduler, so the image path can be tested without waiting on the real cadence timer.
- Adds the global **Refresh NoodleR now** action from Slice 8's plan: runs every automation-enabled creator with bounded concurrency and independent per-creator outcomes (one creator's failure doesn't affect the others), prioritizing creators scheduled soonest. This closed a gap versus the living plan, which specified this action but only the per-creator route existed.
- Surfaces Bio/Stage voice as the explicit automatic-post guide from the Automation modal, and removes the "Max images per post" 2-4 picker since image generation only ever produces one image per post — those options never had any effect.

## Why
Slice 8b is the next tranche after text-only automatic posting (Slice 8): it lets automatically-posting creators include access-protected generated images the same way the manual Guide path can. The global refresh action and the dead settings option were cleanup found while extending this work, tracked against the same living plan.

## Test plan
- [ ] Manually verify image generation on the manual Guide path in the browser (draft prompt review, finalize, delete cleanup).
- [ ] Manually verify automatic-post image generation end-to-end with a real image provider connection (this PR's `pnpm check` + `regression:noodle` runs do not exercise a real provider).
- [ ] Manually verify the per-creator "Run now" trigger produces a post/image the same way a scheduled run would.
- [ ] Manually verify **Refresh NoodleR now** across multiple automation-enabled creators: concurrent runs, one creator with a provider/connection failure not blocking the others, and that each successful creator's `nextRunAt` advances under its own cadence.
- [ ] Manually verify the private-media endpoint enforces access control (non-subscriber/non-owner cannot fetch private images).
- [ ] Manually verify the Automation modal's new "Edit Bio & Stage voice" shortcut and updated copy.

No linked issue exists yet; one should be opened before this is taken out of draft.

## Summary by CodeRabbit

* **New Features**
  * NoodleR posts can now generate private images with optional prompt review before publishing.
  * Added controls to enable images, run automatic posting immediately, and refresh all eligible creators.
  * Added protected access to private post media.
* **Bug Fixes**
  * Improved cleanup of private media when posts or accounts are deleted.
  * Added safeguards for failed image generation and incomplete runs.
* **Tests**
  * Expanded regression coverage for image settings, prompt protection, scheduling, and private media access.


Gunterlie:
Thanks for the thorough review, @xelvanas — the threat-model framing was especially helpful for scoping these. All 7 behavioral findings and the 3 nitpicks are addressed in the latest push:

**Medium**
- **Atomic image-post commit** (`noodle-private-generation.service.ts`): the immediate path now derives the serving URL from a pre-generated post id and commits the image URL + media metadata in a single insert, with `promote()` inside the compensation boundary. Provider failure and persistence failure are distinguished, so a persistence error propagates instead of inserting a duplicate text fallback.
- **Slot consumed only on success** (`noodle-private-post.operation.ts`): global refresh now runs generation first and advances `nextRunAt` only when the result is `generated`. Busy/failed/skipped runs leave any explicit schedule edit untouched.
- **Guide review modal reachable** (`NoodlerHome.tsx`): the `ImagePromptReviewModal` is hoisted into a shared layer rendered by every branch that owns the Guide action, including the selected stage-profile view.

**Low**
- **Stale disclosure on reviewed images** (`noodle-private-images.service.ts`): the profile is re-read before finalize; if disclosure or the linked public identity changed during provider work, the staged image is discarded and the claim finalized as failed.
- **Dead `maxImagesPerRun`**: removed from schema, types, normalization, patch merge, client, and regressions. `imagesEnabled` is now the single one-image policy.
- **Contradictory pending state** (`finalizePostImageClaim`): the terminal transition now clears the `imagePendingReview` marker on both success and failure.
- **Global refresh feedback** (`use-noodle.ts` + `NoodlerHome.tsx`): outcome union moved into the shared contract; the UI now summarizes generated / skipped / failed with a reason when all or some creators fail.

**Nitpicks**
- Removed the unused `mediaPath` return field (callers use `metadata.privateMediaPath` + `stagedMedia`) and fixed the stale comment.
- Simplified per-creator Run-now to return a plain post (it never requests review) and dropped the unreachable client branch.
- `nextRunAtMs` now sorts the account objects already in hand instead of re-scanning by id.

`pnpm check` and `pnpm regression:noodle` pass. The manual test-plan items (real image provider, multi-creator refresh, access-control enforcement, etc.) remain unchecked per repo policy and still need in-browser verification.

Gunterlie:
@coderabbitai review

Gunterlie:
@coderabbitai Addressed in 612ec9636 (pushed).

**Fixed:**
- Living-plan Slice 8b entry (line 52): removed the obsolete per-run quota claim, now describes `imagesEnabled` as a single-image-per-post policy. Confirmed `noodle-private-generation.service.ts` has no per-run quota logic — one `imagePrompt` is proposed per post.
- `refreshAllNoodlerCreatorsNow` / client refresh flow: `"disabled"` and `"busy"` outcomes now map to `"skipped"` in `noodle-private-post.operation.ts` before being returned to the client, so a busy/kill-switched creator no longer inflates the "failed" toast count. `"generated"` and genuine failure statuses are unchanged, and slot consumption still only happens on `"generated"`.
- `imagesEnabled` checkbox: added the missing `onError` toast, matching the neighboring enabled/cadence controls.
- `confirmReviewedImagePrompts`: now reads the `finalized` count from the `confirmImagePrompts` response and shows a failure toast when `finalized === 0` instead of a false-positive success toast; the success path and review-state cleanup are unchanged.

**Skipped (nitpicks):**
- Async `existsSync` in `/noodler/posts/:id/media`: left as-is. Synchronous `existsSync` for a single local-file existence check ahead of a 404 is the established pattern used throughout this repo's route handlers (e.g. `gallery.routes.ts`, a dozen+ other route files), including for very similar "serve a file or 404" checks. Converting only this one call would create inconsistency without a measurable benefit.
- Provider timeout for `refreshAllNoodlerCreatorsNow`'s per-creator calls: already covered. Every outgoing LLM request in this codebase routes through `withLlmRequestTimeout`/the shared undici agent in `base-provider.ts`, which enforces headers/body timeouts at the provider boundary, so a hung call cannot stall the bounded-concurrency batch indefinitely.

`pnpm check` passes.

PR #4025, opened 2026-07-24
fix(agents): configurable AGENT_CALL_TIMEOUT_MS so slow local models aren't cancelled mid-stream (#3958)

### LINKED ISSUE #3958: [Issue]: Local LLM model's can time out if generation takes longer than 600 seconds, even when streaming.
## Summary
My local LLM's are 'failing' JSON parser validation when running as agents, despite generating valid JSON. Marinara Engine is canceling the task midway through and then rejecting the incomplete response.

## Expected behavior
Local endpoints should not be restricted by typical endpoint timeout constraints. Most of my model's run at 10-15 tk/s and can take a good 2-3 minutes to prompt process a long context window chat. This is expected behavior for a local instance and the timeout should be configurable within the connection settings.

## Actual behavior
LLama.cpp reports the task as 'canceled' and the web console from Marinara Engine shows a partial but not malformed JSON response:
```
26.36.453.098 I slot print_timing: id  0 | task 14020 | n_decoded =   4470, tg =  15.50 t/s, tg_3s =  15.44 t/s
26.39.484.757 I slot print_timing: id  0 | task 14020 | n_decoded =   4517, tg =  15.50 t/s, tg_3s =  15.50 t/s
26.42.524.759 I slot print_timing: id  0 | task 14020 | n_decoded =   4564, tg =  15.50 t/s, tg_3s =  15.46 t/s
26.42.847.579 W srv          stop: cancel task, id_task = 14020
26.42.911.950 I slot      release: id  0 | task 14020 | stop processing: n_tokens = 11819, truncated = 0
26.42.911.965 I srv  update_slots: all slots are idle
```
Response as listed in console (with the story data omitted):
Image
The final line of JSON shows it getting cut off midway through the generation:
```
{REDACTED STORY EDITS}
</span>\n</div>",
  "changes": [
    {
      "description": "Formatted the AL
```

## Steps to reproduce

1. Use a local LLM endpoint (IE llama.cpp).
2. Enable any agent which takes a long time to run (like the HTML reformatter).
3. Feed it a long task which will require at least 600s of generation time from llama.cpp for your particular model.

## Environment

- Marinara Engine version: V2.3.3
- Install type: Releases page ZIP
- OS + version: Windows 11
- Browser or app shell: Firefox

## Additional context
I have had this exact issue with a number of LLM tools at this point. Mainly Killocode and other VSCode based harnesses. The ideal solution for all of these in the past had been to add a configuration option for the timout to allow it to be disabled for local instances (and in some cases, add a defaults check which shuts it off automatically for known local address ranges: 127.0.0.1, 192.168.X.X and 10.10.X.X).

Furthermore, in reference to my previous issue from yesterday #3931 it seems this was actually the root cause. Of my local models:
* DeepSeek-V4-Flash: Works (Processing speed is fast enough to not hit the timeout most of the time).
* MiMo-V2.5: Fails
* Stepfun-3.7-Flash: Fails
* Hy3: Fails

After figuring out that the debug logs are in the browser and not the terminal, I checked and none of the failing models are responding with invalid JSON data, they are all getting cut off on the same manner as the logs I provided above.

(Also apologies if this *technically* counts as a feature request and not a bug. Wasn't sure which one to put it under.)

## Template check

- [] I DID NOT read this template and provide the requested details.

SpicyMarinara:
## Why

Fixes the root cause reported in #3958 (and its follow-up comment): local LLMs running as agents are cancelled mid-generation and then fail JSON validation on truncated output, even when actively streaming.

The reporter set `CHAT_GENERATION_TIMEOUT_MS=3600000` and still saw failures. That's because **agent LLM calls never consulted that setting.** Every agent call (trackers, HTML reformatter, etc.) was bounded by a hardcoded constant:

```ts
const DEFAULT_AGENT_CALL_TIMEOUT_MS = 5 * 60_000; // 300s, hardcoded
```

turned into `AbortSignal.timeout(300_000)` in `agentCallSignal()`. Two consequences:

1. It's a **total-duration** cap, not an idle cap — a healthy model streaming at 10–15 tk/s is still killed at exactly 300s. That's the "even when streaming" symptom and matches the llama.cpp `cancel task` log in the issue.
2. It ignores every configured timeout. Only the illustrator got a special 30-minute constant.

This also explains the reporter's "mysterious 600s second pass": when the first agent pass is cut at 300s, the truncated output fails JSON validation and the executor automatically retries once with a **fresh** 300s signal — so from the wall clock the second cancel lands at ~600s. It was 300 + 300, not a different timeout.

## What changed

- **`runtime-config.ts`**: new `getAgentCallTimeoutMs()` reading `AGENT_CALL_TIMEOUT_MS` (valid range 10000–3600000, default 300000), hot-reloaded per run and validated/warned exactly like `CHAT_GENERATION_TIMEOUT_MS`.
- **`agent-executor.ts`**: `agentCallSignal()` now uses the configured value; the illustrator keeps at least its built-in 30-minute budget via `Math.max`. The hardcoded `DEFAULT_AGENT_CALL_TIMEOUT_MS` constant is removed (the default now lives in runtime-config).
- **`.env.example`** and **`docs/CONFIGURATION.md`**: document the new knob next to `CHAT_GENERATION_TIMEOUT_MS`, noting it is a total-duration cap applied even while streaming.
- **`CHANGELOG.md`**: Unreleased → Added.

The automatic invalid-JSON retry behavior is unchanged — but with a correctly-sized budget the first pass should now complete, so the double-timeout stops happening.

## Test plan

- [ ] With a slow local endpoint (e.g. llama.cpp), enable an agent (HTML reformatter) and a task needing >5 min of generation; confirm it is cancelled at 300s with the default, and completes after setting `AGENT_CALL_TIMEOUT_MS=1800000`.
- [ ] Confirm an invalid `AGENT_CALL_TIMEOUT_MS` (e.g. `abc`, `5`, `99999999`) logs the range warning once and falls back to 300000.
- [ ] Confirm the value hot-reloads from `.env` without a server restart (like `CHAT_GENERATION_TIMEOUT_MS`).
- [ ] Confirm the Illustrator agent still gets at least its 30-minute budget.
- [ ] `pnpm check` passes. (It passed in the authoring session; re-verify locally.)

Closes #3958.

## Summary by CodeRabbit

* **New Features**
  * Added the configurable `AGENT_CALL_TIMEOUT_MS` setting for controlling the maximum duration of individual agent calls, including streaming.
  * Supports values from 10 seconds to 1 hour and applies to the next agent run without restarting.

* **Documentation**
  * Documented the new timeout setting in the example environment configuration, configuration reference, and changelog.

* **Bug Fixes**
  * Improved support for slow local models by replacing the previous fixed timeout with a configurable limit.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -48,6 +48,11 @@ EMBEDDING_TIMEOUT_MS=300000
 # Ordinary Conversation, Roleplay, and Game provider timeout in milliseconds.
 # Valid range: 10000-3600000. Changes apply to the next generation without a restart.
 CHAT_GENERATION_TIMEOUT_MS=300000
+# Total-duration cap for one agent LLM call (trackers, HTML reformatter, ...),
+# applied even while the response is streaming. Raise this for slow local
+# models (e.g. llama.cpp on modest hardware). Valid range: 10000-3600000.
+# Changes apply to the next agent run without a restart.
+AGENT_CALL_TIMEOUT_MS=300000
 # Tool-calling loops and custom tool execution.
 MAX_TOOL_ROUNDS=100
 CUSTOM_TOOL_TIMEOUT_MS=60000
--- a/.env.example
+++ b/.env.example
@@ -43,16 +43,16 @@ LOG_PRESET=default
 LOG_LEVEL=warn
 LOG_DISABLE_REQUEST_LOGGING=false
 
-# Slow local embedding servers may need longer lorebook vectorization requests.
-EMBEDDING_TIMEOUT_MS=300000
-# Ordinary Conversation, Roleplay, and Game provider timeout in milliseconds.
-# Valid range: 10000-3600000. Changes apply to the next generation without a restart.
-CHAT_GENERATION_TIMEOUT_MS=300000
 # Total-duration cap for one agent LLM call (trackers, HTML reformatter, ...),
 # applied even while the response is streaming. Raise this for slow local
 # models (e.g. llama.cpp on modest hardware). Valid range: 10000-3600000.
 # Changes apply to the next agent run without a restart.
 AGENT_CALL_TIMEOUT_MS=300000
+# Slow local embedding servers may need longer lorebook vectorization requests.
+EMBEDDING_TIMEOUT_MS=300000
+# Ordinary Conversation, Roleplay, and Game provider timeout in milliseconds.
+# Valid range: 10000-3600000. Changes apply to the next generation without a restart.
+CHAT_GENERATION_TIMEOUT_MS=300000
 # Tool-calling loops and custom tool execution.
 MAX_TOOL_ROUNDS=100
 CUSTOM_TOOL_TIMEOUT_MS=60000
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 
 ### Added
 
+- Added an `AGENT_CALL_TIMEOUT_MS` override for agent LLM calls (trackers, HTML reformatter, and other agents). The previous fixed 5-minute cap cancelled slow local models mid-stream and then burned a second 5 minutes on the automatic invalid-JSON retry; the cap is now configurable from 10 seconds to 1 hour and documented next to `CHAT_GENERATION_TIMEOUT_MS` (#3958).
+
 - Added safe host-rendered contribution slots to Browser Personal Extensions. `marinara.ui.registerContribution(...)` can add themed top-bar buttons, Extensions menu items, and right-side panels containing bounded text, actions, inputs, selects, toggles, sliders, and color controls. Marinara validates and renders every descriptor while activation and form events return only to the exact-hash-approved sandbox Worker; extensions never receive host DOM, markup, styling, React, network, or direct API authority. The existing constrained `marinara.ui.showWindow(...)` window supports the expanded controls as well.
 - Added Atlas Cloud as an image and video generation service, including curated starter models, text/reference image requests, asynchronous job polling, connection tests, and Game/scene-video routing (#3989).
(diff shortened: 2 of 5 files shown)


SpicyMarinara:
@coderabbitai review

SpicyMarinara:
Applied CodeRabbit's substantive findings in a4db67996:

- **agent-executor** (Major): each `provider.chatComplete` in the tool loop now gets a fresh `agentCallSignal()`, so `AGENT_CALL_TIMEOUT_MS` caps each LLM call rather than the whole loop — matching the documented "one agent LLM call" contract.
- **CONFIGURATION.md** (Minor): the reload-semantics/invalid-value paragraph now covers agent timeouts too.

Skipped the `.env.example` ordering nitpick: dotenv-linter isn't wired into this repo's CI, and the file is grouped by topic rather than alphabetized (`EMBEDDING_TIMEOUT_MS` already precedes `CHAT_GENERATION_TIMEOUT_MS`). I intentionally kept `AGENT_CALL_TIMEOUT_MS` next to `CHAT_GENERATION_TIMEOUT_MS` under their shared comment; moving it wouldn't make the section alphabetical anyway.

SpicyMarinara:
Addressed the remaining dotenv-linter nitpick in the latest commit: moved `AGENT_CALL_TIMEOUT_MS` to the **top** of the timeouts group. Note that CodeRabbit's literal suggestion (place it just before `CHAT_GENERATION_TIMEOUT_MS`) wouldn't have cleared the flag — the key would then be unordered against the preceding `EMBEDDING_TIMEOUT_MS`. `AGENT_CALL_TIMEOUT_MS` sorts alphabetically first among the keys in this block, so putting it at the group's head is what actually makes dotenv-linter clean without touching the pre-existing (topical, non-alphabetical) ordering of the rest of the file.

PR #4043, opened 2026-07-24
fix(cards): non-destructive version restore, correct timestamps, wrapping diffs (#4040)

### LINKED ISSUE #4040: [Issue]: Improve card history versioning and diff view
## Summary
Make card history behave more like Git commits by preserving every saved version, including the current version. Provide a detailed diff view for comparing card versions. EDIT: My forgetful ass forgot we already have version comparison. BUT, the current view doesn't properly wrap the text (very long words). In most cases this is not a problem but in creator notes, people seem to love putting a lot of fancy html which includes links.

## Expected behavior
Each saved card version should be preserved as a distinct history entry with its timestamp, and restoring an older version should not permanently lose newer versions. Users should also be able to view detailed differences between versions. Texts in the version comparisons should wrap.

## Actual behavior
The current card version is not saved until a newer version is created, so its history timestamp reflects the later save. Restoring an older version can permanently lose the newer version. Comparisons only identify changed sections, not the detailed text differences.

## Steps to reproduce
1. Start with a card at version v1.
2. Modify the card and save it as v2.
3. Open version history and observe that v1 appears with the timestamp of the v2 save.
4. Restore v1 and observe that v2 is no longer available in version history.
5. Compare two versions and observe that only changed sections are shown, without an inline detailed diff.

## Environment
- Marinara Engine version: v2.3.4
- Install type: Not provided
- OS + version: Not provided
- Browser or app shell: Not provided

Not provided

## Additional context
The detailed diff view could resemble a GitHub-style diff, with old and new content displayed side by side. Marinara acknowledged that the version-history concern makes sense; the diff UI was considered potentially out of scope but feasible.

Discord thread: [Feature Request] Card history enchantments
Discord thread link: https://discord.com/channels/1417099416812392641/1530175005013446756
Requested by: Marinara (marinara_spaghetti, Discord ID 169561011928432653)

<details>
<summary>Discord thread transcript (2 messages)</summary>

```text
[2026-07-24T11:29:10.472Z] czoko220: Hi, I love the functionality of seeing version history of character card and being able to revert to older ones. But using it for the first time was very confusing to me until I did some tests to figure out how it works. From my understanding, the current version of the card is not being displayed/saved in the version history? 
Why what is confusing to me? 
- Let's say right now my card version is v1, since it's the current version it's not saved in the version history. Then, I make some change to the card, thus creating a v2 version of it. The moment I press the save button, the v1 version of the card appears in the version history with the current timestamp - the timestamp of when I created v2, no v1. 
- Then, if I were to restore to v1 version, the v2 version would be lost forever since it wasn't saved in version history at any point. As far as I can tell, current version of the card is saved in history only when creating a newer version.

So in, short I believe it would be more intuitive (at least to me?) if the version history worked more like git commits rather than it's current state. 

I am not sure if this should be a separate post or not, but I also have another request related to the card history. Again, I really like the fact that I am able to compare current version of the card with older ones, but currently it only shows which section of the card has been changed. For example, description, starting message etc. This already is a huge help when comparing versions, but still, it means that if I really want to see the difference between two versions of 4k long character descriptions I need to manually copy both of them in some outside tool like diffchecker.com and compare them there. Having a view similar to that or github diff viewer would be amazing, but I will understand if you decide this is out of scope for what this feature should do or simply not worth the work for such niche functionality. 

ME version: v2.3.4
[Discord message: https://discord.com/channels/1417099416812392641/1530175005013446756/1530175005013446756]

[2026-07-24T11:36:25.492Z] Marinara: The first part definitely makes sense; the second is a little out of scope for now, since it would require coming up with a new UI… Well, actually, just a window displaying old on the left, new on the right? Shouldn't be too hard to implement.
[Discord message: https://discord.com/channels/1417099416812392641/1530175005013446756/1530176829619699834]
```

</details>

## Template check

SpicyMarinara:
## Why

From #4040. The reporter raised three things about character-card version history; two the maintainer endorsed, and the reporter's EDIT narrowed the third to a concrete bug:

1. **Restoring an older version permanently loses the newer one.** `restoreVersion` overwrote the current card without snapshotting it, so the newer state was gone forever.
2. **A saved version shows the wrong timestamp.** When you save v2, the snapshot of v1 was stamped with the v2 save time (`now()`), not with when v1 was actually saved — so v1 appeared to have been created at the moment you replaced it.
3. **The comparison view doesn't wrap long text.** The side-by-side panes used `whitespace-pre-wrap`, which wraps on spaces but not on long unbroken tokens — creator notes full of HTML and links overflowed the modal horizontally.

(The full inline word-level diff the reporter also mentioned is already covered by the existing per-field side-by-side view; the maintainer had deprioritized a fancier diff UI, so this PR keeps that view and just makes it wrap.)

## What changed

Server — `characters.storage.ts` (both characters and personas):
- **`restoreVersion` / `restorePersonaVersion`** snapshot the current card to history *before* overwriting it (source `"restore"`), skipped when the current card already equals the target version. Restoring is now non-destructive.
- **`update` / `updatePersona`** stamp the pre-save snapshot with the replaced card's own `updatedAt` instead of the new save time, so each version keeps its real date.

Client:
- **`CharacterEditor.tsx` / `PersonaEditor.tsx`** — the compare panes now use `min-w-0 break-words [overflow-wrap:anywhere]`, so long URLs/HTML wrap instead of overflowing; the restore confirmation now says the current card is saved to history first.

## Test plan

- [ ] Save a card as v2, then restore v1; confirm v2 now appears in version history (not lost) and the card is v1.
- [ ] Confirm a version's history timestamp reflects when that content was saved, not when it was replaced.
- [ ] Open the compare modal on a version whose creator notes contain a very long URL / unbroken HTML; confirm it wraps within the pane and the modal doesn't scroll sideways.
- [ ] Repeat the restore + wrap checks for a persona card.
- [ ] Confirm restoring a version identical to the current card does not create a redundant snapshot.
- [ ] `pnpm check` passes. (Passed in the authoring session; re-verify locally.)

Fixes #4040.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restoring an older character or persona card version now saves the current version to history first, preventing newer edits from being lost.
  * Version history entries now retain the correct edit timestamps.
  * Improved version comparison displays so long text, URLs, and HTML wrap correctly within creator notes.
  * Updated restore confirmation messages to clarify that the current card is preserved in version history before restoration.


SpicyMarinara:
@coderabbitai review

SpicyMarinara:
Applied both CodeRabbit findings in a9b6b4692:
- **Major (atomicity)**: both `restoreVersion` and `restorePersonaVersion` now snapshot the current card and overwrite it inside a single `db.transaction` (using `tx` for the read, snapshot insert, and update), so an interleaving edit or an interruption can't leave a half-applied restore or drop the history snapshot.
- **Minor (wording)**: the character and persona restore dialogs now note the pre-restore snapshot is skipped when the current card already matches the selected version ("…first (unless it already matches this version)…").

`pnpm check` passes.

PR #4100, opened 2026-07-25
Add Documentation Language selector with downloadable language pack (first pack: Spanish)

kolacheee:
## Why

International users currently get English-only in-app guides. This PR adds a **Documentation Language** setting (Settings › General, plus a final step in the onboarding tutorial) with **download-on-demand language packs**, and ships the first full translation: **Spanish**, covering all 123 guides (developer docs included, so non-English-speaking contributors can build PRs and extensions too).

Packs live on the companion **`docs-i18n` orphan branch** — never in user checkouts. Launchers only fetch `main`/`staging` and the installer clones a single tag, so translations add **zero checkout/image size**: important for Termux and thin-client installs as FIGS(P)/CJK land later (~1.3 MB per language ≈ 13 MB at ten languages, downloaded at most once by the users who want it).

## How it works

- **Selector**: Settings › General › Documentation Language (below the UI-language selector; findable via settings search) and a dedicated final onboarding step (pre-selects the UI language on fresh installs; on replays it pre-selects the user's existing choice, so an untouched picker is always a no-op).
- **Download & Replace**: picking an uninstalled language downloads its pack from `docs-i18n` into `DATA_DIR/doc-packs/<lang>` — manifest first, then every file verified against a **sha256 + byte size**, written to a temp dir, atomically swapped in — then the setting flips and **the replaced language's pack is deleted** (only one downloaded language on disk, ever). English is built-in and never downloaded or deleted. Live progress (`n/123 files`) shows in the button while downloading.
- **Robustness**:
  - Downloads are **pinned to the branch head's commit SHA** (resolved via one GitHub API call) so the raw CDN can never serve a mixed manifest/file set after a translation push; if the API is unreachable it falls back to branch-ref URLs where hash checks still guarantee integrity.
  - Fetches go through `safeFetch` (https-only, reserved/private-IP blocking — the `DOCS_I18N_BASE_URL` override for forks/mirrors can't become an SSRF vector), downloads are gated to the static supported-language list, and manifests get archive-grade path validation (traversal, Windows drive/ADS colons, case-folded duplicates, size caps).
  - Crashed installs leave only `.tmp-`/`.old-` folders that are swept at startup, install start, and Fix. A half-written pack can never serve: the manifest lands last and marks a pack installed.
- **Fix button** (shown when integrity breaks): full hash re-verification of the active pack, **re-download** when broken, sweep of crashed-install leftovers, **deletion of any other language's pack**, and reset-to-English only as a last resort when re-download is impossible.
- **The choice survives every update path** (data dir is gitignored / volume-persisted): launcher updates, in-app Apply Update, installer reinstalls, `docker compose pull`. A new English doc that the pack doesn't cover yet simply serves in English with an `EN` badge (per-file fallback) — features are never blocked on translations.
- **Packs auto-refresh after updates**: once per Engine build (APP_VERSION + git HEAD, so staging git updates count), the first start compares the selected pack against the `docs-i18n` manifest and re-downloads it if the translations changed; a stored choice whose pack is missing (downgrade, data restore) is restored the same way. Offline boots keep the installed pack serving and retry next start (documented in UPGRADING.md).
- **Search** keeps the literal substring + hit-count contract over the active language. Translations follow a terminology glossary (one standardized Spanish term per concept; community loanwords like *prompt/preset/lorebook/swipe* kept) so searches hit consistently.

## Drive-by fix

The shared markdown renderer accepted an indented code-fence *opener* but not an indented *closer*, so fences nested inside list items (e.g. `docs/development/architecture-map.md` § Migration rules) never closed and rendered their ` ``` ` markers as literal text. Close detection is now symmetric and, per CommonMark, the opener's indent is stripped from content lines so list-nested blocks don't render or copy phantom leading spaces. Column-zero fences (all chat content) are unchanged by construction.

## New API surface

- `GET /api/docs`, `/search`, `/content`: optional `?lang=`, additive `language` fields; `root` now reports the active language's on-disk folder.
- `GET /api/docs/language` — active/configured, supported languages with installed + coverage state, install progress (client polls at 1s during downloads), integrity flags. Never hashes on this path (cheap stat checks only).
- `PUT /api/docs/language` — validated switch; downloads when needed (400 unsupported / 409 install-in-progress / 502 download-failed).
- `POST /api/docs/language/fix` — the failsafe described above.
- New env var `DOCS_I18N_BASE_URL` (documented in CONFIGURATION.md) for forks/mirrors.

## Validation done

- `pnpm check` green; reworked `pnpm regression:docs` lane (fixture pack in a temp DATA_DIR): two-root containment + traversal, manifest path-safety (zip-slip equivalents, case-folded dupes, size caps), truncation detection, build-manifest determinism, and a pinned invariant that doc content never renders via `dangerouslySetInnerHTML` (downloaded packs are less-trusted than in-repo docs).
- Runtime-tested against a throwaway data dir: startup sweep, switch with leftover-pack deletion, Spanish serving/search from a materialized pack, corrupt-pack detection, Fix (failed re-download → clean reset + cleanup), 400/409/502 paths. The 502 test exercised the real `raw.githubusercontent.com` request path.
- Translation QA: per-file structural verification, terminology-consistency sweeps, independent language-QA panel over 20 representative guides.
- Two adversarial multi-agent review rounds; all confirmed findings fixed (including a tutorial-replay path that silently reverted the language, and the CDN manifest/file skew closed by commit pinning).

## Summary by CodeRabbit

* **New Features**
  * Added a Documentation Language setting with Spanish support.
  * Documentation packs download on demand, verify integrity, refresh automatically, and fall back to English when needed.
  * Added documentation language selection to onboarding.
  * Added English fallback badges and localized documentation viewer controls.

* **Bug Fixes**
  * Improved rendering of indented and nested code blocks.
  * Fixed phantom indentation when copying code.

* **Documentation**
  * Added guidance for maintaining translated documentation and configuring translation sources.


kolacheee:
@coderabbitai review

PR #4129, opened 2026-07-26
Rename Noodle "private" to "noodler" across code, data, and UI

Gunterlie:
## Linked issue

**No issue exists for this yet** — one should be opened before this PR is submitted, per `CONTRIBUTING.md § Before You Open a Pull Request`. The rename came out of hands-on Noodle/NoodleR work rather than a filed report.

Closes #

## Why this change

"Private" was the word for the NoodleR side of the Noodle/NoodleR split. It was the wrong word in two directions:

- **Users read it as a real privacy guarantee.** In a local engine, "private account" sounds like a security promise. It never was one — it only means the account lives on the other simulated platform. Everything is on the user's own disk either way.
- **It collided with the genuine privacy concept one level down.** Post-level `access: "public" | "subscriber" | "ppv"` is the actual paywall/visibility control. Having two unrelated meanings share one word made both humans and coding/review agents reason about the wrong thing.

The flag was `noodleAccounts.visibility` — semantically *"which simulated platform does this account belong to"*, not a visibility control. The field name was as misleading as its values, so both were renamed.

**Why now:** the vocabulary had to be settled before anything is built on top of it — every later slice would otherwise inherit and spread the wrong word. With that done, the next slice can move to onboarding, accessibility, and usability.

## What changed

Behaviour-preserving rename. `visibility: "public" | "private"` → `platform: "noodle" | "noodler"`.

| Before | After |
|---|---|
| `noodleAccounts.visibility` (column `visibility`) | `noodleAccounts.platform` (column `platform`) |
| `NoodleAccountVisibility` | `NoodlePlatform` |
| `publicAccountId` / `public_account_id` | `noodleAccountId` / `noodle_account_id` |
| `noodle-private-*.ts` (5 files) | `noodle-noodler-*.ts` (`git mv`, history preserved) |
| `generatePrivatePost` + `generateNoodlePrivatePost` | `generateNoodlerPost` + `generateAndApplyNoodlerPost` |
| `privateGenerationGuidance` | `noodlerGenerationGuidance` |
| `POST /accounts/:id/private` | `POST /accounts/:id/noodler` |
| `mode: "private"`, `NOODLE_PRIVATE_POST_*` | `mode: "noodler"`, `NOODLER_POST_*` |

NoodleR media naming takes a clean break rather than a migration, since NoodleR is in alpha and orphaning NoodleR-only data is acceptable:

| Before | After |
|---|---|
| on-disk prefix `"noodler-private/"` | `"noodler-media/"` |
| persisted `metadata.privateMediaPath` | `metadata.noodlerMediaPath` |

⚠️ **Existing NoodleR post images will 404 after this change.** NoodleR *accounts* still migrate properly — getting that wrong leaks them into the Noodle timeline, which is Noodle breaking rather than NoodleR breaking.

This also exposed a latent duplication, now fixed: the prefix was hardcoded a second time in `noodle-noodler-images.service.ts`, so the two copies had to agree by hand. The constant is exported and reused.

**Deliberately NOT renamed** — neither is a migration-cost decision, so alpha status doesn't change them:

- Post-level `access: "public" | "subscriber" | "ppv"` — this *is* the real privacy concept and should keep owning the word. Renaming it would leave no honest name for the one genuine privacy control.
- HTTP `Cache-Control: private, no-store` — that's the HTTP spec's keyword, not our vocabulary.

### Two defects the rename surfaced

1. **Broken cascade.** `CASCADES` in `file-backed-store.ts` references the account link column *by string*, so deleting a Noodle account had silently stopped cascading to its NoodleR profile. TypeScript cannot see this; the regression lane caught it.
2. **Settings would have silently reset.** Noodle settings persist as one JSON blob, so renaming the guidance key would have reverted every user's customized guidance text to the shipped default. Added a legacy alias in `normalizeNoodleSettings`, following the existing `maxImagePromptsPerDay` precedent — and folded it into #4109's new partial-fallback recovery so a customized string survives both the rename and a field-level validation failure.

### Rollback safety

Caught by booting this branch against a legacy data dir, then booting a **staging** build against the same dir: the NoodleR account and its subscriber-only post both showed up on the **public Noodle timeline**. Staging doesn't know `platform` exists, so it fell back to the column default and treated the stage profile as an ordinary Noodle account.

Fixed by keeping `visibility` and `publicAccountId` as **write-only rollback mirrors**. Nothing reads them; the migration no longer strips them, and both account insert sites write them alongside the new fields. Re-tested both paths — migrated legacy rows *and* profiles created fresh on this branch — and staging now sees neither on the public timeline. They can be dropped once no supported version reads `visibility`.

### Data migration

`packages/server/src/db/noodle-platform-migration.ts`, applied at load in `file-backed-store.ts` **before** row normalization. It has to run there: the loader falls back to a column's default when neither its key nor db name is present, so an unmigrated NoodleR profile would have loaded as `platform: "noodle"` and **leaked into the public Noodle timeline**. Idempotent; rewrites the legacy shape on next flush.

## Validation

Commands run locally:

- `pnpm check` — passes. The one eslint warning is pre-existing (confirmed by running eslint against a stashed tree).
- `pnpm regression:noodle` — 14/14 pass, including the new `noodle-platform-migration.regression.ts` and staging's `noodle-context` lane.
- `pnpm regression:docs` — passes.

New regression `scripts/regressions/noodle-platform-migration.regression.ts` covers the pure migration (legacy → new, already-migrated → untouched, missing field → `"noodle"`) and drives a legacy `visibility: "private"` row end-to-end through the real file store, asserting it does not appear in `listAccounts()` and still resolves as the linked NoodleR profile.

**Rename correctness was verified mechanically**, not by eyeball: for each renamed file, old→new tokens were normalized in the pre-rename version and diffed against the current tree, so every surviving difference had to be justified. `noodle-response-format.ts` and `noodle.schema.ts` diff to zero (pure rename); everything else reduced to the intended name-collision split, route path, DB column, and copy rewrites. This is how the leftover `setDraftPublicAccountId` setter and three doubled-up comments were found.

### Manual verification notes

**None of the below has been done — no UI was driven for this change.** These are the checks a human still needs to perform:

- Manually verify in a browser: with `enableNoodler` off, NoodleR surfaces still 404 / show the opt-in screen.
- Manually verify creating a NoodleR profile from a Noodle account still works, and that a second attempt on the same account still returns the 409 conflict with the new wording.
- Manually verify generating a NoodleR post with an image, then confirm a locked (subscriber/PPV) post exposes neither image URL nor prompt.
- Manually verify no NoodleR post appears in the Noodle timeline, follows, or refresh.
- Manually verify deleting a Noodle account also removes its NoodleR profile (this is the cascade fixed above — worth checking in the real app, not just the regression).
- **Migration check on real data:** back up a data dir from before this branch, boot the server, and confirm existing NoodleR profiles still load with their posts and media rather than appearing on the Noodle timeline. Boot twice to confirm idempotency.
- Manually verify a previously customized NoodleR generation guidance string survives the upgrade instead of resetting to the default.
- Manually verify that NoodleR posts created *after* this change render their images correctly under the new `noodler-media/` namespace (pre-existing NoodleR images are expected to 404 — that is the accepted alpha breakage, not a bug to fix).

## Docs and release impact

- [ ] No docs changes needed

Notes for whoever ticks these:

- `docs/noodle/settings.md` changed, and now states plainly that Noodle/NoodleR separation is content separation between two simulated apps, **not** a privacy or security feature. **A `docs-i18n` update or a `[docs-i18n] docs/noodle/settings.md` follow-up issue is still needed.**
- `en.json` and `ko.json` copy updated. Other locales carry only ~1–6% of keys and were untouched.
- `.github/plans/noodl-split/noodler-pr-split-living-plan.md` gained a terminology section so remaining slices are written in the new vocabulary; historical slice descriptions were left as-is.
- `docs/noodle/overview.md` and `docs/development/noodle-internals.md` contain no NoodleR content at all and are stale — out of scope here, worth a follow-up issue.
- No version bump.

None — no visual changes beyond label/copy text. Copy changes are listed above.

## Summary by CodeRabbit

* **New Features**
  * Introduced consistent NoodleR terminology across stage profiles, posts, navigation, generation, and account management.
  * Added clearer separation between the Noodle app and NoodleR, including updated settings and visibility guidance.
  * Added automatic migration support for existing NoodleR account data.

* **Bug Fixes**
  * Improved locked-post labels, accessibility text, localization coverage, and NoodleR media handling.

* **Tests**
  * Expanded regression coverage for localization, migration, generation, settings, and autopost images.


Gunterlie:
@coderabbitai Review

Gunterlie:
@coderabbitai Went through every finding. 9 fixed, 1 skipped — details below.

**Fixed**

- `en.json` adult-access copy: "Private creator accounts" → "NoodleR creator accounts"; mirrored the same de-"private" wording in `ko.json`.
- `noodle.routes.ts` removeImage cleanup: the post is now re-read *inside* `tryNoodlerAccountOperation`, and `unlinkNoodlerMedia` uses that in-lock path instead of the pre-lock `existing` snapshot. 409/404/200 behavior unchanged.
- `noodle-response-format.ts`: added `imagePrompt` to `noodlerPostSchema`. Note it also had to go in `required` — the schema runs with `strict: true`, where every property must be required and optionality is expressed as a nullable type. Without this, `generated.imagePrompt` was permanently null on structured-output models.
- `noodle.schema.ts`: the draft-request refinement is now XOR (`Boolean(a) !== Boolean(b)`), message preserved. Previously both-set silently ignored `noodleAccountId` in `noodle-stage-profile-draft.service.ts`.
- `localization-key-references.regression.ts`: `KNOWN_MISSING` is now a `Set` of the 31 actual keys. Fails on any missing key outside the list, and also fails (asking you to shrink the baseline) when an allowlisted key starts resolving.
- `noodle-platform-migration.regression.ts`: retained the close callback and moved the call into `finally` before `rmSync`.
- `noodle-settings.regression.ts`: `assert.ok` on `firstSubscription` / `firstUnlock` before the existing ID equality checks, which passed on `undefined === undefined`.
- Composer placeholder: renamed `…whatSSimmeringPrivately` → `…whatSSimmering` in `en.json` + `ko.json`. There were **two** call sites, not one — `NoodlerHome.tsx:3665` and `NoodlePostCard.tsx:1477`; both updated.
- `noodle.storage.ts`: dropped the local `PollVoteAuthorPlatform` union for the shared `NoodlePlatform`. Small correction: it was not previously imported in that file, so this added the import.

**Skipped: media-route authorization (`/noodler/posts/:id/media`, persona-less branch)**

There is no existing request auth or ownership/management check to reuse — `preHandler`, `requireAuth`, `authGuard`, `isManagement` all return zero hits across `packages/server/src/routes/`. Marinara Engine's server is a local single-user app and every route is equally unauthenticated, so this isn't a regression introduced by the rename; adding an auth layer is a separate feature, not a fix belonging on this PR.

**Validation:** `pnpm check` passes. `regression:localization`, `noodle-platform-migration`, and `noodle-settings` regressions all pass. Not manually verified in a browser yet: the reworded adult-access copy and both renamed composer placeholders.

Gunterlie:
Thanks @xelvanas — all five findings are fixed in 1059782b1.

**High — importer bypassed the platform migration** (`backup.routes.ts`)
`noodle_accounts` rows now go through `migrateLegacyNoodleAccountRow()` before insert, so both the fresh insert and the `onConflictDoUpdate` set see `platform`/`noodleAccountId`. Your diagnosis was exactly right: the v1 snapshot path (and the v2 archive path that reuses it) was the one insert route that never touched the loader migration.

**Medium — persisted UI store never migrated** (`ui.store.ts`)
Bumped to `version: 86` and the guard is now `version <= 85`. You're right that the base already shipped 85, so the migration was dead code for every realistic upgrader.

**Low — old NoodleR destinations discarded** (same guard)
Folded into the above: `profiles`, `profile` + `accountId`, and `create-profile` + `publicAccountId` → `noodleAccountId` are now translated structurally; the hub is only the fallback for malformed values.

**Low — rollback mirror dropped on settings write** (`noodle.storage.ts`)
`updateSettings` now serializes `privateGenerationGuidance` as a write-only mirror of `noodlerGenerationGuidance`, matching the `visibility`/`publicAccountId` mirrors, with a comment on when it can be dropped.

**Medium — Korean copy still said "private"** (`ko.json`)
Rewrote the five live values to describe NoodleR / a separate simulated app rather than `비공개`. Also fixed `noodlerIsAnOptionalPrivateSpace` while in the file — it carried the same misleading claim and would have been a trap for whoever wires it up later.

**Validation**
- `pnpm check` — passes; the single eslint warning is the pre-existing `localizeUi` dependency one.
- `pnpm regression:noodle` — 14/14.
- `pnpm regression:docs` — passes.
- New coverage: `noodle-platform-migration.regression.ts` drives a legacy row through the importer path and asserts it stays out of `listAccounts()` and resolves as NoodleR; `noodle-settings.regression.ts` asserts legacy-key read → `updateSettings` → old key still present in the persisted blob.
- `pnpm regression:localization` fails on `chat.delete.branchChoice` and `ui.panels.ttsconfigcard.loadedVoiceCount` — confirmed pre-existing by re-running against a stashed tree, unrelated to this PR.

**Still needs a human in a browser** (unchanged from the PR description, none of this was driven in UI):
- Reopen the app with a persisted v85 store sitting on a NoodleR profile and confirm it lands on that profile rather than the hub.
- Restore a real pre-rename backup archive and confirm the NoodleR account and its posts stay off the Noodle timeline, including after a restart.
- Confirm the Korean strings render correctly at their actual widths.

Gunterlie:
Thanks again @xelvanas — all three are fixed in a35d643a3.

**Medium — localization checker rejected valid plural keys**
You were right that the failing keys were not the problem; the gate was. The checker now treats an i18next plural family (`_zero|_one|_two|_few|_many|_other`) as resolving its base key. That turned out to explain **30 of the 31** `KNOWN_MISSING` entries — the allowlist was almost entirely false positives that I had inherited without checking. Only `ui.game.gamecharactersheet.regenerating` is genuinely dangling and remains listed.

`pnpm regression:localization` now passes, so `pnpm regression` is no longer a deterministic failure.

**Nitpick — `private` still in localization IDs**
- `ui.noodle.noodlerframe.private` → `ui.noodle.noodlerframe.noodler`, renamed in `en.json`, `ko.json`, and its one caller in `NoodlerHome.tsx`.
- `noodlerIsAnOptionalPrivateSpace` and `noodlerIsIntendedForAdultsPrivateCreatorAccountsStay` were unreferenced in both catalogs, so I deleted them rather than renaming dead entries. Say the word if you would rather they were kept and renamed — they are cheap to restore, and the second one also still carried the "isolated from the public Noodle timeline" phrasing this PR is retiring.

**Nitpick — misleading migration comment**
`file-backed-store.ts` now says the flush persists the renamed fields *alongside* the retained `visibility` / `publicAccountId` rollback mirrors, instead of claiming the legacy shape is dropped.

**Validation**
- `pnpm check` — passes; still the one pre-existing `localizeUi` exhaustive-deps warning.
- `pnpm regression:localization` — passes (1 known-missing, none new).
- `pnpm regression:noodle` — unaffected by this commit and passing as of the previous push.

No new manual verification beyond what is already listed; the key rename is covered by the reference regression that now actually works.

SpicyMarinara (approved):
Maintainer review: CodeRabbit feedback is complete, review threads are resolved, and required checks are green.

PR #4167, opened 2026-07-27
fix(chat): prefill default connection on new chats + label wizard connection select

Gunterlie:
## Why

With a **Main Default** and a **Fallback** connection configured in the Connections tab, clicking **New Chat** (from either the Characters or Chats menu) opens the setup wizard with the Connection field empty. The dropdown also had no field label, unlike every other field in that step.

It looked like a config mistake but it isn't:

- All new-chat entry points pass `connectionId` **only** from a starred chat preset — `useStartChatFromCharacter` (`use-start-chat-from-character.ts:48`), `ChatSidebar.tsx:622`, `HomeNewChatLauncher.tsx:38`. With no starred preset, `connectionId` is `undefined`.
- The server create route didn't fill it in either, so `chat.connectionId` stayed `null` and the wizard rendered "None".
- Generation still worked, which is why this reads as cosmetic: it falls back to `connections.getDefault()` at request time (`chats.routes.ts:1165`). The *fallback* connection is only for post-failure retry, so it was never going to appear here.
- `NewChatConnectionGate` did prefill, but with `connectionRows[0]` — first in the list, not the user's default. A third inconsistency.

## What changed

- `packages/server/src/routes/chats.routes.ts` — `POST /chats` seeds `connectionId` from `connections.getDefault()` when the client omits it. Fixing it here means all four callers agree instead of patching each one.
- `packages/client/src/components/chat/ChatSetupWizard.tsx` — added the missing `<label>` wrapper on the roleplay step's connection `<select>` (the conversation step already had one). Also an a11y gap: the control had no accessible name.
- `packages/client/src/components/chat/NewChatConnectionGate.tsx` — prefill the default connection rather than the first row. Note `isDefault` arrives as the string `"true"`/`"false"` over the wire, so this compares explicitly instead of testing truthiness.

No schema change, no migration. Existing chats are untouched — this only affects newly created ones.

## Validation

- [ ] `pnpm check`
- [ ] `pnpm version:check` (not expected to be relevant — no version-bearing files touched)

## Test plan

- [x] Configure a Main Default connection and a Fallback connection, with **no** starred chat preset. Click New Chat from the Chats menu — manually verify the wizard's Connection dropdown shows the default connection, not "None".
- [x] Repeat from the Characters menu (Start Chat on a character) and from the Home screen New Chat launcher.
- [x] Manually verify the Connection dropdown in the **roleplay** wizard step now shows a "Connection" label above it, matching the conversation step.
- [x] With zero connections configured, click New Chat — manually verify `NewChatConnectionGate` still shows the "No connections found" empty state and that creating a chat after adding one works.
- [x] With a starred chat preset that specifies a connection, manually verify the preset's connection still wins over the default.
- [x] Manually verify in the browser that changing the connection in the wizard still persists to the chat.
- [ ] Regression check: import a chat (the import path also hits `POST /chats`) and manually verify an imported chat's connection is not unexpectedly overwritten.

## Notes

There is no linked issue for this — one should be opened before this PR leaves draft, per `CONTRIBUTING.md § Before You Open a Pull Request`.

No user-facing docs under `docs/` changed, so no `docs-i18n` follow-up is needed. The one new UI string reuses the existing `ui.chat.conversationquicksetup.connection` key, so no new locale entries.

## Summary by CodeRabbit

* **New Features**
  * Improved connection selection in the chat setup wizard with a clearer labeled layout.
  * Automatically selects the configured default connection when starting a new chat.
  * New chats now use the default connection when no connection is explicitly selected.


Gunterlie:
@coderabbitai review

PR #4373, opened 2026-07-31
fix(capabilities): require refresh after client updates

### LINKED ISSUE #4372: [Issue]: Updated capability client keeps the previous custom-element UI until page refresh
## Summary

Updating an installed capability package that contributes client UI can leave an already-open Marinara browser tab running the previous custom-element implementation. The server and installed manifest report the new package version, but newly added UI remains unavailable until the user performs a full page refresh.

This was observed with World Maps 1.2.3 after the shared-world Game Setup integration landed in Pasta-Devs/Marinara-Engine#4329 and Pasta-Devs/Marinara-Agents#149: Game Setup listed both shared worlds and templates, but shared-world cards still exposed the older non-setup actions while templates could be selected.

## Environment

- Engine release: v2.3.5
- Live Engine commit: `e1322b0ff1a0`
- World Maps: 1.2.3, active and ready
- Installed/served World Maps client SHA-256: `e7376ff6da77adb475bf6bdf42db9238108e0e69c4b37be4a6c97552e6bab713`

## Evidence

- The installed manifest and live `/api/capability-packages/hierarchical-maps/client?v=1.2.3` response have the same SHA-256.
- The served module contains `supportedSelectionKinds`, `Use shared world`, and the `setup-apply` view.
- `useCapabilityClientModules()` keys dynamic imports by package version, but the package module registers its custom element only when `customElements.get(tag)` is absent.
- After an update, the old element constructor remains registered for the lifetime of the page. Importing the new module cannot replace it, while the loader currently treats the existing tag as proof that the new module is ready.

## Reproduction

1. Open Marinara with an older version of a capability package that contributes client UI.
2. Update the package without closing/reloading the browser tab.
3. Complete any required Engine restart.
4. Open a surface that uses UI added by the new package version.
5. Observe that the old custom-element implementation remains active until a full page refresh.

## Expected behavior

When a ready client-capability version changes while an older constructor is already registered, Marinara should complete the update by refreshing the web shell (or clearly require and enforce a refresh) instead of silently reporting the new module as ready while rendering the old UI.

## Scope

This is an Engine capability-client lifecycle bug, not a World Maps source/artifact omission. A hard page refresh is the immediate workaround.

thetopham:
Closes #4372

## Why this change

An already-open browser tab cannot replace a registered custom-element constructor when a downloaded capability package updates. Marinara imported the new module and reported it as ready, but continued rendering the old client implementation until the user happened to refresh the page. This made World Maps 1.2.3 appear to be missing its shared-world Game Setup action even though the installed and served artifact contained it.

## What changed

- Preserve the version of capability client code that is actually registered for the lifetime of the page, including across restart-required package states.
- Detect a newer ready package when the previous version's custom element is still registered.
- Serialize in-flight client imports per package/version so an older pending import cannot race a newly installed version.
- When an older import settles after a newer version is observed, require refresh if its custom element registered; otherwise retry the latest installed client.
- Block the stale interface and show a localized **Refresh now** state instead of reporting the new module as ready.
- Add focused regression coverage for old, current, fresh-page, unregistered-element, and overlapping-import version combinations.

## Validation

- `pnpm check`
- focused capability-client version/import race regression
- `pnpm localization:check`
- `git diff --check`

## Manual verification

- [ ] Update a client-capability package while its previous version is loaded and verify its surfaces show **Refresh now** rather than stale UI.
- [ ] Refresh and verify the newly installed client UI loads without another refresh prompt.
- [ ] Complete a restart-required client package update, restart Marinara Engine, and verify the same refresh state appears.
- [ ] Verify a first-time package install loads normally without an unnecessary refresh prompt.


thetopham:
@coderabbitai review

PR #4379, opened 2026-07-31
feat(maps): add direct world library navigation

### LINKED ISSUE #4365: [Feature]: Add first-class navigation for an installed World Maps package

World Maps is a substantial optional feature with account-owned shared worlds, reusable templates, and per-chat runtime state, but its global library is reached through **Global Agents → World Maps → Open Map Templates**. Chat-owned maps are reached separately through the active chat's Agent menu.

Beta testers found this difficult to discover and described the shared-world workflow as back-and-forth. One tester specifically expected a globe/World Maps entry in the main navigation beside Lorebooks. The current placement makes a feature-sized workspace feel like a deeply nested agent setting and obscures the distinction between global shared-world authoring and chat map access.

## Proposed solution

When the World Maps capability package is installed and active, expose a conditional top-level **World Maps** navigation item—using a globe or map icon—in the main application navigation near Lorebooks.

- The global entry should open the package-owned World Maps library for shared worlds and templates.
- It should be hidden or unavailable when the package is not installed/active, preserving World Maps as an optional feature.
- Chat-specific map access should remain available from the chat/runtime surface and should be labeled distinctly from the global library.
- Mobile navigation should expose an equivalent reachable entry without crowding the primary bar.
- Prefer a generic capability-package navigation contribution if feasible so Engine does not hardcode World Maps-specific UI ownership.
- Preserve package ownership: Engine supplies navigation and lifecycle hosting; Marinara-Agents continues to own the actual World Maps library and editor.
- Coordinate with the package-side shared-world navigation issue so the destination clearly distinguishes shared worlds, templates, and the active chat map.

## Alternatives considered

- Keep access only under Global Agents: functional, but testers did not discover the shared-world library naturally.
- Documentation alone: useful for onboarding, but does not reduce repeated navigation cost while actively editing shared worlds across chats.
- Promote all World Maps implementation into Engine core: unnecessary. A conditional first-class entry can improve access while the feature remains optional and package-owned.
- Add a permanent entry when Maps is not installed: misleading and contrary to the optional-package model.

## Additional context

- Marinara Engine version: v2.3.5
- Engine build: a121d12
- World Maps package version: v1.2.3
- Reported use case: editing one shared world used by two Roleplay chats

The package already contributes its global library/editor UI; this request is for host-level navigation to that installed capability, not for Engine to duplicate the feature.

Related but distinct: closed #4323 added shared-world selection during Game setup. This issue concerns everyday global navigation after installation.

## Template check

thetopham:
## Why

World Maps currently behaves like a substantial world-building workspace, but users have to know that its global library is hidden behind Agents while its per-chat controls live elsewhere. That makes shared-world ownership and editing harder to discover.

Closes #4365.

## What changed

- show a dedicated World Maps globe in the desktop top bar when the package is active and ready
- show the same action in the mobile Chats drawer without crowding the mobile top bar
- open the existing package-owned World Maps detail/library instead of duplicating map UI in Engine
- keep the Agents button from also appearing active while the dedicated World Maps view is open
- make **Back to Agents** close the Maps detail and activate the Agents panel
- localize the new accessible label and extend the existing Hierarchical Maps browser smoke

## Validation

- `pnpm check`
- `pnpm localization:check`
- `pnpm impeccable:check`
- `pnpm exec playwright test -c playwright.config.ts --project=desktop-chromium --grep "Hierarchical Maps settings stay inside the active agent entry"`
- `git diff --check`

## Manual verification

- [ ] Manually verify the globe appears in the desktop top bar only while World Maps is active and ready.
- [ ] Manually verify the globe appears in the mobile Chats drawer and closes the drawer when opened.
- [ ] Manually verify both entry points open the existing World Maps library with the current chat context when available.
- [ ] Manually verify **Back to Agents** activates the Agents panel.
- [ ] Manually verify the globe follows the selected theme/RGB colors.

## Summary by CodeRabbit

* **New Features**
  * Added a World Maps button to desktop navigation when the feature is available.
  * Added a globe button to the mobile sidebar for quick access to World Maps.
  * World Maps navigation now uses localized labels and visual active-state styling.

* **Bug Fixes**
  * Returning from a map detail view now restores the Agents panel.
  * Improved navigation state so the Agents section is not incorrectly marked active while viewing map details.


thetopham:
@coderabbitai review

PR #4407, opened 2026-08-01
Fix character-card action sizing regression assertion

### LINKED ISSUE #4406: Fix character-card action sizing regression assertion
## Problem

The full UI smoke suite passes 173 tests but fails the existing character-card action sizing assertion in both desktop and mobile Chromium.

The character-row action container is a three-column grid. In the failing flow, Add to active chat is intentionally hidden because the character is already assigned, but the grid still reserves all three columns. The assertion calculates the full-width Chat action from only the two visible Duplicate and Delete columns, so it undercounts one column and one gap.

## Scope

Update the Playwright sizing assertion to account for the fixed three-column action grid. Keep this as a focused test-only correction with no runtime UI changes.

## Acceptance criteria

- The focused character action flow passes in desktop Chromium.
- The focused character action flow passes in mobile Chromium.
- The assertion continues checking consistent action heights, equal visible column widths, and full-row Chat sizing.

SpicyMarinara:
Closes #4406

## Why

The character-row action container is a three-column grid. In this flow, Add to active chat is intentionally hidden because the character is already assigned, while the grid still reserves all three columns. The existing Playwright assertion calculated the full-width Chat button from only the two visible action columns, causing the otherwise healthy smoke suite to fail in both desktop and mobile viewports.

## What changed

- Verify Duplicate and Delete retain equal column widths.
- Calculate the full-width Chat button from all three grid columns and both gaps.
- Keep this as a test-only correction; no runtime UI behavior changes.

## Validation evidence

- Focused Playwright run: 2 passed (desktop Chromium and mobile Chromium).
- `pnpm check`: passed.

## Manual validation

- [ ] Run the focused Playwright test in desktop and mobile Chromium.
- [ ] Run `pnpm check`.

### DIFF
--- a/e2e/core-flows.e2e.ts
+++ b/e2e/core-flows.e2e.ts
@@ -1997,6 +1997,10 @@ test("Character Chat actions reuse mode selection and seed the chosen setup wiza
 
     const actions = characterRow.locator("[data-character-row-actions]");
     await expect(actions).toBeVisible();
+    const addToActiveChatButton = actions.getByRole("button", {
+      name: `Add ${characterName} to the active chat`,
+      exact: true,
+    });
     const duplicateButton = actions.getByRole("button", { name: "Duplicate", exact: true });
     const deleteButton = actions.getByRole("button", { name: "Delete", exact: true });
     const chatButton = actions.getByRole("button", {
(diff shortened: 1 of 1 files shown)


SpicyMarinara:
@coderabbitai review

PR #4413, opened 2026-08-01
feat(chat): mobile drop dock for chat resource drag-and-drop

### LINKED ISSUE #4412: Mobile: no drag-and-drop target for chat resources while the library panel is open
The chat resource drag-and-drop flow (characters, lorebooks, personas, presets, connections, backgrounds) is effectively desktop-only.

On mobile the right-hand library panel renders as a full-width overlay (`AppShell.tsx`, `!fixed bottom-0 right-0 z-40 !w-full`) with a backdrop above the chat, so once a user long-presses a row to start a touch drag there is no visible `[data-chat-resource-drop-surface]` to drop onto. Touch drags currently only resolve against folder targets inside the panel.

The tap-only `ChatResourceActionButton` covers the capability, so this is a parity/discoverability gap rather than a missing feature.

Proposal: a small drop dock pinned to the left edge, rendered only while a chat-resource touch drag is active, that dispatches the existing `CHAT_RESOURCE_ASSIGN_EVENT` assignment flow (same path as the row action button) and closes the panel so the result, toast and undo are visible.

Implementation in progress on `feat/mobile-chat-resource-drop-dock` (@Gunterlie).

Gunterlie:
Closes #4412

## Why

Chat resource drag-and-drop is desktop-only in practice. On mobile the library panel is a full-width overlay over the chat, so a touch drag has nowhere to land — there is no visible `[data-chat-resource-drop-surface]`, and touch drops only ever resolved against folder targets inside the panel. Users on mobile can only use the tap-only `ChatResourceActionButton`, so the gesture that the whole feature is built around simply does not exist for them.

## What

A left-edge drop dock that appears only while a chat-resource **touch** drag is active, sits above the panel (`z-9000`, below modals at `z-10000`), and shows the same icon/accent/title as the desktop overlay — including the blocked state for already-active or unsupported resources.

Dropping on it closes the right panel and dispatches `requestChatResourceAssignment(payload)`, i.e. the exact same assignment path the desktop overlay and the row action button already use. No new mutation, confirmation or modal logic.

- `chat-resource-drag.ts`: touch drags carry no `dataTransfer`, so the active payload is held in module state with a subscribe/notify pair the dock reads through `useSyncExternalStore`.
- `use-touch-folder-drag.ts`: one optional `chatResourcePayload` start option; the hook sets it on activate and clears it on cancel/drop, so lifecycle stays in one place. No behavior change when the option is omitted.
- The 8 chat-resource drag cards pass the payload they already build for `onDragStart`.
- The dock hit-tests its own cached rect rather than reading `getBoundingClientRect` per `touchmove`, and listens in the capture phase so the payload is still set when it reads it.

Panel folder drops need no changes: they resolve via `elementFromPoint(...).closest("[data-*-folder-id]")`, and the dock is not a folder, so that branch already no-ops.

### Panel closes on drop

Deliberate divergence from the tap button, which leaves the panel open. A drop means "put this in the chat"; if the panel stayed open the success toast — the only route to **undo** — would fire behind a full-screen overlay. The connection-needs-setup branch reopens the panel afterwards, which is the correct ordering since the close happens before the assignment dispatches.

## Validation

- [ ] `pnpm check`
- [ ] `pnpm localization:check`
- [ ] `pnpm regression:chat-resource-drop`
- [ ] `pnpm smoke:ui`

## Test plan

- [ ] Manually verify on a real touch device (or mobile emulation) that long-pressing a character row in the library panel shows the dock at the left edge
- [ ] Manually verify the dock highlights when the finger enters it and returns to rest when it leaves
- [ ] Manually verify dropping on the dock closes the panel and adds the character to the chat
- [ ] Manually verify the blocked state renders for a resource already active in the chat, and that dropping it does nothing
- [ ] Manually verify each resource kind: character, lorebook, persona, preset, connection, background
- [ ] Manually verify the greeting/lorebook modal after a character drop is legible and dismissible at 360px width
- [ ] Manually verify the preset choice modal (`ChoiceSelectionModal`) at 360px width
- [ ] Manually verify the persona/preset/connection replacement confirmation appears above the panel and its first tap is not swallowed
- [ ] Manually verify dropping a connection with no model still opens connection setup
- [ ] Manually verify the success toast and its Undo action are reachable after the panel closes
- [ ] Manually verify dragging a row onto a folder still moves it into that folder (no regression to folder drag)
- [ ] Manually verify the dock does not appear on desktop drags

## Notes

- Automated coverage is the touch-drag state lifecycle in `chat-resource-drop.regression.ts`; the dock geometry and gesture itself are not covered by the Playwright lane and need the manual pass above.
- Untranslated: `ui.chat.chatresourcemobiledropdock.label` is English-only, as with other new keys.

## Summary by CodeRabbit

* **New Features**
  * Added mobile touch-and-drop support for assigning agents and other resources to chats.
  * Added a mobile drop dock with clear action states, accessibility labels, and blocked-drop feedback.
  * Added touch dragging for agents, characters, connections, lorebooks, personas, presets, and backgrounds.
  * Chat settings now reopen the previous panel after resource assignment when appropriate.
  * Added agent setup handoff that opens chat settings automatically.

* **Bug Fixes**
  * Improved panel restoration and drag-state cleanup during resource assignment and cancellation.


Gunterlie:
@coderabbitai review

PR #4419, opened 2026-08-01
fix(agents): remove the duplicate drag handle on agent cards

Gunterlie:
Follow-up to #4413.

## Why

That PR added a `TouchDragHandle` to agent cards so agents could be touch-dragged. Agent cards already carried their own decorative grip — hover-revealed, `[@media(pointer:coarse)]:hidden`, no handler, purely an affordance for the native HTML5 drag. The new handle also hover-reveals on fine pointers, so hovering an agent card on desktop showed **two** grips side by side.

## What

Remove the decorative grip and keep `TouchDragHandle`, which already covers both pointer types and is exactly what the other six library panels do. The handle now also reuses the existing `ui.panels.agentcard.dragAgent` key, so the duplicate `ui.panels.agentspanel.dragAgent` key added in #4413 is dropped.

One small behavior difference: the old grip was hidden in selection mode, the shared handle is not — again matching the other panels.

## Validation

- [ ] `pnpm check`
- [ ] `pnpm localization:check`

## Test plan

- [ ] Manually verify hovering an agent card on desktop shows exactly one drag handle
- [ ] Manually verify the native desktop drag of an agent card still works
- [ ] Manually verify the handle is visible on a touch device and still starts a touch drag
- [ ] Manually verify agent cards in selection mode look right with the handle present

## Summary by CodeRabbit

* **UI Improvements**
  * Simplified agent cards by removing the redundant desktop drag handle.
  * Retained touch-based drag-and-drop controls for supported devices.
  * Removed an unused localization entry.

### DIFF
--- a/packages/client/src/components/panels/AgentsPanel.tsx
+++ b/packages/client/src/components/panels/AgentsPanel.tsx
@@ -18,7 +18,6 @@ import {
   Check,
   FolderPlus,
   ArrowUpDown,
-  GripVertical,
   ShieldCheck,
   TriangleAlert,
 } from "lucide-react";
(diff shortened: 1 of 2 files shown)


Gunterlie:
Folded into #4420 so both review follow-ups ship together. The commit is cherry-picked there unchanged.

PR #4422, opened 2026-08-01
fix: clear stale untracked source files on channel switch (#4421)

### LINKED ISSUE #4421: Channel switch (main → staging) can leave stale source files that break the build
When a user switches from the stable channel (main) to staging, files that exist on main but were deleted on staging can persist in the working tree as untracked files. The build then fails because `tsc` picks them up via `include: ["src/**/*.ts"]` and they reference types/enums that no longer exist on staging.

The immediate trigger is `packages/shared/src/constants/chat-mode-capabilities.ts` — present on main (where `ChatMode` includes `visual_novel`), deleted on staging via PR #4356. After a main→staging switch the file stays behind, and `tsc` errors on the stale `visual_novel` key:

```
src/constants/chat-mode-capabilities.ts:149:3 - error TS2353:
  Object literal may only specify known properties, and 'visual_novel'
  does not exist in type 'Record<ChatMode, ChatModeCapabilities>'.
```

### Why the file persists

On Windows, `git checkout` can fail to delete individual files (antivirus scanner, Windows Search indexer, or another process briefly locking the file). Git may still report the checkout as successful (HEAD moves to the target) while the locked file remains as an untracked leftover.

Once the file lingers, `start.bat` makes the problem **self-perpetuating**: its `git stash push -u` (line 190) captures the untracked file before every update, and `restore_stashed_changes` puts it back afterward — so the stale file survives every subsequent relaunch.

The in-app updater (`updates.routes.ts`) avoids this particular trap (its `createUpdateStash` uses `--untracked-files=no`), but neither update path runs any post-checkout cleanup to verify the working tree matches the target commit.

### Try it yourself

1. Fresh install of v2.3.4 (Windows zipfile/NSIS installer — creates a depth-1 clone at the tag).
2. Switch to the staging channel (in-app updater or manual `git checkout -b staging origin/staging`).
3. If the checkout silently fails to delete `chat-mode-capabilities.ts` (Windows file lock), the build fails with TS2353.
4. Relaunch via `start.bat` — the file is stashed with `-u`, restored after the no-op update, and the build fails again.

### Suggested fix

After a successful channel-switch checkout (both `start.bat` and `checkoutOrCreateUpdateBranch` in `updates.routes.ts`), run a targeted cleanup of source directories:

```bash
git clean -fd -- packages/shared/src packages/server/src packages/client/src
```

This removes untracked files in the source trees without touching user data, node_modules, or config files. A full `git clean -fd` would be too aggressive (could delete `.env`, data files, etc.), but scoping it to `packages/*/src` is safe — those directories should only contain tracked source.

An alternative (less targeted but more thorough): after any checkout that changes channels, compare `git ls-tree -r --name-only HEAD` against the working tree and remove files that exist locally but not in the target tree.

Gunterlie:
Fixes #4421.

## Why

A user switching from the stable channel to staging can end up with a checkout that never builds again. Files that exist on `main` but were deleted on `staging` can survive the switch as untracked leftovers — on Windows, `git checkout` can fail to delete an individual file (antivirus scanner or Windows Search indexer holding a brief lock) while still reporting the checkout as successful. `tsc` then picks the leftover up via `include: ["src/**/*.ts"]` and fails on types that no longer exist:

```
src/constants/chat-mode-capabilities.ts:149:3 - error TS2353:
  Object literal may only specify known properties, and 'visual_novel'
  does not exist in type 'Record<ChatMode, ChatModeCapabilities>'.
```

The launchers turned that from a transient glitch into a permanent one. `git stash push -u` captured the untracked leftover before every update and `restore_stashed_changes` put it back afterwards, so the broken tree survived every relaunch. There was no way out short of the user deleting the file by hand — and no message telling them which file to delete.

## What changed

Untracked files under `packages/{shared,server,client}/src` are now removed:

- after a channel-switch checkout, in `checkoutOrCreateUpdateBranch` (`packages/server/src/routes/updates.routes.ts`)
- once per launch in `start.bat`, `start.sh`, and `start-termux.sh`

Placement notes, both deliberate:

- **Before `git stash push -u`**, so the stale file can't be captured and restored again.
- **Above the `SKIP_UPDATE` / `AUTO_UPDATE_DISABLED` bails.** Removing a file the checkout should have deleted is working-tree repair, not an update. Users with auto-update off previously jumped straight past any cleanup and stayed stuck indefinitely.

All three launchers shared the same stash-and-restore cycle, so all three are patched — the bug was not Windows-specific in mechanism, only in trigger.

Safety of the scope:

- Pinned to the three source trees, so user data, `.env`, config, and `node_modules` are untouched.
- No `-x`, so gitignored build output survives.
- No `-q`, so git prints `Removing <path>` when it actually deletes — a stray file of your own leaves a trace in the console instead of vanishing silently.
- The in-app clean is wrapped in try/catch and logs a warning. A still-locked file must not fail the whole update, since the launcher retries the same clean on the next start.

A full `git clean -fd` would have been too aggressive (it could take `.env` and data files); the `ls-tree`-diff alternative from the issue solves the same case with considerably more code.

## Verified

- `pnpm check` passes.
- `bash -n` on both shell launchers.
- Git semantics exercised in a scratch repo on Linux: the scoped clean removes an untracked stale file, keeps tracked files, keeps gitignored build output, leaves `data/` and `.env` alone, and exits 0 when a pathspec doesn't exist. `git clean -nd` against this repo lists nothing, so the change is a no-op on a healthy tree.

## Not verified — needs a human on Windows

**None of this has been run on Windows, and the bug is Windows-triggered.** `start.bat` has no automated test lane, so the ordering claim (clean runs before `stash push -u`) is read off the control flow rather than observed. Please confirm before merging.

## Test plan

- [ ] Fresh install of a release build on Windows (zipfile or NSIS installer).
- [ ] Switch to the staging channel via the in-app updater; manually verify the build succeeds.
- [ ] Reproduce the locked-file case (hold `packages/shared/src/chat-mode-capabilities.ts` open during the switch) and manually verify the next `start.bat` launch removes it and builds.
- [ ] Manually verify `start.bat` prints `Removing <path>` when it deletes a leftover, and prints nothing on a clean tree.
- [ ] Manually verify a launch with `AUTO_UPDATE_ENABLED=false` still clears a stale file.
- [ ] Manually verify an untracked file under `packages/*/src` that a developer created is reported before deletion (expected behaviour change — flag if too aggressive).
- [ ] Manually verify on Linux via `start.sh` that a normal launch and update are unaffected.
- [ ] Manually verify on Termux via `start-termux.sh`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically removes untracked source files during startup and after switching update channels.
  * Cleanup runs whether automatic updates are enabled, skipped, or completed.
  * Cleanup failures no longer interrupt application updates or startup.
  * Applies only to Git-based installations; removal activity remains visible where supported.


Gunterlie:
@coderabbitai Both findings verified against the current code and fixed in 01b941d.

**1. `getManualGitApplyCommand` — valid, fixed.** Confirmed the bypass is real: when `UPDATES_APPLY_ENABLED=false`, the 403 handler hands back a `manualUpdateCommand` that ran `fetch → checkout → install → build` with no cleanup, so the disabled-apply path walked straight into the same TS2353 failure the in-app path now avoids. The command now runs the same scoped clean between checkout and install, and the path list moved into a `STALE_SOURCE_CLEAN_PATHS` constant shared with `cleanStaleSourceFiles` so the two can't drift.

The staging variant of `checkoutCommand` is an unparenthesized `A && (B) || C` chain, so inserting a segment after it is precedence-sensitive. Checked rather than assumed — the clean runs on both the existing-branch and new-branch paths, and the install still doesn't run when the checkout fails:

```
false && (checkout-existing) || checkout-new && CLEAN && pnpm-install   → checkout-new, CLEAN, pnpm-install
false && (x) || false && CLEAN && pnpm-install                          → nothing, exit=1
```

**2. Cleanup as a stash prerequisite — valid, fixed.** Real gap: when the clean failed (the locked-file case this PR exists for), the launchers fell through to `stash push -u`, which captured the leftover and restored it after the update — recreating exactly the self-perpetuating loop the PR is meant to break. All three launchers now capture the clean's exit status and route a failure into each script's existing skip-update path, preserving the per-platform status handling: `SKIP_UPDATE_FOR_LOCAL_CHANGES` in `start.sh` / `start-termux.sh`, `STASH_FAILED` in `start.bat`.

Two notes from implementing it:

- The rebase onto the latest `staging` collided with the new `protect-launcher-data.mjs` snapshot guard, which lives in the same position and sets the same variable. Both guards are kept — the snapshot runs first, and the clean gate is chained behind `[ "$SKIP_UPDATE_FOR_LOCAL_CHANGES" != "1" ]` so neither masks the other.
- `start.bat` needed no equivalent change: its snapshot guard exits via `goto :skip_update` before control reaches the stash gate.

Nothing was skipped; both findings held up.

One consequence worth a maintainer's eye: a clean that fails for any reason now suppresses the auto-update for that launch. That's the intended trade — a skipped update is recoverable, a stashed-and-restored stale file was not — but it does mean an undeletable stray file under `packages/*/src` stops updates until it's cleared. On Windows that path prints two warnings, the specific one plus the pre-existing generic "Could not stash local changes", which reads slightly off; I left the shared status handling alone rather than reword it.

`pnpm check` passes on the rebased branch. The Windows verification in the test plan is still outstanding and still the thing that actually matters here — none of this has been exercised on Windows, which is where the bug triggers.

PR #4424, opened 2026-08-01
AFTER UPDATE: docs(noodler): spec the Slice 8f access model and onboarding overhaul

Gunterlie:
## Why

Docs-only layer, split out so the ~2100 lines of planning prose do not sit in the same review as the code that implements it.

Updates the NoodleR living plan with the current slice order and status, adds the access-model and onboarding design document that Slice 8f's detail hangs off, and updates the user-facing noodle settings doc.

## Stack

Layer 2 of 4. Base: `stack/1-admission` (#4423).

## Notes

- Zero code change, so CI here proves only that nothing else moved.

## Test plan

- [ ] Read the living plan's status table against what actually lands in layers 3 and 4
- [ ] Confirm `docs/noodle/settings.md` matches the settings surface shipped in layer 4
- [ ] Confirm whether the `docs/` change needs a `docs-i18n` follow-up issue per CONTRIBUTING § Translated documentation


Gunterlie:
Folded into #4426, which now targets \`staging\` directly.

This spec describes the access model that #4425 and #4426 implement, and those two cannot land separately, so keeping the docs as their own PR only meant staging would document something not yet built. Reviewers see the spec and the implementation together in #4426.

No code is lost: \`big-chungus-1\` already contains this branch. The branch stays for reference.

PR #4425, opened 2026-08-01
AFTER UPDATE: feat(noodler): collapse post access to public|locked, add creator replies and reserve

Gunterlie:
## Why

The NoodleR access model had grown three overlapping concepts — `subscriber`, `ppv`, `ppvPrice`, `subscriptionIncludesPpv` — that users experienced as one question: can this viewer read this post. This collapses post access to `public | locked` with a data migration, and builds the two features that depend on that settled model: creators replying to the viewer, and a front-loaded reserve of scheduled posts so a creator's feed does not go silent when generation is slow.

Carries Slices 8f-1, 8f-2, 8g, and 8f-3 from the living plan.

## Why these four slices are one PR

The living plan orders them as separate slices, and that was the right *build* order. They cannot be separated *after the fact* at file granularity: all four turn on the same access and settings type collapse, which ripples through `noodle.routes.ts`, the autopost scheduler, and the generation service. Splitting them here would mean cutting those three files hunk-by-hunk and hand-repairing each intermediate state so it compiles — real risk of subtly wrong intermediate commits, for no reviewer benefit.

The client surface for this work is deliberately held back to layer 4.

## Stack

Layer 3 of 4. Base: `stack/2-docs` (#4424).

## Notes

- `noodle-autopost-cadence.ts` is deleted; the reserve operation replaces it.
- Includes a migration for existing `noodle_posts` rows off the retired access values. Reviewers should look hardest here — it rewrites stored user data on first boot.

## Test plan

- [ ] `pnpm check`
- [ ] `pnpm regression:noodle` (15 regressions, including the new `noodle-creator-reply` and reserve coverage)
- [ ] Manually verify the migration against a pre-8f data directory: existing `subscriber`/`ppv` posts land as `locked`, and no post becomes readable that was not readable before
- [ ] Manually verify a locked post's media stays inaccessible to a non-subscribed viewer, including by direct media URL
- [ ] Manually verify creator replies appear for the addressed viewer and respect access rules
- [ ] Manually verify the reserve refills and that a paused creator stops consuming it


Gunterlie:
Folded into #4426, which now targets \`staging\` directly.

The access-model change here edits \`packages/shared\` types and schemas that the client in #4426 consumes, so merging this alone would fail \`pnpm check\` on staging. Server and client are one change and have to land together.

No code is lost: \`big-chungus-1\` already contains every file from this branch. The branch stays for reference.

PR #4426, opened 2026-08-01
AFTER UPDATE: feat(noodler): guided creator onboarding and access client surface

Gunterlie:
## Why

NoodleR had no way in. A user enabling it landed on an empty surface with no explanation of what the product is, and creator setup exposed every dial at once. This adds the guided onboarding wizard with an emulated Professor Mari teaching post, an explicit opt-in gate that explains NoodleR before asking, and the client surface for the access model, creator replies, and reserve that landed in layer 3.

Carries Slice 8f-4 from the living plan, plus the client half of 8f-2, 8g, and 8f-3.

## Stack

Layer 4 of 4 (top). Base: `stack/3-server` (#4425).

Merge bottom-up: #4423 → #4424 → #4425 → this.

## Notes

- No linked issue exists for this work; one should be opened before this is taken out of draft.
- This layer is large (19 files, ~4400 insertions) and is almost entirely UI. It resisted further splitting because the wizard, the opt-in gate, and the settings reorganization share the same navigation state.
- Merging `staging` in brought a viewport-breakpoint → container-query migration in the noodle tree (`lg:` → `@min-[1024px]:`, `xl:` → `@min-[1280px]:`). Our new noodle components were converted to match. **This is the highest-risk part of this PR for visual regressions** and is not covered by any automated check.
- `ko.json` had 38 keys orphaned by the key cleanup on this branch; they were pruned, and one translation was repaired to preserve its `{{amount}}` interpolation.

## Test plan

- [ ] `pnpm check`
- [ ] `pnpm regression:noodle`
- [ ] `pnpm smoke:ui`
- [ ] Manually verify the full first-run path in browser: opt-in gate → wizard → teaching post → first real creator
- [ ] Manually verify desktop and mobile layouts at the 1024px and 1280px boundaries, since those breakpoints were rewritten as container queries
- [ ] Manually verify the NoodleR tab and settings section stay hidden when NoodleR is disabled
- [ ] Manually verify Korean UI renders correctly on the noodle surfaces after the locale prune
- [ ] Manually verify light and dark themes, plus loading, empty, and error states on the wizard

PR #4503, opened 2026-08-03
feat(noodler): Slice 8f access model, creator replies, reserve, and guided onboarding

Gunterlie:
## Why

NoodleR had no way in. A user enabling it landed on an empty surface with no explanation of what the product is, and creator setup exposed every dial at once. This adds the guided onboarding wizard with an emulated Professor Mari teaching post, an explicit opt-in gate that explains NoodleR before asking, and the client surface for the access model, creator replies, and reserve.

Carries Slice 8f from the living plan: the spec (8f-1), the server access model, creator replies and reserve (8f-2, 8f-3, 8g), and the client onboarding surface (8f-4).

## Why this is one PR

This started as a four-PR stack. The split was drawn along file-type lines rather than along "can this land on its own", which is the wrong axis:

- The server layer collapses post access to `public|locked` in `packages/shared` types and schemas, and the client layer consumes exactly those types. Merging the server half alone fails `pnpm check` on staging. They are one change.
- The docs layer specs the model those two implement, so on its own it would have documented something staging had not built yet.

Those three are now merged here and their PRs (#4424, #4425, #4426) closed as folded in. The connection-admission fix stayed independent as #4423 and is reviewed separately.

## Relationship to #4423

This branch contains #4423's commits, because it was built on top of that layer and has been kept merged up with it. **Merge #4423 first.** Once it lands, its changes disappear from this diff and what remains is the Slice 8f work.

## Notes

- This is large (~7000 lines across server, shared, client, docs and regressions) and the client half is almost entirely UI. It resisted further splitting because the wizard, the opt-in gate, and the settings reorganization share the same navigation state.
- Merging `staging` in brought a viewport-breakpoint to container-query migration in the noodle tree (`lg:` to `@min-[1024px]:`, `xl:` to `@min-[1280px]:`). The new noodle components were converted to match. **This is the highest-risk part of this PR for visual regressions** and is not covered by any automated check.
- `ko.json` had 38 keys orphaned by the key cleanup on this branch; they were pruned, and one translation was repaired to preserve its `{{amount}}` interpolation.
- Docs under `docs/` change here, so the `docs-i18n` language packs need matching updates or a `[docs-i18n] <paths>` follow-up issue before this merges.

## Validation already run

On the current head, after merging staging: `pnpm check`, `pnpm regression:providers` and `pnpm regression:noodle` all pass. The boxes below are for the human contributor and are intentionally left unchecked.

## Test plan

- [ ] `pnpm check`
- [ ] `pnpm regression:noodle`
- [ ] `pnpm smoke:ui`
- [ ] Manually verify the full first-run path in browser: opt-in gate to wizard to teaching post to first real creator
- [ ] Manually verify desktop and mobile layouts at the 1024px and 1280px boundaries, since those breakpoints were rewritten as container queries
- [ ] Manually verify the NoodleR tab and settings section stay hidden when NoodleR is disabled
- [ ] Manually verify Korean UI renders correctly on the noodle surfaces after the locale prune
- [ ] Manually verify light and dark themes, plus loading, empty, and error states on the wizard
- [ ] Manually verify locked vs public posts, creator replies, and reserve end to end against a real connection


Gunterlie:
Closing in favour of #4515, re-cut as a fresh branch off `staging`. Same content; the admission half is now reviewed independently in #4514. Merge #4514 first.

PR #4514, opened 2026-08-03
fix(generation): centralize connection admission and provider fallback

Gunterlie:
## Why

Automatic work (NoodleR autopost, the timeline refresh scheduler) shares API connections with
whatever the user is generating right now. Nothing coordinated the two, so a scheduled run could
start on a connection mid-conversation and compete for the same rate limit, parallel-job budget,
or local model. This adds one admission layer that every generation path routes through:
foreground work always proceeds, background work yields to it and waits out a short idle window.

Fallback and cleanup came along because they are the same problem: an admitted slot that is never
released is worse than no admission at all, and a fallback that resolves a logical attempt must
report that attempt's real outcome.

## What changed

- **Admission core** (`services/generation/connection-admission.ts`): per-connection foreground
  count and background flag, a 30s post-foreground idle window, and a provider decorator that
  holds a slot for the duration of a chat, completion, or embed call.
- **Background producers**: the NoodleR autopost scheduler and the automatic timeline refresh now
  run as background. A busy connection is a deferral, not a failure — the refresh route answers
  409 so the scheduler retries shortly instead of backing off as a configuration error.
- **Stream cleanup**: the main generation route and the fallback provider close their generators
  on abort, disconnect, and token-write failure, so slots are always released.
- **One attempt across fallback**: a background attempt is booked once on the primary and
  finalized once after the primary-plus-fallback chain settles, so a successful fallback is not
  recorded as a failed attempt. Admission rejections are rethrown rather than triggering fallback.
- **Image admission keys** are built from the physical endpoint: RunPod endpoint ids stay
  separate, and the origin/`/v1` spellings of one OpenAI-compatible endpoint converge.

## Linked issue

No issue is linked. One should be opened for this before merge.

## Automated validation (run locally, passing)

- `pnpm check`
- `pnpm regression:providers` — admission rejection, fallback skip on admission failure, attempt
  outcomes across a successful and a failed fallback, stream abandonment releasing the slot,
  image admission key convergence and separation.
- `pnpm regression:noodle`

## Manual verification still needed

Automated coverage is unit/regression level only — no test exercises a real scheduler racing a
real user generation against a live provider. Please verify by hand:

- [ ] Start a long streaming generation on connection A, then trigger the NoodleR autopost
      scheduler for a creator on connection A; confirm the automatic post is deferred (logged as
      `[noodle-autopost] Deferred ...`) rather than running concurrently.
- [ ] Repeat with the automatic timeline refresh; confirm the scheduler retries in ~60s instead
      of recording a configuration failure and backing off for minutes.
- [ ] Confirm a manual "Refresh NoodleR now" and a manual timeline refresh still run immediately
      while nothing else is generating (manual paths must stay foreground).
- [ ] Abort a streaming generation mid-response, then start scheduled work on that connection
      after the idle window; confirm it is admitted (no stranded foreground slot).
- [ ] Run a dry run against the local sidecar while a local-sidecar generation is active; confirm
      behavior is unchanged for the user.
- [ ] Generate an image on an OpenAI-compatible connection configured with the bare origin, and
      again with the `/v1` form; confirm both still generate successfully.

## Known gaps

- Only the Noodle schedulers produce background mode. Local-sidecar consumers (capability LLM,
  Professor Mari workspace, memory recall, lorebook vectorization, and three more) still run
  unadmitted; tracked as #4520. Follow-up adoption, not regressions.

## Summary by CodeRabbit

- **New Features**
  - Added connection-aware generation controls to LLM, embedding, and image requests.
  - Prevented overlapping foreground work and deferred background generation when connections are busy.
  - Added clear conflict responses when generation cannot start immediately.
  - Applied admission handling across Noodle generation, scheduled posts, refreshes, and image retries.

- **Bug Fixes**
  - Improved cleanup of interrupted streaming responses.
  - Preserved original generation errors during stream cleanup.
  - Prevented connection-limit failures from triggering unnecessary fallback or retries.


Gunterlie:
Thanks @xelvanis for the thorough review — the findings were specific enough to act on directly.

Pushed 7f4bb3d, which addresses F1–F6:

- **F1** — NoodleR autopost and the automatic timeline refresh now pass background admission through the text and image paths. A busy connection is a logged deferral; the refresh route answers 409 so the scheduler retries shortly instead of backing off as a configuration failure. Manual refreshes stay foreground.
- **F2** — the main streaming route closes its generator in a `finally`, so abort, disconnect, and token-write failures release the foreground slot.
- **F3** — the dry-run local-sidecar path now goes through the admission wrapper. The other four sidecar consumers are listed as a known gap in the PR body rather than fixed here, to keep this diff reviewable.
- **F4** — one logical attempt now spans primary plus fallback: booked once on the primary, finalized once after the chain settles, so a successful fallback is recorded completed. A regression asserts both the successful-fallback and total-failure outcomes.
- **F5** — OpenAI-compatible image admission keys resolve to the request endpoint, so origin and `/v1` converge while distinct hosts and RunPod endpoint ids stay separate.
- **F6** — added the fallback-skip regression for both `chat` and `chatComplete`; verified it fails when the `isConnectionAdmissionFailure` guard is removed.
- **F7** — PR body rewritten with the rationale, the missing-issue note, and an explicit manual-validation list.

Three CodeRabbit inline comments I did not act on, with reasons:

- *Iterative unwrap including `ConnectionDefaultProvider`* — the only caller of `unwrapConnectionAdmissionProvider` never passes a `connectionId` to `createLLMProvider`, so no admission layer exists there; the `ConnectionDefaultProvider` naming imprecision predates this PR.
- *`ConnectionAdmissionProvider.chat` reporting completed on early close* — the paths that now close early are aborts, which `failed` describes correctly.
- *Idempotent admission wrapping* — `createLLMProvider`'s `connectionId` argument is used at exactly one call site, which never feeds `withConnectionFallbackProvider`, so no double wrap exists to guard against.

`pnpm check`, `pnpm regression:providers`, and `pnpm regression:noodle` pass locally. The concurrency behavior itself still needs the manual checks listed in the description — I have not exercised a live scheduler-versus-user race.

Gunterlie:
@coderabbitai — addressed one of the two outside-diff comments, skipped the other.

**Fixed: connection-admission failures swallowed into image-failure metadata.** Valid, and it defeated the change this PR is about. `prepareGeneratedNoodleMedia` and the NoodleR immediate-image path both caught every error and persisted `imageGenerationFailed` metadata. Once background admission is in play, a rejection there means "the image connection is busy with user work" — a deferral, not a failure — but the post was persisted permanently branded as image-failed and the run reported success. Both catch blocks now rethrow when `isConnectionAdmissionFailure(err)` is true, before any failure metadata is set; every other provider error keeps its existing text-only fallback. The rejection reaches the autopost scheduler (logged deferral, retried next cadence) or the refresh route (409, scheduler retries shortly).

The preview path (`previewOnly: true`) needed no guard — it returns before `generateImage`, so no admission is ever taken.

**Skipped: extracting a shared `closeAsyncGenerator` helper.** The duplication is real but it is one `await gen.return(undefined).catch(...)` line per site, with different log prefixes and different explanatory comments. A helper in `base-provider.ts` would wrap a single stdlib call and add an export plus an indirection for two callers, without removing a branch or a failure mode. Happy to add it if a third site appears.

Validation: `pnpm check`, `pnpm regression:providers`, `pnpm regression:noodle` all pass. No regression added for the rethrow guards — exercising them needs a live image connection under contention, so it is on the manual-verification list in the description.

Gunterlie:
Pushed 5d35faf.

## F1 — already fixed, before this review

Both automatic image boundaries got the `isConnectionAdmissionFailure` rethrow in dd1443ae (pushed with the previous round): `noodle-generated-activity.service.ts:175` and `noodle-noodler-generation.service.ts:401`. The review looks to have been run against b63cf27 or earlier. No change needed — but it did surface one adjacent hole, now fixed:

`generateNoodleImageWithRetry` treated the rejection as a transient provider fault and retried it after a 500ms sleep before finally rethrowing. The deferral still happened, but late and with a misleading "attempt 1/2 failed" warning. It now rethrows admission failures immediately.

## F4 — you were right on both counts; fixed properly this time

My previous fix moved the finalizer out to the chain but still inferred the outcome from whichever leg finalized last. That was the actual defect, and it explains both gaps:

- **Text**: an empty-but-successful primary is a `completed` leg inside an attempt that delivered nothing. The fallback's admission rejection is then raised *between* legs, where no leg finalizer runs at all, so `completed` survived as the recorded outcome.
- **Image**: `generateImage` was never converted — it still admitted the primary with the caller's full mode, so the primary's finalizer fired `failed` before the recursive fallback call started.

The fix is the same one in both places: no leg's outcome is trusted, because no leg knows the chain's result. `splitConnectionAttemptAcrossFallback` now hands the primary a no-op leg finalizer, gives the fallback a slot-only mode that books nothing, and `settle(outcome)` takes the outcome from the caller driving the chain — `ConnectionFallbackProvider.chat`/`chatComplete` and `generateImage` each report `completed` only if their own call returned.

## New regressions

Both mutation-verified (confirmed failing against the unfixed code, not just passing against the fixed code):

- Empty primary → refused fallback records `failed`, on the streaming and `chatComplete` paths. Fails with `total failure: the attempt must be finalized once as failed` if the chain outcome tracking is broken.
- Successful image fallback over two local HTTP backends: asserts the fallback image is returned, exactly one booking, and outcome `completed`. Fails with `a successful image fallback must be recorded completed` if `generateImage` stops reporting the fallback's success.

`pnpm check`, `pnpm regression:providers`, `pnpm regression:noodle` pass.

Still not covered end to end, and still on the manual list: a live scheduler racing a live user generation. The admission accounting is now regression-covered at both seams, but the contention itself is not.

Gunterlie:
Valid — fixed in 4d8fcb6.

Confirmed against current code: `chat`'s `finally` reported the outcome from whether `yield*` returned normally, so a consumer closing the stream after reading usable tokens settled the attempt `failed` even though the generation delivered.

One deviation from the suggestion: I track delivery in the outer generator instead of reusing `chatChain`'s `emittedUsableOutput`. That flag only watches the primary leg, so reusing it would still mis-settle a chain where the *fallback* delivered the tokens and the consumer walked away. The outer loop sees everything the chain yields. It also has to close the chain explicitly, since a manual loop does not forward an early return the way `yield*` does — the same idiom already used one level down.

Streams closed before any usable token still settle `failed`.

Two regressions, both mutation-verified against the unfixed code:

- close after a usable chunk → `completed`, and the primary's slot is still released;
- close after a whitespace-only chunk → `failed`.

Worth recording from writing those: a stream closed before its first `next()` never runs its body at all, so it books nothing and admits nothing — there is no accounting to assert on that path, and the second regression advances once on purpose to reach it. Noted in the test.

Skipping `pnpm check` locally as instructed — typecheck and `pnpm regression:providers` pass; leaving the rest to CI.

PR #4515, opened 2026-08-03
feat(noodler): Slice 8f access model, creator replies, reserve, and guided onboarding,

Gunterlie:
## Why

NoodleR had no way in. A user enabling it landed on an empty surface with no explanation of what the product is, and creator setup exposed every dial at once. This adds the guided onboarding wizard with an emulated Professor Mari teaching post, an explicit opt-in gate that explains NoodleR before asking, and the client surface for the access model, creator replies, and reserve.

Carries Slice 8f from the living plan: the spec (8f-1), the server access model, creator replies and reserve (8f-2, 8f-3, 8g), and the client onboarding surface (8f-4).

## Why this is one PR

This started as a four-PR stack. The split was drawn along file-type lines rather than along "can this land on its own", which is the wrong axis:

- The server layer collapses post access to `public|locked` in `packages/shared` types and schemas, and the client layer consumes exactly those types. Merging the server half alone fails `pnpm check` on staging. They are one change.
- The docs layer specs the model those two implement, so on its own it would have documented something staging had not built yet.

Those three are now merged here and their PRs (#4424, #4425, #4426) closed as folded in. The connection-admission fix stayed independent and is reviewed separately.

## Relationship to #4514

#4514 is merged. `staging` was merged in here, so its commits no longer appear in this diff and
what remains is the Slice 8f work. Standalone against `staging` — not a stack.

## Notes

- Implements #4525.
- This is large (~7000 lines across server, shared, client, docs and regressions) and the client half is almost entirely UI. It resisted further splitting because the wizard, the opt-in gate, and the settings reorganization share the same navigation state.
- Merging `staging` in brought a viewport-breakpoint to container-query migration in the noodle tree (`lg:` to `@min-[1024px]:`, `xl:` to `@min-[1280px]:`). The new noodle components were converted to match. **This is the highest-risk part of this PR for visual regressions** and is not covered by any automated check.
- `ko.json` had 38 keys orphaned by the key cleanup on this branch; they were pruned, and one translation was repaired to preserve its `{{amount}}` interpolation.
- `docs/noodle/settings.md` changes here and the ten translated copies now describe the removed subscriber/PPV model. Tracked as #4524.

## Validation already run

On the current head, after merging staging: `pnpm check`, `pnpm regression:providers` and `pnpm regression:noodle` all pass. The boxes below are for the human contributor and are intentionally left unchecked.

## Test plan

- [x] `pnpm check`
- [x] `pnpm regression:noodle`
- [x] `pnpm smoke:ui`
- [x] Manually verify the full first-run path in browser: opt-in gate to wizard to teaching post to first real creator
- [x] Manually verify desktop and mobile layouts at the 1024px and 1280px boundaries, since those breakpoints were rewritten as container queries
- [x] Manually verify the NoodleR tab and settings section stay hidden when NoodleR is disabled
- [ ] Manually verify Korean UI renders correctly on the noodle surfaces after the locale prune
- [ ] Manually verify light and dark themes, plus loading, empty, and error states on the wizard
- [x] Manually verify locked vs public posts, creator replies, and reserve end to end against a real connection

## Summary by CodeRabbit

* **New Features**
  * Added guided Noodler onboarding for account selection, privacy, activity, images, and review.
  * Introduced Public and Locked posts with subscriptions or coin-based unlocks.
  * Added creator discovery, following, search, creator replies, locked previews, and refreshed Viewer Hub navigation.
  * Added reserve-based automatic posting, publishing status, creator refresh controls, responsive navigation, localization, accessibility improvements, and new visual assets.

* **Documentation**
  * Updated settings documentation for simplified access and publishing.

* **Bug Fixes**
  * Improved identity protection, legacy access migration, duplicate handling, media privacy, backup/import consistency, and image cleanup.


Gunterlie:
@coderabbitai — worked through all 29 findings against the post-merge code. Pushed 9567c2c: **24 fixed, 5 skipped**. `pnpm check`, `pnpm regression:noodle` and `pnpm regression:providers` pass.

## Fixed

**Server**

- **Reserve picker (`noodle-noodler-reserve.operation.ts:72`)** — both halves confirmed. `Date.parse("0")` returns `946681200000` (1999-12-31 local), not `0`; verified in node. Real zero now, and the per-creator reads fan out instead of running one round trip at a time.
- **Promoted media leak (`:170`)** — the image is promoted out of staging before the row is written, so a persist failure orphaned the file. Same cleanup the missed-window path already did.
- **Deleted auto-post republished (`noodle.storage.ts:1761`)** — the most consequential one. Resetting a `published` row to `prepared` keeps its now-past `publishAt`, and `publishDueNoodlerPreparedPosts` republishes it within 60s. Re-preparing is now restricted to rows whose slot is still ahead; the rest are discarded.
- **Manual create returning 500 on cleanup failure (`noodle-noodler-post.operation.ts:186`)** — the post already exists at that point, so failing the request invited a retry that posts twice. Best-effort with a warn.
- **Bulk wizard replay (`noodle.routes.ts:862`)** — the account row and its scheduler settings are two writes, so a retry after a failure between them reported a creator as created that never received its config. Auto-posting is now applied on both replay paths.
- **Creator-reply statuses (`noodle.routes.ts:504`)** — `exhausted` → 429, `ineligible` → 404. `duplicate` stays a success: it carries the interaction the caller wanted. This also resolves the `use-noodle.ts:424` finding — the existing `catch` in `submitReply` now fires and toasts the server message, with no new i18n keys needed in every locale.
- Also: targeted-refresh id dedupe, startup reconciliation tracked in `running`, and the stale "subscriber access" comment.

**Client**

- **`settingsSection` clamp (`NoodleHome.tsx:739`)** — confirmed: a persisted `creators` section plus the tab falling back to `noodle` leaves no matching chip and an empty settings body.
- **Age gate (`:52`, `:86`)** — the media query is live, so reduced motion turning on mid-animation tore down the interval before `charged` was ever set, disabling Enter permanently. Enter is also now a one-way door and clears its timer on unmount.
- **Wizard (`:163`, `:218`, `:851`)** — the 30s bootstrap refetch was overwriting the activity choice; Retry's rejection left the completion screen untouched with no toast; `postsPerDay` now rounds, since a `4.5` was rejected server-side and surfaced as a generic wizard failure.
- **Reserve status polling, schedule-toggle disabled until settings load, profile-initial diameter, mobile Profile entry, and the lowercased translated label.**

On the label: I dropped `toLowerCase()` rather than adding dedicated keys. Your reasoning about German nouns and the Turkish dotted/dotless I is right, and the label reads correctly untransformed in every locale — adding keys would mean touching every language pack for an aria-label.

## Skipped

- **`noodle.routes.ts:441` — unauthenticated media links.** The premise does not hold. `app.ts:177` applies `basicAuthHook` as a global `onRequest` hook, so every request to this server is already authenticated the same way. There is no per-user identity: personas are in-fiction actors, not principals, and the subscription/unlock model is a roleplay device rather than a security boundary. The no-persona branch is the owner path, the same trust surface as every other `/noodler/accounts` route, which the comment above it already documents.
- **`NoodlerBulkCreatePanel.tsx:289` — fresh `executionId` on retry.** The idempotency lookup is `getNoodlerPostByWizardExecution(accountId, executionId)`, so it only matches creators that already produced a post. Retry passes `failedIds`, which by definition produced none, so the lookup misses and generation runs again. Retries do recover.
- **`noodle.schema.ts:452` — content requirement on the media schema.** `decodeNoodlerMediaRequest` selects `withMedia` only when `media` is present (line 246: `media ? schemas.withMedia : schemas.withoutMedia`). A request validated by that schema always has an image, so the "body, image, or poll" requirement is already satisfied — by the uploaded file, which `uploadedImageUrl` does not represent.
- **`noodle-noodler-generation.service.ts:286` — duplicate linked-account read.** Real, but it is one local SQLite read next to an LLM call, and removing it means threading a new shared identity builder through both call sites. Not worth the coupling.
- **`image-generation.ts:317` — the `ponytail:` marker.** Deliberate repo convention, not an ad-hoc tag: these markers record a known ceiling with its upgrade path and are harvested into a debt ledger. Left as is.

## One thing to look at

The profile-initials fix changes a rendered size (`h-16 w-16` → `h-24 w-24`) to match `Avatar size="lg"`. It removes a layout shift between creators with and without a picture, but it is a visual change in a PR whose stated top risk is visual regression, so it wants a human eye rather than a green check.

No regression covers the fixes themselves — the server ones need a live scheduler and a real connection, and the client ones are UI state. They are on the manual list in the description.

Gunterlie:
Follow-up on the skipped findings, and the loose ends this PR was carrying.

**Fixed the one I called real and skipped anyway** (54c1a29): `generateNoodlerPost` read the linked public account and then called `resolveNoodlerPublicIdentity`, which read the same row again. Both paths now go through `noodlerPublicIdentityFor`, which takes an already-resolved account. I had waved this off as "one local read next to an LLM call" — that is a reason it is cheap, not a reason it is correct.

**The other four stay skipped, because there is nothing to fix.** Restating the evidence rather than the verdict:

- `noodle.routes.ts:441` — `app.ts:177` registers `basicAuthHook` as a global `onRequest` hook. Every request is already authenticated identically; personas are in-fiction actors, not principals. Gating the owner path would remove access to the user's own management surface and buy no boundary that does not already exist.
- `NoodlerBulkCreatePanel.tsx:289` — the idempotency lookup keys on `(accountId, executionId)` and only matches accounts that already produced a post. Retry passes `failedIds`, which produced none.
- `noodle.schema.ts:452` — `decodeNoodlerMediaRequest` line 246 picks `withMedia` only when `media` is present, so a request validated by that schema always carries an image.
- `image-generation.ts:317` — repo convention with tooling behind it, not a stray tag.

**Closed the two loose ends the description was carrying as intentions:**

- #4525 — the missing linked issue for this work. `CONTRIBUTING.md` asks for one; the body said "should be opened" and then this PR left draft without it.
- #4524 — the `[docs-i18n]` follow-up. This is worse than an untranslated doc: `docs/noodle/settings.md` collapses access from `public|subscribers|ppv` to `public|locked`, so all ten translated copies now instruct readers to choose between Subscribers and PPV and to find a "Subscriptions include PPV" toggle that no longer exists. I filed it rather than machine-translating substantive prose into ten shipped docs.

Both are linked from the description now. `pnpm check` and `pnpm regression:noodle` pass.

Gunterlie:
@coderabbitai Valid, and it was mine — introduced with the replay fix in 9567c2c. Fixed in efc6e5bc.

You are right about the mechanism: that branch already sits inside the outer `catch`, so an unguarded throw from `applyAutoPosting` walks past the `for` loop and out of the handler, turning one creator's failed settings write into a 500 for the whole batch. It now uses the same try/catch the `existing`-account replay branch has, marking only that creator as failed.

I took the diff rather than the committable suggestion — the suggestion block was anchored to the wrong hunk and would have replaced the non-race branch.

`pnpm check` and `pnpm regression:noodle` pass.

Gunterlie:
Thanks for the review, @xelvanis — this was a genuinely useful pass. Findings are addressed as follows.

### Fixed

| Finding | Change |
|---|---|
| F2 | `claimNoodlerCreatorReply` now runs its duplicate check *after* the eligibility and post-access checks, so replaying known IDs from a persona that does not own the parent comment returns `ineligible` instead of the stored reply. |
| F8 | The same move puts the duplicate check after orphan-expiry pruning, so a claim orphaned by a crash stops blocking its comment permanently. |
| F10 | Deleting a comment now deletes its creator reply, that reply's digests, and the claim, in the same transaction — via a shared helper used by both delete paths. |
| F3 | `noodler_prepared_posts` and `noodler_creator_reply_claims` are included in profile-import account-ID remapping. |
| F7 | Prepared slots more than an hour past their publish time are retired instead of published with backdated timestamps. |
| F11 | Reserve invalidation moved into `generateAndApplyNoodlerPost`, so it covers refresh-now and targeted refresh at the shared seam rather than one caller. |
| F23 | The 429 body now states the reply ceiling is installation-wide. |
| F14 | Selection is capped at a new shared `NOODLER_BULK_ACCOUNT_MAX` (the schema uses the same constant) with a visible limit notice. |
| F15 | A failed settings write no longer strands created profiles: generation still runs for them, and onboarding still reports as failed. |
| F16 | Publishing Settings renders loading and error/retry states instead of a false zero and a false "no creators yet". |
| F17 | All three publishing toggles surface a failure toast alongside the existing rollback. |
| F18 | Locked-post counters carry sr-only labels; the icons are `aria-hidden`. |
| F19 | "Create a persona" only renders after the personas query succeeds; the error state offers retry. |
| F20 | The empty creator page offers retry when the eligible-sources query failed. |
| F21 | An all-skipped refresh no longer reports "Generated 0 posts" as success. |
| F24 | Deleting a creator opened from Settings returns to Settings. |
| F25 | `docs/noodle/settings.md` gained a NoodleR automatic-publishing section and defaults rows. |

Regression coverage added for the changed server logic: parent-ownership revalidation, orphan-claim expiry, and reply/claim cascade deletion in `noodle-creator-reply.regression.ts`; elapsed-slot retirement in `noodle-autopost.regression.ts`.

### Not addressed, deliberately

- **F1, F5** — downgrade and migration semantics rather than defects. The schema already carries deliberate rollback mirrors, and whether to broaden legacy PPV access for existing subscribers is a maintainer decision, not something to settle in this pass.
- **F4, F6, F9** — the crash-window durability set (media/row ordering, backup consistency, reply-to-claim reconciliation). Each needs a recoverable operation ledger or export/scheduler coordination; that is a design change rather than a patch, and worth its own issue.
- **F12** — invalid reserve timestamps poisoning the scheduler; the quarantine-versus-reset behaviour needs a decision first.
- **F13** — the retry path exists (`retryFailed`), but reconciling "durably created yet reported as zero" after a lost response needs a server-side execution-ID lookup.
- **F22, F26, F27, F28** — activity-based refresh ordering, terminal prepared-row pruning, a real fake-provider image round-trip regression, and consolidating the duplicated disclosure prompt text.

Validation: `pnpm check` and the full `pnpm regression:noodle` lane pass locally.

Two notes for reviewers:

- F21's all-busy case uses the error toast, because `summarizeRefreshOutcomes` only distinguishes ok from not-ok. It is accurate but harsher in tone than "busy" deserves; a neutral tone would need a third state on that helper.
- The `docs/noodle/settings.md` change needs a `[docs-i18n]` follow-up per `CONTRIBUTING.md § Translated documentation`. Happy to open that issue if it is not already tracked.

Gunterlie:
Thanks again, @xelvanis — this second pass was right on every count, and my previous comment overstated what was closed. Correction and follow-up below; the work is in `a91973463`.

Of the 17 I claimed, 9 were genuinely closed (F2, F14, F17, F18, F19, F20, F21, F23, F24). The other 8 were partial, aimed at the wrong code path, or — in one case — introduced a regression. All eight are now addressed.

### What was actually wrong

| Finding | Diagnosis | Fix |
|---|---|---|
| F10 | The cleanup helper was attached to `deleteStoredInteraction`, which only ever receives like/repost. Comment deletion runs through `deleteInteractionById`. That path also refused the delete outright when a subtree contained a NoodleR-authored reply, because its "known account" set covered only the `noodle` platform. | Claim cleanup moved to `deleteInteractionById`, covering both claims a deleted comment owns and claims pointing at a deleted reply; the known-account set now spans both platforms. |
| F8 | Pruning used `claimedAt < cutoff` while budget membership uses `claimedAt > cutoff`, so a claim landing exactly on the 24-hour boundary was neither counted against the ceiling nor released. | Pruning is now the exact complement of the budget predicate. |
| F7 | Slots more than an hour late retired correctly, but slots inside the grace window still published stamped with their planned time. | A late publish carries the moment it actually published. |
| F3 | Account IDs were remapped, but the linked source ID embedded in each prepared post's `policyFingerprint` was not, so a collision import failed its own policy check and discarded the restored reserve on the next reconcile. | `remapNoodlerReservePolicyFingerprint`, exported from the module that owns the fingerprint format and applied during import. |
| F11 | Reserve-cleanup errors propagated after the post was already persisted, so a retry could post twice, and the execution-ID replay path skipped cleanup entirely. | Cleanup is logged and swallowed, and runs on the replay path too. |
| F15 | A failed settings write was reported as "First posts could not be generated", could report zero failures, and offered no settings retry. | Its own `settingsFailed` completion state with accurate copy and a retry that resumes from the real outcomes. |
| F16 | Aggregate states were fixed, but creator rows still asserted "No prepared post" while reserve status was unavailable. | The row subtitle waits for the status query and distinguishes loading from failure. |
| F25 | Night quiet, disabled-creator refresh outcomes, and the per-creator Automatic default were all described incorrectly. | Corrected: night quiet defers only **character**-linked creators (persona-linked ones can still fill a quiet-hours slot); creators with Automatic off are not in a refresh run at all rather than reported as skipped; per-creator Automatic and Images both default **off** outside guided setup. |

### Two things worth flagging

- The F10 regression I added previously called storage with `type: "reply"`, which the noodler remove-interaction command does not accept. It passed while proving nothing — exactly the false-positive shape the review warned about. It now drives post deletion, which is the deletion path NoodleR actually exposes.
- NoodleR comments have no delete route today: `replyManagement` is only supplied on the public timeline surface, so the NoodleR card's delete button has no handler, and `deleteInteractionById` cannot reach NoodleR interactions (`getInteractionById` requires a public post). Post and account deletion already cascade to claims. So the F10 cleanup is correct but latent rather than a live fix — it matters the moment that button gets wired.

### Validation

`pnpm check`, `pnpm regression:noodle`, `pnpm regression:issues`, `pnpm regression:backup`, and `pnpm regression:docs` pass locally. New coverage: expiry-boundary release, crash-orphaned reply adoption, post-deletion claim cascade, late-publish timestamps, elapsed-slot retirement, poisoned-timestamp recovery, terminal-row pruning, activity-based ordering, prepared-post fingerprint remapping, and a real image lifecycle round trip (`noodle-autopost-image-lifecycle.regression.ts`) to replace the schema-only smoke check.

The earlier deferrals — F4, F6, F9, F12, F13, F22, F26, F27, F28 — were also implemented in the meantime rather than left as issues. Still outstanding from my side: the `[docs-i18n]` follow-up for `docs/noodle/settings.md`.

Gunterlie:
@coderabbitai All five findings from the latest review were verified against current code and were still valid. All five are fixed in fc5ebb77d.

**Inline**
- `noodle-noodler-post.operation.ts` (~57-61): `logger.error` → `logger.warn` in the `discardPreparedPostsAfterManualPost` catch. Error object and message unchanged — the failure is deliberately swallowed because the post is already persisted.
- `noodle-creator-reply.regression.ts` (~204-208): the claim row is now looked up, asserted non-null, and its `id` passed to `releaseNoodlerCreatorReplyClaim`. The previous `?? ""` default meant the release path could silently no-op and still pass.

**Outside diff — `noodle.storage.ts`**
- (~963-981) `deleteInteractionChildren` now builds the full descendant set using the same closure loop as `deleteInteractionById`, then deletes `noodlerCreatorReplyClaims` matching `parentInteractionId` **or** `replyInteractionId` within that set before deleting the interactions. Activity-digest cleanup and direct-child deletion behavior are preserved. This required adding `or` to the `db/file-query.js` import.
- (~1622-1627) `claimNoodlerAutomaticAttempt` now applies the same NaN fallback as `ensureNoodlerReserveState` for `lastObservedBudgetTime`, so `toISOString` never receives an invalid date. `preparationNotBefore` is validated too: an unparseable value now returns `holding` instead of comparing `NaN` and skipping the hold. `ensureNoodlerReserveState` repairs the bad value, so the hold is not permanent.
- (~1907-1918) `reconcileNoodlerPreparedPosts` uses Sets for the invalid and discarded ids. All four membership checks — the `validFuture` filter, the prepared-media unlink loop, the missing-media skip, and the prune filter — use `has()`. The `discarded` array is retained only for the `inArray` calls, so discard contents and processing order are unchanged.

**Validation**
- `pnpm check` clean.
- Regressions pass: `noodle-creator-reply`, `noodle-deletion`, `noodle-autopost`, `noodle-autopost-image-lifecycle`, `noodle-scheduler`.

PR #4528, opened 2026-08-03
feat: let a capability package provide a whole Game mode

### LINKED ISSUE #4526: [Feature]: Let a capability package provide a whole Game mode

A package can extend Marinara in a lot of ways — a panel, a chat surface, a settings section, a
conversation game. What it cannot do is provide a **Game mode of its own**: one that draws its own HUD,
menus and combat, keeps its own state, and replaces the built-in systems it implements.

Building one surfaced four gaps, three of them small and independent of the main idea:

1. **`GameSurface` has no capability slot.** `view="surface"` exists only in `ConversationView`, for
   turn-games. There is no way for a package to render over the game area.
2. **The `prompt-context` permission has no consumer.** It is in the manifest enum, but nothing in the
   server reads it — declaring it does nothing. A package that owns live state has no way to tell the
   model about it, so the model narrates against state it cannot see.
3. **The resource facade is read-only.** A package's setup cannot create the player persona or a
   lorebook, which is the minimum a game mode needs to start.
4. **A package's own art is invisible until the next restart.** The asset scan runs at boot, before
   packages activate and install their assets, and nothing re-runs it. Anything reading the manifest —
   scene analysis, the asset pickers — cannot see that art for a whole session. This one affects any
   package that ships assets, not just game modes.

The motivating case is a Persona-style game mode (calendar, bonds, palaces, card combat).

## Proposed solution

Four separable pieces. Everything is declarative: the engine never names a package, an experience
DECLARES what it replaces, and anything left undeclared stays built-in. A game without an experience
behaves exactly as it does today.

**1. Re-run the asset manifest scan after packages activate.** The same idempotent scan the upload routes
already re-run, guarded so a package's freshly written files can't take down startup.

**2. A per-turn `prompt-context` hook.** Gives the existing permission a consumer: a package can append a
block to the system prompt of each turn. Contributors are keyed by package id, gated on the permission,
and a contributor that throws is logged and skipped so it can never break a turn. A contribution may also
declare which built-in systems it replaces (`provides`, today `{ inventory }`), and the GM prompt stops
driving those.

**3. Resource writes.** Six optional methods on `CapabilityResourceHost`, thin passthroughs to storage, so
a setup flow can find-or-create the player persona and its lorebook.

**4. A `game-surface` slot.** The experience is chosen when the game is CREATED, from an Experiences block
in the setup wizard, and belongs to that game for its lifetime — not a per-chat toggle, since an
experience owns the whole run. Which package it is comes from the manifest. The seams are declarative and
each is gated host-side: background, speaker avatars, active speaker, an in-flow choice slot, sprite
scale, and a chrome declaration by which the experience says which built-in chrome it replaces.

## Alternatives considered

**Doing it entirely package-side.** This was the starting assumption and most of it does work that way —
the state, the assets, the sprites, the combat, the storage all live in the package. These four are what
could not: there is no slot to render into, the format reminder is host-built, the resource facade is
read-only, and the manifest scan is host-side.

**Injecting per-turn state through a constant lorebook entry** instead of a prompt hook. Rejected after
measuring: that block is ~3-5k tokens on a typical turn, and a lorebook entry lands inside the lorebook's
own `tokenBudget` (15k, shared with every other entry), so it would compete with — and truncate — the
lore it sits next to.

**Using `gameSpecialInstructions`** for the same thing. It is injected every turn, but it is capped at
2000 characters, an order of magnitude short.

**Telling the model not to use the built-in inventory** from the package's own prompt, instead of the
`provides` declaration. The `[inventory: …]` command lives in the **format reminder**, which is pushed as
the last user message — closest to generation. A rule in the package's system prompt loses that argument,
and the model keeps logging items into a panel the player cannot see.

**Hiding built-in chrome with CSS** from the package's stylesheet rather than withholding the callbacks.
There are no stable selectors for those buttons, and hiding leaves the built-in flow reachable underneath
the surface.

**Registering the cast as tracked NPCs** (`chatMeta.gameNpcs`) so the narration would resolve avatars
without a new seam. Rejected: those NPCs are injected into the GM prompt as `<tracked_npcs>`, so a
25-character cast would pollute every turn and the persisted state.

**Enabling the experience as a per-chat agent**, which is how the first version worked. Changed on
@SpicyMarinara's direction, and it is better: toggling mid-run left a half-built game either way.

## Additional context

- Nothing here names a package: `grep` for any package id across the changed files returns zero.
- Validated on `staging`: `pnpm check`, `pnpm version:check`, `pnpm guard:installer-artifacts` and
  `scripts/test-owner-approval.mjs` all pass; `pnpm regression:prompt` passes 147 checks and leaves one
  failure that **reproduces identically on unmodified `staging`**.
- Two pre-existing problems I ran into and can file separately if useful: `pnpm regression:prompt` is
  currently red on `staging` for `Storyboard host consumes package-owned prompts without restoring Engine
  globals`; and `pnpm smoke:ui` cannot run where pnpm is the standalone binary, because
  `e2e/start-servers.mjs` assumes `npm_execpath` points at a JS file (with `@pnpm/exe` it points at
  `pnpm.exe`, so Node tries to parse the executable as a script).
- One open question for the PR: this adds API surface but leaves `supportedCapabilityApi` at 1.7, since
  bumping it felt like a maintainer call. Bumping would be backward compatible either way.

## Template check

SonaGat3:
> [!IMPORTANT]

## Linked issue

Closes #4526

## Why this change

Marinara can already be extended a lot — a package can add a panel, a chat surface, a settings
section. What it cannot do is provide a **Game mode of its own**: one that draws its own HUD, menus and
combat, keeps its own state, and replaces the built-in systems it implements. Everything a package can
contribute today sits *next to* Game mode rather than *being* one.

This adds that seam, plus the three smaller pieces of host support it turned out to need. All of it is
generic: the engine never names a package, an experience DECLARES what it replaces, and anything left
undeclared stays built-in. A game without an experience behaves exactly as it does today.

It's four separable pieces, described below so they can be reviewed one at a time.

-

### 1. Re-scan the asset manifest after packages activate — `packages/server/src/app.ts` (+12)

Unrelated to game modes, and useful to any package: a package can install its own art during
`activate()`, which runs *after* the boot-time asset scan. Until now that art stayed invisible to
everything that reads the manifest — scene analysis, the asset pickers — until the **next** restart.
This re-runs the same idempotent scan the upload routes already re-run.

Guarded with `try`/`catch` + `app.log.warn` on purpose: this call runs over files a package has just
written, and `scanDir` uses bare `readdirSync`/`statSync`. A stale manifest costs that package its art
until the next scan; failing to boot costs the user everything.

### 2. A per-turn `prompt-context` hook — 3 files + 1 new

- **new** `server/services/capability-packages/capability-prompt-context.service.ts`
- `server/services/capability-packages/capability-module-runtime.service.ts` (+17)
- `server/routes/generate.routes.ts` (+23)
- `server/services/game/gm-prompts.ts` (+14/−3)

The `prompt-context` permission already existed in the manifest enum but had **no consumer**, so
declaring it did nothing. This gives it one: a package that owns live state can append a block to the
system prompt of each turn, so the model narrates in sync with what the player is looking at instead of
guessing.

Registration is gated on the permission. Contributors are keyed by package id, so re-activation replaces
rather than stacks. A contributor that returns nothing is skipped and one that throws is logged and
skipped — a package can never break a turn.

A contribution may also declare which **built-in systems** it replaces (`provides`). Today that's
`{ inventory?: boolean }`: an experience with its own item system takes `[inventory: …]` and the
`PLAYER INVENTORY:` block out of the GM format reminder. This has to be declarative rather than a line in
the package's own prompt, because the format reminder is injected as the LAST user message — a package
can't out-argue it from the system prompt. Anything not declared stays built-in.

### 3. Resource writes — 2 files (+97)

- `shared/types/capability-runtime.ts` (+55)
- `server/services/capability-packages/capability-resources.service.ts` (+42)

The resource facade exposed reads only. A package's setup flow needs to find-or-create the player persona
and its lorebook, so this adds six write methods as thin passthroughs to engine storage.

They are **optional** members of `CapabilityResourceHost`, so adding them is not a breaking type change
for any other implementation, and a host that only wants to expose reads doesn't have to stub six
throwers. The input types are declared narrowly enough to hand straight to storage (a closed
`CapabilityLorebookCategory` union, a required entry `name`), so nothing needs a cast — a wrong value is a
compile error in the package instead of a silent rejection at write time.

### 4. The game-mode experience seam — 7 files

- `shared/schemas/capability-package.schema.ts` (+16) — the `game-surface` slot and its
  `contributions.gameSurface.surfaceClass` option
- `shared/types/game.ts` (+9) and `server/routes/game.routes.ts` (+6) — `gameExperienceId` +
  `experienceConfig` on the setup config, copied to the chat metadata on create
- `client/components/game/GameSurface.tsx` (+323/−36) — the mount and every seam
- `client/components/game/GameSetupWizard.tsx` (+10/−1) — an optional `experiencesSlot`
- **new** `client/components/game/NewGameExperienceChooser.tsx` — the Experiences block
- `client/components/game/GameNarration.tsx` (+27/−6) — five inert theming hooks
- `client/src/localization/locales/en.json` (+4)

The experience is chosen when the game is **created**, from an Experiences block in the setup wizard's
first step, and belongs to that game for its lifetime. It is deliberately not a per-chat toggle: an
experience owns the whole run — its own inventory, combat, calendar and state — so switching one on
midway through an ordinary game, or off midway through its own, leaves a half-built run either way.

Which package it is comes from the **manifest** (it declares the `game-surface` slot), so nothing here is
tied to a specific package id. `grep` for any package name across these files returns zero.

The surface mounts twice: the main layer above the narration for HUD and menus, and a presentation-only
**underlay** in the background band for what must render *behind* the dialogue box, like a standing
sprite. Same `view`, distinguished by constant props, so no new value was needed in the
`CapabilityElement` view union.

The seams are all declarative, and each is gated host-side so a stale value can never leak into an
ordinary game: scene background, speaker avatars, the live `activeSpeaker`, an in-flow choice slot above
the narration box, the host's sprite-scale setting, the resolved scene background handed to the underlay,
and a **chrome declaration** (`providesInventory` / `providesCombat` / `providesPlayerInput`) by which an
experience says which built-in chrome it replaces. The Inventory and Start-combat buttons are hidden by
simply not receiving their callbacks — withholding beats hiding, since there's no dead button and no way
to reach the built-in flow underneath.

The five hooks in `GameNarration` are class names and one `data-` attribute, with no logic. Nothing in the
engine styles them, so Classic renders identically; an experience needs them because that chrome renders
*outside* its own element and is otherwise unreachable from its stylesheet.

The first-game tutorial is suppressed for an experience — it spotlights built-in chrome that isn't on
screen. Suppressed, **not consumed**: the persisted "never show again" flag is left untouched, so a player
who runs an experience first still gets the tour on their first ordinary game.

### 5. Clear pending choices and QTEs on a scope change — `packages/client/.../GameSurface.tsx` (+3)

Pre-existing, and the only change here that touches ordinary games. Choices offered by one game stayed on
screen in the next one and were answerable there; an active QTE leaks the same way and carries a timer
that then fires in the new game. Every site clearing either one is driven by a player action or by the
session concluding, and the effect keyed on the scope resets about twenty other pieces of turn state
without touching these two.

It is fixed here rather than separately because it lives in the same effect this PR was asked to reset the
experience seam state in, and resetting ours while leaving these two leaking would be arbitrary. Say the
word and I will split it out.

-

## Validation

### Manual verification notes

Ran against `staging`, with a package that provides an experience installed:

**The experience path.** The surface mounts, HUD and menus render over the narration, and the standing
sprite draws *behind* the dialogue box (the underlay mount). Speaker avatars resolve for a cast that has
no character cards. The experience's choice menu lands above the narration box rather than floating over
it. The built-in Inventory and Start-combat buttons are absent, and the text input disappears while a menu
is on screen and comes back when there is nothing to pick — the chrome declaration is live, not one-shot.

**Switching between modes.** Moved back and forth between an experience game and an ordinary one
repeatedly. No state leaks in either direction: no stuck background, no leftover avatars, no chrome from
one showing up in the other.

**Ordinary Game mode.** Checked on a fresh ordinary game: the setup wizard is unchanged apart from the
added block, and the Inventory / Start-combat buttons, text input, map, party bar, HUD widgets and
per-scene background and music all behave as before. **The first-game tutorial still fires** on the first
ordinary game even after playing an experience first, which is the point of suppressing it rather than
consuming the flag.

One deliberate change to ordinary games, described under part 5 below: pending choices and QTEs are now
cleared when the game scope changes.

**Empty state.** With no experience installed, the block renders its "none downloaded yet" state with an
affordance to open the catalog, and the built-in wizard is untouched.

**Part 1** — the manifest is rebuilt after activation and indexes the package's art (114 entries for the
installed package) on the first restart after installing, with no second restart needed.

**Part 2** — verified in a real game in both modes: an ordinary game still gets the `[inventory: …]`
command, and the experience — which declares `provides: { inventory: true }` — does not.

**Part 3** — creating a game through the experience's setup creates the player persona and its lorebook
with the expected fields.

-

## Docs and release impact

- [ ] No docs changes needed
- [x] Translated docs on the `docs-i18n` branch updated to match, or a `[docs-i18n]` follow-up issue opened (see CONTRIBUTING.md § Translated documentation)

`docs/development/optional-agent-packages.md` — extended "Package model" to cover full Game-mode
surfaces, per-turn prompt contributions, and the resource writes. No version heading, as suggested.
Nothing under `docs/` proper changed, so no `docs-i18n` work is needed. No version bump.

Before:
image

After:
image

## Summary by CodeRabbit

* **New Features**
  * Added complete, package-provided Game experiences with custom HUDs, menus, combat, setup flows, avatars, backgrounds, and visual styling.
  * Added Game experience selection during new-game setup, including catalog access when none are installed.
  * Added support for saving experience-specific setup configuration.
  * Added package-provided turn context and support for replacing built-in systems.
  * Added persona and lorebook creation, editing, and entry management for supported experiences.
  * Newly installed package assets are available immediately after activation.
* **Bug Fixes**
  * Improved resilience when asset rescanning fails during package activation.


SonaGat3:
Thanks for the thorough pass. Pushed a commit addressing five of the eight comments; notes on the
remaining three below.

### Applied

**Seam state now resets when the game scope changes.** This was a real bug and the manual testing missed
it: I moved between an experience game and an ordinary one repeatedly, where `experienceSurfaceActive`
does go false and the host guards do clear everything — but never between two experience games, where it
stays true throughout. `experienceAvatars`, `experienceChrome` and `experienceBackgroundUrl` now clear on
a key combining `sceneRuntimeScopeKey` and `experienceSurfaceId`, following the same
`prevScopeRef` + effect shape the file already uses for the scene restore state. Good catch on
`providesPlayerInput` in particular: a stale `true` would have left the player with no way to send a turn.

**Dialog semantics on the experience setup panel.** Added `role`, `aria-modal` and `aria-labelledby`
pointing at the panel title, matching what `GameSetupWizard` sets on its own panel, plus an Escape
listener gated on `launching` so it can't tear down a run mid-launch.

**`gameExperienceId` and `experienceConfig` are constrained.** The id now uses the same
`/^[a-z0-9]+(?:-[a-z0-9]+)*$/` + `.max(80)` the manifest schema applies, which is right since this value is
matched against a manifest id to mount the surface. `experienceConfig` keys are capped and the whole
record is bounded by a serialized-size limit.

**Pino gets the error object first.** Fixed — that one was a straightforward miss against the contributing
guidelines, and the stack is exactly what you'd want when a contributor misbehaves.

**Per-contributor deadline.** You were right that the `try`/`catch` isolates a throw but not a hang, and
the docblock claimed a guarantee the code didn't deliver. Each contributor now runs under a deadline and a
timeout is treated exactly like a throw: logged, skipped, turn proceeds.

I kept the calls **sequential** rather than moving to `Promise.all`. The contributors compose a single
prompt, so block order has to be stable across a run, and expressing that as "registration order" is
clearer about the intent than relying on `Promise.all` preserving array order. The latency argument is
also weak in practice — a contributor builds a string from state it already holds, and with the deadline
in place the worst case is now bounded either way. Happy to switch if you'd rather have the concurrency.

Verified by running the collector: a contributor that never settles is dropped at the deadline without
delaying the others, a throwing one is still skipped, registration order and the `provides` union are
unchanged, and no timer is left behind.

### Not applied

**Restricting `pushExperienceBackground` to game-asset paths.** I don't think this is a boundary. The
value comes from an installed capability package, whose client bundle is loaded and executed in the page —
it registers the custom element this seam talks to. A package that wanted to render a remote image can
already do so inside its own element without going near this seam, so validating the string here
constrains nothing that isn't already unconstrained. The seam exists so the package's background renders
*behind* the shared narration, which it cannot reach on its own.

That said, I have no objection to adding a resolve-under-`game-assets` check as defence in depth if you'd
prefer packages to be held to it — say the word and I'll add it.

**The open index signature on `CapabilityLorebookEntryInput`.** The three named fields are the ones a
package needs to reason about; the signature is there because the entries go to
`lorebooks.bulkCreateEntries`, which accepts more than those three, and naming them all would couple this
published type to the engine's internal entry shape — every field added to storage would become a change
here. Your point about misspellings is fair, though, and a nested bag would trade one cost for the other.
This feels like your call on where you want that published surface to sit, so I'm happy to take it either
way.

**`app.log.warn` in `app.ts`.** I believe this one is a false positive. `app.ts` doesn't import the shared
logger at all, and all ten of its logging calls go through `app.log.*` — the line right below mine reads
`app.log.warn({ err: error }, "sidecar bootstrap failed")`, which is the same shape. `CONTRIBUTING.md`
lists `app.log`/`req.log` as acceptable, and the call sits inside `buildApp` where `app` is right there. I
kept it consistent with the file, but I'll switch it if you'd rather the shared logger be used
unconditionally.

### Also in this commit

I cut the comments back substantially, which should address the redundant-comments flag on the walkthrough.
The bracketed tags were scaffolding from development and are gone, along with several explanations that
were repeated at three or four call sites. What's left explains why a seam exists or what breaks without
it, rather than restating the line below it.

`pnpm check` passes. `pnpm regression:prompt` is at 147 passing with the one failure that reproduces
identically on unmodified `staging` (`Storyboard host consumes package-owned prompts without restoring
Engine globals`), unchanged from before this commit.

@coderabbitai full review

SonaGat3:
Second round. Three more commits: the four remaining items from the first pass, everything from the full
review, and one bug the playtest turned up.

### From the full review

**`providesChoices` on the chrome declaration.** You were right that the host was depending on the
package's prompt discipline. I took the declarative option rather than giving the surface precedence, and
it turned out to matter: the package I develop against deliberately *relies* on the Classic choice cards
for in-scene `[choices:]` and steps its own menu aside when they appear. Making the surface win
unconditionally would have broken it. Declared, the anchor stays mounted and the Classic cards stand down;
undeclared, nothing changes at all.

**JSON repair on the experience path.** Confirmed and fixed. The modal only mounted inside the built-in
wizard, so the same malformed opening was repairable in one setup path and a dead end in the other. It now
mounts outside the chooser, and a launch failure is offered to the host before being rethrown.

**The `experienceOwnsGame` fallback** is now scoped to the in-flight window of the installed-packages
query, so a game whose package was uninstalled gets the built-in chrome back instead of losing both. Also
applied: the shared eligibility helper (which drops a local type and its cast), `unref` on the deadline,
the localization key for the panel title, and the docs correction — it is Game turns specifically, and a
contributor that exceeds its deadline is skipped the same way one that throws is.

### Found while playtesting — pre-existing, fixed here

Choices offered by one game were still on screen in the next one and answerable there, against a GM that
never asked them. An active QTE leaks the same way and is worse, since it carries a timer that then fires
in the new game.

This is not caused by this PR. Every site that clears either one is driven by a player action or by the
session concluding, and the effect keyed on the scope resets about twenty other pieces of turn state
without touching these two — switching between two ordinary games has always done this. It surfaces here
because an experience draws its own menu, so the stale cards render beside it and read as the new mode
double-listing its options.

I fixed it in the effect that already owns that reset rather than filing it separately, since resetting
the experience seam state there while leaving these two leaking would be arbitrary. **This does change
Classic behaviour**, so I have said so explicitly in the description rather than leaving the earlier
"unchanged" claim standing. Happy to pull it into its own PR if you'd rather keep this one strictly to the
feature.

### Still open for you

**The capability API version.** Still at 1.7, still deliberately not bumped.

**The index signature on `CapabilityLorebookEntryInput`.** Unchanged. These entries go to
`lorebooks.bulkCreateEntries`, which accepts more than the three named fields, so spelling them all out
would couple this published type to the engine's internal entry shape and make every field added to
storage a change here. Your point about misspellings is fair and a nested bag trades one cost for the
other — it is a call about where you want that published surface to sit, and I will take it either way.

**Running contributors concurrently.** Still sequential. They compose one prompt, so block order has to be
stable, and "registration order" says that more clearly than relying on `Promise.all` preserving array
order. Your new point is fair though — the deadline does make the sequential worst case additive at
N × 2s. With one contributor it is moot, and I would rather change it when a second one exists and the
ordering contract can be settled deliberately.

**The client-side inventory parser.** Left as is, and I checked the consequence rather than assuming it.
`buildSessionSummaryPrompt` takes only a language, so `statsSnapshot` is written by the model from the
transcript, not from engine state; across the server, `playerInventory` reaches the GM prompt in exactly
one place, `buildGmFormatReminder`, which the `provides` declaration already gates. The only other reader
is encounter generation, which an experience declaring `providesCombat` never starts. So a tag the GM is
no longer asked to emit would land in a panel the player cannot see and go nowhere else. Gating the parser
would mean another host-side gate for state with no path to the prompt.

### Validation

`pnpm check` passes. `pnpm regression:prompt` is at 147 passing with the one failure that reproduces
identically on unmodified `staging` (`Storyboard host consumes package-owned prompts without restoring
Engine globals`), unchanged throughout.

Beyond that I exercised the parts that types cannot cover: the contributor deadline against the built
service (a contributor that never settles is dropped at the deadline without delaying the others, a
throwing one is still skipped, registration order and the `provides` union hold, no timer is left behind),
and both gate changes as truth tables — the `experienceOwnsGame` change alters exactly one case out of
five, and `providesChoices` exactly one out of eight, in each instance the broken one. And the surface
itself was played against a real installed package, which is how the leak above turned up.

@coderabbitai full review

SonaGat3:
Both rounds addressed. Five of the seven findings are fixed; notes on the other two below.

**The deadline now covers thenables.** This was the worst of the batch and worth saying plainly: the guard
*looked* like it closed the hole while leaving it open. `withDeadline` only raced a native `Promise`, so
anything else fell through to `Promise.resolve`, which *adopts* a thenable — one that never settles slipped
straight past and hung the turn, which is exactly what the deadline was added to prevent. Anything that can
defer now races it; only a primitive is handed straight back, so the ordinary synchronous return still
costs no timer.

Verified against the built service, since types cannot catch this: a thenable that never settles is dropped
at the deadline (2.0s, previously an indefinite hang), one that resolves is used, one that rejects is
skipped, and a plain-object return resolves in 0 ms — the race adds no latency to the normal path.
Registration order, the `provides` union and the existing string, promise and null paths are unchanged.

**A game-surface package must declare a client entrypoint.** Correct — the module loader skips a package
without one, so the wizard could offer an experience that then renders neither its setup view nor its
surface. Enforced as a cross-field rule on the manifest union, so it fails at install with a message naming
the reason, and the shared helper where the chooser and `GameSurface` both resolve experiences carries the
same check for anything installed before the rule existed. No catalog package declares `game-surface`
today, so nothing currently installable is affected.

**The built-in HUD widgets are now hidden on mobile too.** A real gap on my side: I gated the desktop rail
and left the mobile slot and the compact rails beside the choice stage ungated, so a package drawing its
own HUD could still have the built-in widgets rendered over it on a narrow viewport. All three render paths
now share the same condition.

**The package's setup config is read defensively.** Also fair — it is package-supplied, so a null or
non-object return threw on property access instead of failing validation with a message the player can act
on.

### Not applied

**Relaxing the classic fields in `gameSetupConfigSchema` for experience-driven creation.** I think this one
rests on a misreading. The chooser does not supply only `gameExperienceId` — it spreads the entire config
the package produced and adds that one field:

```
setupConfig: { ...cfg, gameExperienceId: activeId }
```

An experience runs its own setup wizard and returns a complete `GameSetupConfig`; the host then creates a
real game from it, with a real GM prompt, party and world. Those fields are required because the game
genuinely needs them, and relaxing them would let a package create a half-formed game that fails later and
further from the cause. The current shape also keeps the contract honest: whatever an experience replaces
in the UI, the game it creates is still a Game-mode game.

**Scoping the resource writes to package-owned records.** The underlying concern is real, so it is worth
answering with specifics rather than waving it off.

A package's `activate(context)` receives `context.app`, the Fastify instance
(`capability-module-runtime.service.ts`), and `app.db` is the database itself — `app.decorate("db", db)` in
`app.ts`. Every package already has unrestricted read and write access to all storage, before this PR and
independently of it. Scoping these four methods would leave the sanctioned path narrower than the
unsanctioned one sitting beside it, which buys the user nothing.

There is also no ownership model to reuse: nothing associates a persona or a lorebook with the package that
created it, and the reads this facade already exposed are not scoped either.

So packages are broadly privileged by design today, and that is an engine-wide property rather than
something these six methods introduce. Changing it looks like real work to me — ownership on the records,
scoping reads as well as writes, and deciding what happens to a package's data when it is uninstalled — and
I would rather not bolt a partial version onto this PR. Happy to open a separate issue with what I found,
or to implement it here if you would prefer it land together.

`pnpm check` passes; `pnpm regression:prompt` is at 147 with the one failure that reproduces on unmodified
`staging`.

@coderabbitai the last two items are deliberate and explained above — please take them into account rather
than re-raising them. No need for another full review; the push already covers the changed files.

SonaGat3:
@coderabbitai the first one is fixed; the second I would rather leave.

**`selectedId` is pinned while a launch runs.** Fair — resolving it from the installed list meant a refetch
during game creation could drop it and unmount the setup element in the middle of the launch it was
driving. The resolution still governs everything else, so an uninstalled package falls back to the
built-in wizard, but while a launch is pending the remembered id holds. The controls are frozen at that
point anyway.

**The duplicated `inputSlot` block is pre-existing.** `staging` already has two `<GameInput>` blocks and two
`inputSlot={` sites, at lines 11710 and 11799 of `GameSurface.tsx` — one per branch. This PR only wraps
each in the `providesPlayerInput` guard, two lines apiece. Hoisting them would be a refactor of host code
this PR does not otherwise touch, on a hot path, and `CONTRIBUTING.md` asks for small focused changes. It
is a reasonable cleanup, just not one that belongs to this branch — happy to open it separately if the
maintainer wants it.

SonaGat3:
@coderabbitai the actionable one is fixed. Here is where I land on the four outside-diff items, so they do
not need to come back each round.

**Speaker-avatar map** — fixed, and normalized rather than just re-checked. The seam is iterated during
render, so a `size` check alone let any object carrying that property through and it threw a `TypeError`
mid-render. The stored value is now always a real Map or an empty one, which also keeps a valid
`playerAvatarUrl` when the map beside it is malformed — a tightened condition would have dropped both.

**Player-only avatar with no persona** — leaving as is. `effectivePersonaInfo` enriches an existing
persona, and with no persona there is no speaker identity in the narration to attach an avatar to. Making
one up to hang the image on would invent an entity the rest of the game does not know about.

**Using `experienceOwnsGame` for the Classic suppression gates** — leaving as is, and the distinction is
deliberate. `experienceSurfaceActive` means the surface is mounted and drawing. `experienceOwnsGame` is
broader: it is already true from the chat metadata before the installed-packages query resolves. Suppressing
built-in chrome on the broader flag would hide the built-in UI while nothing has replaced it yet — a blank
game area on a cold cache. So it gates the things that must never appear for such a game at all (the
first-game tutorial), and the surface flag gates the things that only make sense once something else is
drawing.

**Lazy background resolution** — leaving as is. A package pushes a ready asset path because it knows its own
bundled art; the `scopedAssetMap` lookup is a compatibility fallback, not the primary route. Deferring
resolution to display time would make the seam depend on a manifest the package deliberately does not need.
Happy to revisit if a concrete case turns up where a tag only becomes resolvable later.

**Inventory and combat declarations on every Classic path** — this one I think is a real point rather than a
nit, and that is exactly why I would rather not patch it here. The declaration currently withholds the entry
points, so an experience that ships its own systems hides the buttons; a turn that reaches built-in combat
another way, such as the GM emitting `[state: combat]`, would still enter it. Where that boundary should sit
— whether a declaration suppresses a system everywhere or only its affordances — is a design decision about
the capability contract, and it touches the combat flow well beyond this seam. The maintainer has said he
wants to make the last calls on this PR himself, so I would rather he made that one with the whole thing in
front of him. I will implement whichever way he decides.

SonaGat3:
@SpicyMarinara I thinkg that I defeated the rabbit, it was... a learning experience.

SpicyMarinara (approved):
Maintainer review complete on 95b659232. CodeRabbit has no unresolved threads; focused client/server/shared TypeScript, client ESLint, localization, version consistency, and capability lifecycle validation pass. UI uses theme accent tokens and responsive touch/focus states.

SpicyMarinara (approved):
Reconfirming maintainer approval on 6fd713b31 after the clean staging sync; the reviewed capability and UI changes are unchanged.

PR #4550, opened 2026-08-04
feat(noodle): new-since-last-visit divider and counter (Slice 8f-5, plus Noodle), and blurred locked teasers

Gunterlie:
## Why

**Slice 8f-5 — the watching surface.** NoodleR gave a returning user no reason to come back and no way to see where they stopped. The plan calls this "the cheapest real win in the document" and split it out specifically so a scope cut could not eat it:

> The "new since your last visit" divider plus the entry-point counter costs one stored timestamp per viewer persona and no provider budget, and it is what gives a user a reason to come back at all.

Also carries a side fix found while working in the same files: locked posts had stopped showing a blurred image at all.

## 8f-5: divider and counter

**One timestamp, per viewer persona.** `noodlerFeedSeenAt` joins `NoodleAccountSocialSettings` next to the existing `notificationsReadAt`, and rides the account-settings patch route that already exists — no new endpoint. Per *persona* rather than per user is the plan's explicit requirement, and it matters: NoodleR follows and locked-post access are already persona-scoped, so an account-wide timestamp would let visiting as one persona silently clear another persona's counter despite them having entirely different Following feeds.

**It advances only once the feed is actually shown**, not on app entry and not while discovery search has replaced the feed — again per the plan. The subtlety that costs a line of thought: advancing the stored timestamp the moment the feed renders would erase the divider out from under the reader, so the value the divider reads is frozen per persona at that moment while the stored one moves on.

**The divider** sits at the boundary after the run of new posts (the feed is newest-first, so everything new is one contiguous run at the top). It is shown only when there is something on both sides of it — with no older posts it would sit at the bottom labelling nothing, and with no new ones it says nothing.

**The counter** is a badge on the NoodleR mode toggle, so it is visible from the *Noodle* side — which is where the user is when they are not already watching NoodleR. It excludes the viewer's own stage profile's posts, which are never "new" to the person who just wrote them, and stays silent for a persona that has never been shown the feed rather than announcing the entire backlog on first run.

`countNoodlerPostsSince()` lives in `packages/shared/src/utils/noodle-unseen.ts` — pure, shared, and directly testable rather than buried in a hook.

## Carried to public Noodle

Not in the plan — 8f-5 scopes to NoodleR — so flagging it as an explicit addition rather than something folded in silently. The same blind spot exists on Noodle: characters post while you are elsewhere, and nothing tells you how many landed. The existing bell counts things *addressed to you* (mentions, replies, follows), not new posts in your timeline.

`noodleFeedSeenAt` is a **separate field** from `noodlerFeedSeenAt`, not one shared value — otherwise opening either surface clears the other's counter, which is the same cross-contamination the per-persona scoping exists to prevent.

The one thing that needed real care: Noodle's timeline sorts by **latest activity**, not creation time. Counting by `createdAt` would put the divider somewhere other than the boundary the reader sees. So both the count and the divider measure the same value the sort does — a post bumped by a new reply is unseen again. That expression was inline and evaluated twice per comparison; it is now one `timelineActivityAt()` shared by the sort and both consumers.

The badge and the divider are shared components in `NoodleShell` rather than a second copy of each.

## Side fix: blurred teasers for locked posts

A locked post with an image rendered an empty grey frame. The onboarding wizard, meanwhile, teaches the locked card using a **pre-blurred** demo image (`NoodlerBulkCreatePanel.tsx:77`), described there as "what the user is being taught to recognise" — so the tutorial promised something the feed did not deliver.

Slice 8f-2 caused this deliberately, and its reasoning was right as far as it went: *"a browser-side blur would still disclose the original image bytes."* True — the bytes are in the page. But the remedy was to withhold the image entirely, which removed the teaser as a side effect.

Fixed by blurring **server-side** instead: `readNoodlerLockedTeaser()` downscales to 24px *before* blurring, which makes the derivative destructive rather than merely obscured, and caches it beside the original. `resolveReadableNoodlerPost` now reports `locked` rather than refusing, so the access-checked media route can answer a locked viewer with the teaser; `resolveGatedNoodlerPost` checks the flag, leaving interaction gating unchanged. The viewer projection keeps `imageUrl` for locked posts **only** when it points inside the NoodleR media namespace — media stored elsewhere has no derivative and is still withheld. The card's CSS blur drops `blur-xl` → `blur-sm`, since it is presentation over an already-blurred image now, not the protection.

Fail-closed throughout: no `sharp` (Android/Termux has no prebuild), undecodable source, or missing file all serve nothing and leave the empty frame — today's behaviour, never the original.

Also extracted the lazy `sharp` loader out of `game-assets.routes.ts` into `packages/server/src/utils/sharp.ts` rather than copying it. No behaviour change for game assets.

## Plan updated

`noodler-pr-split-living-plan.md` still described 8f-1/2/3/4/8g as unmerged on `big-chungus-1`; #4515 shipped them on 2026-08-03. The status table and implementation-status section now reflect that, record 8f-5 as done here, and note the locked-media defect so it is not rediscovered. **8f-6** (creator-page operator area, source-changed/source-missing notices) is now the only remaining 8f unit.

## Validation already run

`pnpm check` and `pnpm regression:noodle` pass on this head. Two new regressions, both wired into `regression:noodle`:

- `noodle-feed-seen.regression.ts` — the timestamps survive a storage round-trip, one persona's visit does not clear another's, a Noodle visit does not disturb the NoodleR timestamp, and the counting rules hold (own posts excluded, no-timestamp means zero, unparseable timestamp means zero, a post exactly at the boundary counts as seen).
- `noodle-locked-teaser.regression.ts` — the teaser is not the original bytes, is ≤24px wide and under a quarter the size, is reused from cache without rebuilding, and returns null on an unreadable source. Where `sharp` cannot load it pins the fail-closed branch instead.

The boxes below are for the human contributor and are intentionally left unchecked.

## Test plan

- [x] `pnpm check`
- [x] `pnpm regression:noodle`
- [x] Manually verify the divider appears at the right boundary after publishing new posts and returning to the feed, and does **not** vanish while you are looking at it
- [x] Manually verify the divider copy and its placement read correctly — it sits *below* the new run, marking where you stopped; confirm that is the intended reading rather than a label for the posts beneath it
- [x] Manually verify the counter appears on the NoodleR toggle from the Noodle side and clears after the feed is shown
- [x] Manually verify persona scoping: visit as persona A, switch to persona B, and confirm B's counter and divider are untouched
- [x] Manually verify a persona opening NoodleR for the very first time sees no counter and no divider
- [x] Manually verify entering NoodleR but landing on a creator profile (not the feed) does not clear the counter
- [x] Manually verify the same divider and counter on the **public Noodle** timeline
- [x] Manually verify the two counters are independent: visiting Noodle must not clear the NoodleR count, or vice versa
- [x] Manually verify a post bumped by a new reply (rather than newly created) is counted and sorted as unseen on Noodle
- [x] Manually verify opening Noodle on the notifications or profile view does not clear the timeline counter
- [x] Manually verify a locked NoodleR post with a generated image shows a blurred teaser and reads as "paywalled" rather than broken
- [x] Manually compare the real blurred teaser against the wizard's pre-blurred Mari teaser — different means of production, so blur radius or teaser width may need tuning to match
- [x] Manually verify unlocking or subscribing swaps the teaser for the sharp original
- [ ] Manually confirm devtools/network on a locked post exposes only the ~24px derivative, never the original
- [ ] Manually verify deleting a post removes both the media file and its `.teaser.jpg`
- [x] Manually verify light and dark themes on the divider, the badge, and the locked card
- [ ] Verify game-asset image metadata still works after the `sharp` loader extraction

## New Features
- Added unseen-post counts to Noodle and NoodleR mode controls, capped at 99+.
- Added “since last visit” dividers for newly arrived feed posts.
- Feed visits are tracked independently for Noodle and NoodleR.
- Locked NoodleR media now displays blurred teasers while protected originals remain inaccessible.

## Bug Fixes
- Improved feed ordering to account for reply activity.
- Standardized blurred presentation for unrevealed locked images.

## Tests
- Added regression coverage for feed-seen tracking and locked-media teasers.


Gunterlie:
@coderabbitai review

Gunterlie:
@coderabbitai All eight points were valid and are fixed in 306f4c5f — no skips.

**Divider placement** (`NoodleHome.tsx`, `NoodlerHome.tsx`) — was `findIndex(not-new)`, so an own post interleaved in the unseen run ended it early and the divider landed above posts the viewer had not seen. Now `findLastIndex(isNew)` + 1, still requiring posts on both sides. Own-post exclusion is unchanged in both the badge count and the "is new" predicate.

**Search suppression** — `timelineIsOnScreen` now also requires `!normalizedPostSearch`; NoodleR's `feedIsOnScreen` requires `!search.trim()`; the NoodleR divider is suppressed when `searchTerm` is set (Noodle's already was). A filtered list is not the feed, so it neither records a visit nor carries a boundary marker.

**Locked teaser fallback** (`NoodlerPostCard.tsx`) — added `failedMediaSrc` state set from the ` onError`, suppressing the image for that source. `post.hasImage` still renders the empty frame, including on a 404 teaser.

**Badge a11y** (`NoodleShell.tsx`) — the visual count is now `aria-hidden`, so the localized sr-only label is announced once instead of "3 3 new since your last visit".

**`noodle-feed-seen.regression.ts`** — added an unfollowed external creator with a newer post; expectation is now 2, so a `followed` filter would fail the test.

**`noodle-locked-teaser.regression.ts`** — three additions:
- sharp-unavailable branch now passes an existing, valid PNG, so the `null` proves fail-closed on Sharp rather than on a missing file;
- asserts `meta.format === "jpeg"`, plus a pixel comparison against the same resize *without* `blur()` — max per-channel delta must exceed 32. Verified it fails when the blur is removed (delta drops to 0);
- a non-image file under `workDir` now asserts `null`, covering the decoder-failure path.

**Living plan** — the 8f-5 entries (order table and implementation-status section) now document both timestamps (`noodlerFeedSeenAt`, `noodleFeedSeenAt`) and why they are separate, both counters, and the reply-bumped activity-time counting on public Noodle, with the existing regression and implementation references preserved.

`pnpm check` is clean; both regressions pass.

Gunterlie:
@coderabbitai Second round, pushed in 62b89abf — three fixed, two skipped with reasons.

**Fixed — monotonic feed-seen guard.** Done in the shared implementation as suggested (`noodle.storage.ts` `patchAccountSettings`, `social` branch): both `noodleFeedSeenAt` and `noodlerFeedSeenAt` are clamped to the newer of stored and patched before the subtree merge, so an out-of-order write cannot resurrect a cleared counter. An unparseable incoming value keeps the stored one. Neither call site changed. Covered by a new case in `noodle-feed-seen.regression.ts`.

**Fixed — `logger.debug` → `logger.warn`** in the sharp load-failure catch. The `sharpLoadFailed` guard still limits it to once per process; error object and message unchanged.

**Fixed — `SharpFn`.** Now `typeof import("sharp").default` instead of `any` (sharp is a real dependency of `@marinara-engine/server`, so the types resolve even though the runtime import stays guarded). That immediately surfaced a genuine mismatch at `game-assets.routes.ts:644` — `Buffer.from()` infers `Buffer<ArrayBuffer>`, sharp's `toBuffer()` returns `Buffer<ArrayBufferLike>` — fixed by widening the annotation.

**Skipped — null checks in `sprites.routes.ts`.** Wrong premise: those callers import `getSharp` from `services/image/sharp-runtime.ts`, a different loader whose return type is non-nullable `Promise<SharpFn>` — it *throws* on load failure rather than returning null. There is no null to check, and the file is untouched by this PR.

**Skipped — `limitInputPixels` on the teaser pipeline.** The suggested value is not what that option means: `limitInputPixels` caps the *source* image, so 24 px (the teaser output width) would reject every real post image. Sharp already defaults to a ~268 MP input cap, and the source here is a file our own gallery pipeline wrote, not request bytes. Compare `normalizeGeneratedBackgroundBuffer()` in `game-assets.routes.ts`, which does set it — with a deliberately large constant — because it processes uploaded bytes. Happy to add an explicit cap if you think the gallery path warrants one, but not at teaser dimensions.

`pnpm check` clean; both regressions pass.

Gunterlie:
@SpicyMarinara can’t merge

PR #4604, opened 2026-08-04
fix(noodle): unwrap LLM array responses in noodler post parsing

MetzeVanDeus:
Many LLMs (especially local models via Ollama/KoboldCPP) wrap the expected object in an array ([{"title":...}]) regardless of the prompt instructing "one JSON object". Add a guard in parseNoodlerPost to unwrap all array cases: empty -> {}, single -> first element, multiple -> first element.

This resolves the 'expected object, received array' ZodError that occurs during Noodler post creation regardless of model, context, or retry attempts.

> [!IMPORTANT]
> Contributions target `staging`. Only `SpicyMarinara` may promote this repository's `staging` branch or a same-repository `hotfix/*` branch to `main`.
> Outside and first-time contributors also require an approving review from `SpicyMarinara`.

## Linked issue

Closes #4597

## Why this change

- Noodler post creation fails with expected object, received array ZodError when LLMs (Ollama, KoboldCPP, LM Studio) return response wrapped in an array

## What changed

- Added array unwrap guard in parseNoodlerPost() — handles [] → {}, [{...}] → {...}, [{...}, {...}] → first element

## Validation

- [x] `pnpm check` passes locally
- [ ] Container (Docker / Podman) built and ran without issue
- [x] Ran the app, clicked through the changes manually
- [ ] Checked edge cases (light + dark mode, mobile viewport, empty states, error paths)
- [x] Above manual verification completed (describe below)
- [x] Read and followed `CONTRIBUTING.md`

### Manual verification notes
- Opened app through launcher, noodle settings, manage schedules on noodler, pressed run now on one of the scheduled profiles. It successfully generated a post.

## Docs and release impact

- [x] No docs changes needed
- [ ] Updated docs (README / CONTRIBUTING / android/README / CHANGELOG) as needed
- [ ] Translated docs on the `docs-i18n` branch updated to match, or a `[docs-i18n]` follow-up issue opened (see CONTRIBUTING.md § Translated documentation)
- [ ] Version/release files updated (only if this PR includes a version bump)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of generated responses containing a single result wrapped in an array.
  * Single-element arrays are now validated using their contained result.
  * Empty arrays, multi-element arrays, and direct object responses continue to be validated without changes.

PR #4634, opened 2026-08-05
fix(noodle): improve stage profile prompt and unwrap LLM arrays

### LINKED ISSUE #4626: [Issue]: Noodler stage profile creation fails with ZodError when LLM returns array-wrapped or invalid enum responses
## Summary

Noodler stage profile creation fails with ZodError when local LLM models return array-wrapped JSON or invalid `disclosureMode` enum values.

## Expected behavior

Stage profile draft is created successfully and stored in the Noodler schedule.

## Actual behavior

Creation fails with:
- `Expected object, received array` — LLM returns `[{...}]` instead of `{...}`
- `Invalid enum value. Expected 'open' | 'hinted' | 'secret', received '...'` — LLM returns a description sentence instead of the enum value

## Steps to reproduce

1. Open the launcher and navigate to Noodle settings → Manage Noodler schedules
2. Click "Run now" on a scheduled profile that triggers stage draft creation
3. If the LLM returns an array-wrapped response or invalid `disclosureMode`, the creation fails with a ZodError

## Environment

- Marinara Engine version: 2.4.2 staging
- Install type: source (Windows)
- OS + version: Windows 11
- Browser or app shell: App shell (launcher)

LLM output:
```
[
{
  "displayName": "Taro",
  "handle": "@Taro_One",
  "bio": "...",
  "stagePersonality": "...",
  "disclosureMode": "hinted"
}
]
```
LLM output (invalid disclosureMode):
```
  "disclosureMode": "Always"
```

Server log:
```
ZodError: Expected object, received array
```

Or:
```
ZodError: Invalid enum value. Expected 'open' | 'hinted' | 'secret', received 'Always'
```

## Additional context

- Same pattern as #4597 (noodler post creation) — PR #4604 fixed that one
- This fix adds a prompt update to specify "single JSON object only" and valid enum values.
- [Draft fix](https://github.com/Pasta-Devs/Marinara-Engine/compare/staging...MetzeVanDeus:Marinara-Engine:fix/noodler-stage-profile-prompt)

## Template check

Gunterlie:
Update the system prompt to tell LLMs to return a single JSON object (no arrays) and specify valid disclosureMode enum values. Also add a minimal array unwrap guard for when local models still wrap the object in brackets.

Skipped ZodError catch — the prompt fixes are more reliable.

> [!IMPORTANT]

## Linked issue

Closes #

## Why this change

-

## What changed

-

## Validation

### Manual verification notes

-

## Docs and release impact

- [ ] No docs changes needed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stage-profile generation by requiring a single, consistently structured JSON response with documented expected keys.
  * Added support for LLM responses returned as either a JSON object or a single-item array.
  * Tightened `disclosureMode` handling to only allow supported exact values (while making it non-mandatory during validation).
  * Normalized user handles by removing a leading `@` when present.

### DIFF
--- a/packages/server/src/services/noodle/noodle-stage-profile-draft.service.ts
+++ b/packages/server/src/services/noodle/noodle-stage-profile-draft.service.ts
@@ -118,7 +118,9 @@ export function buildNoodlerStageProfileDraftMessages(input: {
       role: "system",
       content: [
         "Create one editable NoodleR stage profile draft.",
-        "Return JSON only with displayName, handle, bio, stagePersonality, and disclosureMode.",
+        "Return a single JSON object only — no arrays, no wrapping brackets.",
+        "Keys: displayName, handle, bio, stagePersonality, disclosureMode.",
+        "disclosureMode must be one of: \"open\", \"hinted\", or \"secret\" — use only these exact values.",
         "Make the stage identity distinct, concise, and usable for future NoodleR post generation.",
         disclosureRules(input.request.disclosureMode, identity),
       ].join("\n"),
@@ -212,7 +214,10 @@ export async function generateNoodlerStageProfileDraft(
     debugMode,
     responseFormat: noodleResponseFormat(input.connection.model, "noodler_profile"),
   });
-  const parsed = noodleStageProfileDraftResponseSchema.parse(parseGameJsonish(response.content ?? ""));
+  const unwrapped = parseGameJsonish(response.content ?? "");
+  const parsed = noodleStageProfileDraftResponseSchema.parse(
+    Array.isArray(unwrapped) && unwrapped.length === 1 ? unwrapped[0] : unwrapped,
+  );
   const draft = { ...parsed, disclosureMode: input.request.disclosureMode };
   if (stageProfileContainsPublicIdentity(draft, identity)) {
     throw new Error("Generated stage draft included the linked public identity. Try again with different guidance.");
--- a/packages/server/src/services/noodle/noodle-stage-profile-draft.service.ts
+++ b/packages/server/src/services/noodle/noodle-stage-profile-draft.service.ts
@@ -215,9 +215,11 @@ export async function generateNoodlerStageProfileDraft(
     responseFormat: noodleResponseFormat(input.connection.model, "noodler_profile"),
   });
   const unwrapped = parseGameJsonish(response.content ?? "");
-  const parsed = noodleStageProfileDraftResponseSchema.parse(
-    Array.isArray(unwrapped) && unwrapped.length === 1 ? unwrapped[0] : unwrapped,
+  const unwrappedObj = Array.isArray(unwrapped) && unwrapped.length === 1 ? unwrapped[0] : unwrapped;
+  const raw = noodleStageProfileDraftResponseSchema.omit({ disclosureMode: true }).passthrough().parse(
+    unwrappedObj as Record<string, unknown>,
   );
+  const parsed = { ...raw, handle: (raw.handle as string)?.replace(/^@/, "") ?? "" };
   const draft = { ...parsed, disclosureMode: input.request.disclosureMode };
   if (stageProfileContainsPublicIdentity(draft, identity)) {
     throw new Error("Generated stage draft included the linked public identity. Try again with different guidance.");


Gunterlie:
## Review

The array unwrap mirrors the accepted fix in #4604, so it is consistent with precedent. It only closes half of #4626.

### Findings

**1. blocking — the `disclosureMode` half of #4626 is not fixed** (`noodle-stage-profile-draft.service.ts:217`)

`noodleStageProfileDraftResponseSchema` is `noodleStageProfileSchema`, whose `disclosureMode` is a strict enum (`packages/shared/src/schemas/noodle.schema.ts:243,246`). Three lines later the value is thrown away: `{ ...parsed, disclosureMode: input.request.disclosureMode }`. So a model returning `"disclosureMode": "Always"` fails the entire generation over a field the code never uses. That is the second error in the issue report, and prompt wording is not a guarantee — it is the same "local model ignores the instruction" class that motivated the unwrap guard directly above it.

Repair: parse with the field dropped, e.g. `noodleStageProfileDraftResponseSchema.omit({ disclosureMode: true }).parse(...)`. One line, removes the failure mode instead of asking the model nicely.

**2. medium — `.strict()` still rejects any extra key** (same line)

`noodleStageProfileSchema` is `.strict()`, so a model emitting `name`, `avatar`, or a stray `reasoning` field ZodErrors the run. The premise of this PR is that models do not obey the prompt. Consider stripping extras on the LLM response path only — do **not** loosen `noodleStageProfileUpdateSchema`, which is a trust boundary for client input and should stay strict.

**3. low — no guard for an `@`-prefixed handle**

The issue's own sample output is `"handle": "@Taro_One"`. The schema accepts it and display code formats as `@${handle}` (e.g. `disclosureRules`, line 57), yielding `@@Taro_One`. Not a regression from this PR, but it is in the pasted repro.

**4. low — multi-element arrays still throw the raw ZodError.** `length === 1` only. Matches #4604 so not worth blocking on.

### PR hygiene

- The body is an empty template: `Closes #`, "Why this change", and "What changed" are all blank while the work clearly closes #4626. Per `CONTRIBUTING.md` the *why* must be explicit.
- No test. `packages/server/scripts/run-tests.mjs` is the wired surface and `buildNoodlerStageProfileDraftMessages` is already exported; one case per failure mode from the issue (`[{...}]`, `disclosureMode: "Always"`) would keep this from regressing.
- Base `staging`, no docs/version impact — both correct. Validation boxes correctly left unchecked.

### Not verified

`pnpm check` was not run and the flow was not exercised; findings come from reading the diff, the schema, and #4626.

MetzeVanDeus:
I'm not sure of the exact flow, but here's what I changed:
1. Blocking: Used .omit({ disclosureMode: true }) before Zod parse so invalid enum values don't block the whole request (the value is always overridden from the request anyway)
2. Medium: Added .passthrough() to allow extra keys from the LLM without crashing
3. Low: Strip leading @ from handle to avoid @@ in display
4. Prompt: Added "single JSON object only" and valid enum values for disclosureMode

pnpm check passed locally. I didn't add tests yet but can if that's preferred, based on what I can tell while not that reliable it doesn't consistently fail on my local setup now. Sorry if I'm making mistakes with the process idk how any of this works ngl

Gunterlie:
## Review — head `7faae61`

Findings 1–3 from the previous pass are all addressed, and `.omit({ disclosureMode: true })` is the right call — the value is overridden three lines down anyway, so the strict enum no longer kills the run. That closes the second half of #4626. Nice work turning this around quickly, and thanks for the PR.

Two nits below, neither blocking — this is mergeable as it stands.

**a. `.passthrough()` could be `.strip()`** (`noodle-stage-profile-draft.service.ts:219`)

`passthrough()` keeps the junk keys and carries them into `draft`, which goes straight back to the client (`noodle.routes.ts:839`). The client posts that back to `/accounts/:id/noodler`, validated with `noodlerAccountCreateSchema` — `.strict()`. So a stray `reasoning` key survives generation and then 400s at create instead. The bulk path is fine (`noodle.storage.ts:1355` picks fields by name), but the interactive path is the common one. `.strip()` is the same edit length and drops extras rather than forwarding them.

**b. handle is stripped after validation** (line 222)

`min(1)` runs on the raw value and the `@` comes off afterwards, so `"handle": "@"` yields `""` past the schema and into `normalizeHandle("")`. Stripping on `unwrappedObj` before `.parse()` lets the schema cover it — that also makes the `as string` cast and `?? ""` unnecessary, since `raw.handle` is already `string` after `.omit()`.

**On tests:** worth adding if you're up for it, but don't let it hold the PR. The wired lane is `packages/server/scripts/run-tests.mjs`, and `buildNoodlerStageProfileDraftMessages` is already exported. Two cases — `[{…}]` input and `disclosureMode: "Always"` — pin exactly the two failures from #4626 so a future prompt edit can't quietly bring them back. Your point about it not reproducing consistently locally is actually the argument *for* them: a flaky failure is the kind that comes back unnoticed.

*Not verified: didn't run `pnpm check` or exercise the flow; read from the diff, `noodle.schema.ts:238-247,318`, and both call sites.*

SpicyMarinara:
Can't update this PR, since it's from a fork. Closing. Please reopen properly.

PR #4686, opened 2026-08-06
feat(agents): per-chat image connection override + snapshot button for custom image agents

### LINKED ISSUE #4682: [Feature]: Per-chat image connection override + snapshot button for custom image agents (approved QoL)

From a Discord request (maintainer-approved, see Additional context): custom agents that generate images with their own connection can't be tuned per-chat the way the vanilla Illustrator can —

> "would it be possible to be able to change their prompt and image connections, plus image style, in the chat settings? Just like how a user is able to do so with the vanilla Illustrator Agent. Another nice feature would be to add a snapshot button to generate the Custom Agent's image, without having to use keywords."

Today's workaround is duplicating the agent once per style/connection combo, which is exactly the copy-paste churn the requester wants to avoid.

**Investigation result — two of the three requested overrides already work per-chat for custom agents:**

- **Prompt**: `chatMeta.agentPromptTemplateIds` is keyed generically by agent type; every active custom agent already gets an `AgentPromptTemplateSelect` dropdown in the chat-settings drawer (`ChatSettingsDrawer.tsx:3953-3958`), applied server-side for all agents (`agent-resolution.ts:399-404`).
- **Image style**: the chat-settings "Image style" select writes the chat-wide `chatMeta.imageStyleProfileId`, which the `image_prompt` handler reads **unconditionally** (`generate.routes.ts:8785-8790`; retry mirror `retry-agents-route.ts:3296-3299`) — no illustrator gate — so it already restyles custom-agent images.

The genuinely missing pieces are: **(1) a per-chat image connection override for custom image agents**, and **(2) the snapshot button**. The image connection is locked to vanilla by `usesChatIllustratorSettings = resultAgent?.type === "illustrator" || (!resultAgent && result.agentType === "illustrator")` (`generate.routes.ts:8590-8591`; retry twin `:3238-3239`) — the else-branch reads only the agent's own `settings.imageConnectionId`.

## Proposed solution

Both parts extend existing patterns; no new endpoints, no migrations, no new execution paths (~150-250 LOC across ~9 files).

**1. Per-chat image connection for custom image agents**

- `ChatMetadata` (shared `chat.ts`, next to `knowledgeAgentSources`): `customAgentImageSettings?: Partial<Record<string /* agent type */, { imageConnectionId?: string | null }>>`. The inner key name deliberately ends in `ConnectionId`: the dangling-connection sweeper cleans **any** metadata key matching `*ConnectionId` at any depth, and its module comment already anticipates exactly this `<type>.imageConnectionId` map shape (`connection-reference-cleanup.ts:7-21`) — so deleted connections need **zero** new cleanup code.
- Server: one ~25-line helper mirroring `applyKnowledgeAgentChatSettings` (`knowledge-agent-settings.ts:34-53`) that overwrites the resolved `settings.imageConnectionId` for custom agents when the chat map has an entry; called at the two existing custom-agent settings-merge sites (`agent-resolution.ts:380`, `retry-agents-route.ts:1498`). Both `image_prompt` handlers read the *resolved* settings, so they don't change at all — the override inherits the existing dangling-connection warning and default-Images fallback for free.
- Housekeeping per convention for keyed agent maps: register the key in `sanitizePresetAgentMap` (`chat-presets.storage.ts:64`, preserve-block `:305-310`) and the capability-package cleanup list (`capability-packages.routes.ts:42`).
- UI: in the drawer's existing per-custom-agent card (`renderActiveCustomAgentSettingsCard`), for agents with the `trigger_image_generation` capability, render the same image-connection select the Illustrator section uses ("Agent default" empty option + reset affordance).

**2. Snapshot button**

The plumbing nearly exists: the vanilla Illustrate button calls `retryAgents(chatId, ["illustrator"], { illustratorRetryTargets: ["illustration"] })` → forced manual plan; custom agents already have a per-chat re-run button (`handleRerunCustomAgent` → `retryAgents(chat.id, [agentId])`), but on that path image generation remains the agent LLM's decision (`retry-agents-route.ts:3243`) and `illustratorRetryTargets` is rejected for non-illustrator agents (`:3846`). Add a camera button next to the existing re-run button that passes a new retry option forcing `shouldGenerate: true` on that agent's `image_prompt` result.

*Honest caveat:* a forced run still requires the agent's own prompt template to emit an image prompt — an agent whose LLM returns no prompt generates nothing (`:3249`). The vanilla button avoids this by bypassing the LLM entirely with a hardcoded manual plan; generalizing that to arbitrary user-authored agents is the scope trap this proposal deliberately avoids. Document the requirement instead ("your agent should always emit a prompt when asked").

**Out of scope (keeping it small):** per-chat override of the custom agent's *text* (prompt-writing) connection — the same keyed-map pattern extends to it later if wanted; a chat-settings UI hint that prompt/style already work could ship as just a docs note.

## Alternatives considered

- **Duplicate agents per style/connection combo** (current workaround): works, but multiplies maintenance and clutters the agent list — the request exists because this is painful.
- **Widen `usesChatIllustratorSettings` to all image agents**: rejected — it would apply the *Illustrator's* chat-level keys (`illustratorImageConnectionId` etc.) to every custom agent, conflating unrelated agents' settings instead of keying per agent.
- **Generalize the manual Illustrator prompt plan to custom agents** (LLM-free snapshot): rejected as the ballooning path — vanilla's plan is hardcoded; synthesizing one for arbitrary user-authored agent prompts is a redesign, not QoL.

## Additional context

Maintainer approved with "keep it small-scale" and delegated implementation ("Yes... Just keep it small-scale"). I intend to follow up with the PR per that approval; this issue records the request and the verified implementation shape for review. All file:line references verified against current staging (2.4.2). Related recent work: #4337 (custom image agents + ComfyUI references).

Known edge cases accounted for: connection deletion (auto-swept, see above); agent deletion leaves an inert map entry (harmless — keyed lookups just miss; same behavior as `agentPromptTemplateIds` today); group chats unaffected (plain per-chat metadata); retry path covered symmetrically (both handler twins get the same resolved settings).

## Template check

kolacheee:
## Linked issue

Closes #4682 

## Why this change

Custom agents that generate images can't be tuned per-chat the way the vanilla Illustrator can — the only workaround is duplicating the agent once per style/connection combo. #4682's investigation found per-chat **prompt template** and **image style** already apply to custom agents; this PR closes the two real gaps: a per-chat **image connection** override and an on-demand **snapshot button** (generate now, no activation keywords).

## What changed

**Per-chat image connection override**

- `ChatMetadata.customAgentImageSettings?: Partial<Record<agentType, { imageConnectionId }>>` (shared). The inner key deliberately ends in `ConnectionId`: the existing dangling-connection sweeper prunes any such key at any depth (its module comment anticipates exactly this map shape), so deleted connections need zero new cleanup code.
- New `applyCustomAgentImageChatSettings` helper (mirrors `applyKnowledgeAgentChatSettings`; no-ops for built-in agent types), chained at both custom-agent settings-merge sites (`agent-resolution.ts`, `retry-agents-route.ts`) — both `image_prompt` handlers read resolved settings, so they're untouched and the override inherits the existing missing-connection warning + default-Images fallback.
- Key registered with the chat-preset sanitizer + preserve block and the capability-package cleanup list; removing an agent from a chat prunes its entry (same pattern as the prompt-template override).
- Drawer: an **Image Connection** select on each image-capable custom agent's card (Agent default + missing-connection affordances, matching the Illustrator's select).

**Snapshot button**

- Camera action on the same card → `retryAgents(chatId, [agentId], { forceImageGeneration: true })`.
- Server: the flag (a) injects a `<manual_image_request>` directive into the agent prompt — mirroring the Illustrator's manual-request block — so the agent knows the user explicitly asked and shouldn't decline; (b) forces `shouldGenerate` on that agent's `image_prompt` result, scoped via `!usesChatIllustratorSettings` so the vanilla Illustrator path is untouched; (c) bypasses review-before-send for forced snapshots (the camera press is the explicit request, and the review approval round-trip only supports the vanilla Illustrator); (d) emits an `agent_error` if the forced agent still returns no prompt, so the button never silently no-ops.

**Docs**: new "Per-chat controls for image agents" section in `docs/agents/custom-agents.md`. A `[docs-i18n]` follow-up issue will be opened for the language packs.

**Pre-PR adversarial review**: a multi-lens review of this diff found 3 real issues before submission — review-mode breaking/rerouting forced snapshots, the force being post-hoc only (silent no-op on declined prompts), and stale overrides after agent removal. All three are fixed in this PR and re-verified; the review also confirmed no third settings-resolution path bypasses the override.

## Validation

### Manual verification notes

Automated/scripted checks already performed (green):

- `pnpm check` (lint + production build) — passes, run twice (after implementation and after review fixes)
- Live on the **built** app (production `node dist`, disposable test agent + dummy connection, cleaned up after):
  - Image Connection select renders **only** for image-capable custom agents; the prompt-template picker and re-run button are unchanged for others
  - Selecting a connection persists exactly `customAgentImageSettings: { "<agentType>": { imageConnectionId } }` in chat metadata; **Agent default** reset deletes the entry
  - Camera click sends `POST /generate/retry-agents` with `agentTypes: ["<that agent>"]` and `forceImageGeneration: true`

Still to verify manually in a real browser (please tick above once done):

## Docs and release impact

- [ ] No docs changes needed

`docs/agents/custom-agents.md` updated in this PR; `[docs-i18n]` follow-up to be filed for translated packs. No version bump.

## Summary by CodeRabbit

* **New Features**
  * Added per-chat image settings for custom agents, including image connection selection.
  * Added a camera action to trigger image generation directly from chat settings.
  * Added support for forcing image generation during eligible custom-agent retries.
  * Added clear error handling when no usable image prompt is produced.
  * Preserved custom-agent image settings when applying chat presets.

* **Documentation**
  * Documented custom-agent image controls and camera-triggered generation.


kolacheee:
@coderabbitai review

Issue #4707, opened 2026-08-07
[Issue]: .env watcher stat-polls every 2s forever — highest-frequency permanent idle timer on the server


kolacheee:
## Summary

The `.env` hot-reload watcher stat-polls the env file **every 2 seconds for the server's lifetime** (`fs.watchFile` is stat-polling, not inotify). It is the highest-frequency permanent timer on an idle server — individually cheap, but 30 wakeups/minute forever prevents deep idle on a phone-hosted (Termux) install.

Follow-up from the mobile-battery investigation (passive-server round).

## Expected behavior

Env changes are still picked up promptly (the no-restart security-config path stays intact), with near-zero idle wakeups.

## Actual behavior

`packages/server/src/config/env-watcher.ts:159` — `watchFile(envPath, { interval: 2_000, persistent: false }, handler)`; started unconditionally at boot (`packages/server/src/index.ts:41`). One stat + timer wake every 2s, forever.

## Steps to reproduce

1. Start the server; observe the 2s stat cadence on `.env` for the process lifetime.

## Environment

- Marinara Engine version: staging d6275fa0b
- Install type: any; battery-relevant on Termux
- OS + version: any

## Logs, screenshots, or video

Code trace, line references against staging d6275fa0b.

### Regression-safe fix (designed + red-teamed)

1. `fs.watch` the **parent directory** of the env path, filtered to the `.env` basename — directory-watching survives atomic-replace editors (rename to a new inode) and the file not existing yet, both of which break naive `fs.watch` on the file itself. Keep the existing mtime/size dedupe (`env-watcher.ts:138-148`) and add a ~250ms event debounce (inotify fires multiple events per save).
2. On `fs.watch` error/unsupported filesystem (some FUSE/network mounts), fall back to the existing `watchFile` handler at `interval: 30_000` instead of 2_000 — env edits are human-timescale; 30s worst-case latency cuts wakeups 15× where inotify is unavailable. (inotify works for Termux home-directory paths, where the checkout and `.env` live.)
3. Optionally: a documented `MARINARA_ENV_WATCH=0` opt-out, checked once at startup — but do **not** default it off in the Termux launcher; that would silently kill the documented no-restart security-config reload path (`env-watcher.ts:4-7`) for exactly the users who rely on it.

Filed alongside #4704 / #4705 / #4706 from the same follow-up audit.

## Template check

- [ ] I DID NOT read this template and provide the requested details.

Issue #4704, opened 2026-08-07
[Issue]: Background-autonomous poll fetches full /chats every 30s app-wide (+ O(messages) DM-cleanup scans per hit; mis-wired 30s character-list invalidation)


kolacheee:
## Summary

The background-autonomous poller fetches the **full `/chats` list every 30 seconds, app-wide, even when zero chats have autonomous messaging enabled** — and each of those GETs also triggers `cleanupEmptyRoleplayDmChats`, which runs an O(total-messages) `countMessages` scan per roleplay-DM chat on the file-native store. On a Termux self-host the phone pays both sides of this every 30s for the whole session. A sibling in the active-chat autonomous poll additionally invalidates the **full character list** every 30s tick — and the invalidation is mis-wired (its stated purpose reads a different cache).

Follow-up from the mobile-battery investigation (#4628/#4629/#4630 round; the reporting Termux user still sees no improvement — this poll is the heaviest remaining client-side chatter).

## Expected behavior

Idle-tab network activity near zero when no autonomous chats exist; when they do exist, discovery stays within the current ≤30s window (including chats enabled from another device/tab), with a payload proportional to what the poller actually needs (chat ids), not every chat's full metadata/summary blobs.

## Actual behavior

- `useBackgroundAutonomousPolling` (`packages/client/src/hooks/use-background-autonomous.ts:64`, mounted once at `AppShell.tsx:239`) fetches `api.get<RawChat[]>("/chats")` every 30s (`:92`, cadence `:332-339`, visible-tab gated `:80`), then filters client-side for conversation chats with `metadata.autonomousMessages` (`:99-110`). The full list fetch **is** the discovery mechanism — it runs forever regardless of whether any chat qualifies.
- Server side, **every** `GET /chats` first runs `cleanupEmptyRoleplayDmChats` (`chats.routes.ts:627-631`, `:579-594`): `storage.list()` + `storage.countMessages` per roleplay-DM-thread chat — `countMessages` is a full messages-table filter (`chats.storage.ts:844-847`) — then `storage.list()` **again** and serializes every chat row including full metadata blobs. All of this every 30s, per open tab.
[...]

Issue #4631, opened 2026-08-05
[Feature]: "Reduce ambient animations & effects" toggle (consolidated switch for decorative animation/effect work)


kolacheee:
## What problem does this solve?

Two mobile users reported battery drain / phone heat while using Marinara. An investigation found the biggest causes are fixable bugs (filed separately: #4628, #4629, #4630) — but beyond those, the client has accumulated a large set of **always-on ambient visual effects** that each cost a little continuous rendering work while the screen is on, and collectively add up on phones:

- **41 `infinite` CSS animations** in `globals.css`, several on paint-expensive properties (gradient cyclers: `rgb-pulse-border` 20s, `gradient-rotate` 4s, `pastel-gradient-cycle` 25s; Game Mode ambience: fog shimmer, map marquee, party-pulse, campfire flicker, combat-pulse)
- **Weather effects**: full chat-sized canvas redrawn at ~30 fps continuously while a weather scene is active
- **RGB / pulse accent mode**: rewrites `:root` CSS variables from a JS timer every 500 ms → app-wide style recalculation twice per second (already well-gated to visible+focused, but still a continuous cost when on)
- **Professor Mari pixel scene**: 5 infinite animations running whenever the Home screen is idle; a looping animated GIF while agents run
- **Home ambience**: 653 KB looping GIF logo, JS starfield spawner, gradient title animation
- **JS-driven UI carousels**: RoleplayHUD tracker/inventory 3 s cycles, party-bar 2.5 s avatar cycle, merged-narrator avatar 2 s crossfade (some of these tick in hidden tabs, since JS timers don't pause like CSS animations do)
- **Streaming typewriter**: a requestAnimationFrame loop at display refresh rate (60–120 fps) for the entire generation — the dominant *screen-on* cost during long streams on slow models
- **Pervasive `backdrop-filter` blur** (168 occurrences across 79 files) — GPU-expensive on mobile

Some of these already have individual switches (`gameTextEffectsEnabled` from #3801, streaming speed 100 = instant reveal, partial `prefers-reduced-motion` coverage), but they're scattered, incomplete, and hard for a battery-conscious user to discover.

## Proposed solution

A single **"Reduce ambient animations & effects"** toggle (Settings → Appearance or Accessibility) that turns off the catalog above in one go:

- stop/hide the infinite CSS animations (Mari pixel scene, gradient cyclers, Game ambience, text effects, glows/halos)
- pause the weather canvas (render one static frame)
[...]

Issue #4628, opened 2026-08-05
[Issue]: Professor Mari streaming has no render throttling — per-token full re-render + full markdown re-parse (O(n²), mobile heat/battery)


kolacheee:
## Summary

The Professor Mari assistant streams with **no render throttling**: every SSE token event triggers an immediate state update that re-renders the entire home chat component, and the growing reply's markdown is fully re-parsed on each of those renders — O(reply-length²) work at token rate for the whole generation. The main chat path was given exactly this fix for #2878 (`useThrottledStreamBuffer`); the Mari path never got it. Reported by two mobile users as phone heat/battery drain that is "worse with the assistant", especially on slow models (long streams).

## Expected behavior

Mari streaming coalesces token updates (once per animation frame, like the main chat's `useThrottledStreamBuffer`, added for #2878) and does not re-parse the full transcript markdown on every token.

## Actual behavior

- Every SSE `token` event calls `setWorkspaceTimeline((current) => appendTextTimeline(current, event.data))` immediately (`packages/client/src/components/chat/HomeProfessorMariChat.tsx:3385-3392`) — no rafThrottle / buffer anywhere in the file.
- The component (~4,400 lines) contains **zero `React.memo`**, so every timeline item and history message re-renders per token.
- `CompactMarkdown` (`HomeProfessorMariChat.tsx:1149-1160`) calls `renderMarkdownBlocks()` (`packages/client/src/lib/markdown.tsx:404` — a recursive regex parse) inline in the render body with no memoization, over the full accumulated text, on every render.

Net cost per token is O(current transcript length); over a generation it is quadratic in reply length. On a slow model this is **minutes of sustained CPU on the client** — matching the user reports (phone warm, "worse with the marinara helper", worse with slow models).

## Steps to reproduce

1. On a phone (or with DevTools CPU throttling), open the Home screen and ask Professor Mari something that produces a long reply on a slow connection/model.
2. Observe sustained CPU/heat for the entire stream; profiler shows per-token full component render + markdown re-parse of the growing reply.
3. Compare with a normal conversation chat, which coalesces via `useThrottledStreamBuffer` (`packages/client/src/hooks/use-throttled-stream-buffer.ts:9-26`, cited to #2878) — same stream, far less render work.

## Environment

- Marinara Engine version: 2.4.2 (staging, 93738d9b1)
- Install type: source
- OS + version: reported on Android (browser client + Termux); reproducible on any platform with CPU profiling
[...]

PR #4727, opened 2026-08-07
fix(server): adaptive sandbox IPC polling instead of fixed 25ms intervals

### LINKED ISSUE #4706: [Issue]: Personal server-extension IPC: 25ms file polls x2 + 250ms watchdog + 1s heartbeat writes (~80 wakeups/s per extension, from boot)
## Summary

Each running personal **server** extension costs ~**80 timer wakeups per second plus one flash write per second, continuously from boot**: the host polls the sandbox output file every 25ms and runs a 250ms double-stat watchdog; the runner polls its input file every 25ms and heartbeat-writes every 1s. Flagged inside #4629 during the mobile-battery investigation ("related, happy to split") but never fixed — on a Termux self-host with a server extension installed this is the largest remaining passive drain.

## Expected behavior

Idle extensions cost near-zero wakeups; polling tightens only while traffic is flowing (startup handshake, storage bursts, stop), preserving today's latency where it matters.

## Actual behavior

- Host: `personal-server-extension-runtime.ts:394` — `outputPoller` `setInterval(25ms)` doing an fstat (+read) per tick; `:289-312` — watchdog `setInterval(250ms)` doing two `stat()`s per tick.
- Runner: `personal-extension-runner.mjs:226` — `inputPoller` `setInterval(25ms)`; `:228-230` — heartbeat `writeFile` every 1s (a flash write per second, forever).
- Conditional: only spawns per enabled+approved extension with `serverJs` (`runtime.ts:128`, `:239-242`) — zero-extension installs pay nothing.

## Steps to reproduce

1. Install/enable one personal server extension.
2. Observe the sandbox dir: input/output stat cadence + heartbeat mtime churn; ~80 wakeups/s in the two processes at full idle.

## Environment

- Marinara Engine version: staging d6275fa0b
- Install type: any with server extensions; battery-critical on Termux
- OS + version: any

Code trace, line references against staging d6275fa0b.

### Regression-safe fix (designed + red-teamed)

Keep the file transport untouched (it is the sandbox security design) — change only the **polling discipline**:

1. **Adaptive cadence, both sides.** Replace the fixed intervals with self-scheduling unref'd `setTimeout` chains: host output poll 25ms while HOT / 1000ms while idle; runner input poll 25ms HOT / 250ms idle. HOT := startup not settled, or bytes moved / a message was sent within the last ~3s. Startup handshake and storage bursts keep today's 25ms feel; steady-state drops from 40 stats/s to ~1/s per side. Latency contract verified: no storage/tool deadline exists to trip (`runner.mjs:39-45`, `runtime.ts:201-227` — no timeouts either direction; host→runner messages are exactly start/stop/storage-result), so the worst-case ≤1s first-message-after-silence delay is safe; stop responsiveness is bounded by the runner's 250ms idle cadence (rescale `CLEANUP_TIMEOUT_MS` 3000→3500 for margin).
2. **Heartbeat 1s → 5s** with the watchdog stale-threshold rescaled proportionally (same missed-beat multiple), cutting flash writes 5×. Watchdog cadence 250ms → 1s.
3. `fs.watch` only as an **optional fast-path** behind a runtime probe (inotify works on Termux home paths but is unreliable on FUSE/shared mounts) — never as the correctness mechanism; the poll chain remains the fallback.

**Amendments from the adversarial pass (must be in the implementation):**

- **`MAX_PROTOCOL_BYTES` false-kill:** `pollOutput` reads all available bytes and checks `outputBuffer.byteLength > 2MB` **before** line-splitting (`runtime.ts:336-344`). At 25ms that approximates a per-message cap; at a 1s idle cadence it measures *bytes accumulated across the idle gap* — two individually-legal ~1.5MB messages bursting after >3s of silence would batch to ~3MB in one read and kill a healthy extension. Move the cap to the **residual (incomplete-line) buffer after line extraction**, which is the actual single-message semantic. Burst-after-silence is exactly the traffic shape adaptive polling creates — this is not optional.
- **Final-drain ordering on close:** the "one last poll" must (a) respect the `pollingOutput` re-entrancy guard (`runtime.ts:325-326`) by awaiting any in-flight poll or looping to a clean pass, (b) complete **before** `outputHandle.close()` (`:400`) and before `sandbox.cleanup()` (`:408`) rm-rf's the dir, and (c) run **before** the `expectedStop` check so a drained `fatal` message (`:375`) doesn't suppress or get suppressed by the generic exit fallback (`:403-406`) in the wrong order.
- **Self-scheduling chains must reassign the stored timer handle every tick** (`active.outputPoller` / `inputPoller`) — clearing a stale handle from a previous tick would leave the live timer running and hold the process alive.

Related: #4629 (the client-side polling half, fixed in #4643); #4704 / #4705 (remaining autonomous-polling costs).

## Template check

kolacheee:
> [!IMPORTANT]

## Linked issue

Closes #4706

## Why this change

- Each running personal **server** extension cost ~**80 timer wakeups per second plus one flash write per second, continuously from boot**: the host polled the sandbox output file every 25 ms and ran a 250 ms double-stat watchdog, while the runner polled its input file every 25 ms and heartbeat-wrote every second. On a phone-hosted (Termux) install with a server extension enabled this was the largest remaining passive drain from the mobile-battery investigation (#4628–#4631, #4704/#4705, #4707).
- The file transport itself is untouched — it *is* the sandbox security design. Only the polling discipline changes.

## What changed

- **Adaptive cadence on both sides** (new pure module [`sandbox-protocol.ts`](packages/server/src/services/extensions/sandbox-protocol.ts) + [`personal-server-extension-runtime.ts`](packages/server/src/services/extensions/personal-server-extension-runtime.ts) + [`personal-extension-runner.mjs`](packages/server/src/assets/personal-extension-runner.mjs)): the fixed intervals are now self-scheduling timeout chains that poll at the old 25 ms only while traffic is flowing (startup handshake unsettled, or bytes/messages moved within the last 3 s) and drop to **1 s (host) / 250 ms (runner)** at idle. Each side's `send()` flips the other direction back to hot, since a reply is expected — storage round-trips and the stop handshake keep today's latency. Steady-state wakeups drop from ~80/s to ~5/s per extension.
- **Heartbeat 1 s → 5 s** (5× fewer flash writes) with the host's unresponsive threshold rescaled 5 s → 25 s — the same five-missed-beats multiple as before — and the watchdog cadence 250 ms → 1 s. `CLEANUP_TIMEOUT_MS` gains headroom (3 s → 3.5 s) for the runner's idle stop pickup.
- **Per-message size cap** (red-team amendment from the issue): the host's `MAX_PROTOCOL_BYTES` check previously applied to the whole buffered chunk — a fair approximation at 25 ms, but at an idle cadence the buffer batches everything that arrived across the silence, so several individually-legal messages bursting after a quiet spell would have been killed as one "oversized message". Extraction now caps each message individually (newline index per line, residual buffer after extraction) via the pure `extractProtocolLines`.
- **Final drain on close** (red-team amendment): `handleClose` now waits out any in-flight poll and runs one last read **before** closing the file handle, before the `expectedStop` check (so a drained `fatal` message attributes the exit correctly instead of racing the generic fallback), and before `sandbox.cleanup()` removes the directory.
- **Timer-handle hygiene** (red-team amendment): the stored poller handles are reassigned on every tick of the chains — clearing a stale handle from a previous tick can't leave a live timer running. The runner's heartbeat interval remains the process-liveness anchor.
- The runner inlines its constants (it's a standalone sandbox asset that cannot import server modules) with a sync comment; the regression pins the pairing.
An adversarial review pass on this diff (verifier ran an A/B harness driving the **real runner** against a faithful host transcription) caught and fixed two things:

- The new close-time drain could resolve the startup handshake from a child that had already crashed right after `ready` (unhandled rejection → `fatal` → exit): `load()` would then insert the dead sandbox into the active set, report it as **running**, and a later stop-all would fail on its deleted input file — aborting shutdown of the *healthy* extensions (reproduced 6/6 pre-fix). `load()` now refuses to activate a sandbox whose child has exited or is marked stopped (surfacing the drained `fatal`'s message instead), and a failed stop-write no longer aborts the kill/cleanup path.
- Two of the regression's "no fixed 25 ms interval" source pins were regexes that could not match the code they forbid (a negated character class can't cross the `)` in `() => void …`) — assertions that could never fail. Replaced with exact-substring forbids plus positive pins asserting the chains exist *and* their delay expressions actually consult the idle constants.

- **New regression** ([`extension-ipc.regression.ts`](scripts/regressions/extension-ipc.regression.ts), wired into `regression:issues`): extraction semantics (a legal burst is not oversized; per-line and residual caps; split-message reassembly), cadence floors (no 25 ms idle timer or 250 ms watchdog can quietly return), the five-missed-beats multiple, host/runner constant sync via source scan — plus a **live end-to-end of the real runner under plain `node`**: start handshake, extension log delivery, a stop sent after the hot window lapses is picked up within the idle-cadence bound, clean exit 0.

## Validation

### Manual verification notes

- `pnpm check` and the full `regression:issues` lane pass locally, including the new IPC regression's live runner pass (real `personal-extension-runner.mjs` under plain `node`: handshake, log delivery, stop-after-silence within the idle bound, exit 0).
- The built `dist/assets/personal-extension-runner.mjs` matches source byte-for-byte after `pnpm build` (the build copies assets).

Still to verify manually (contributor to-do — Windows cannot run the OS-sandboxed spawn, so the full host+sandbox pair needs a Linux/Termux pass):
- Manually verify on Termux/Linux with a real server extension: it starts (`Sandboxed … with …` log), storage get/patch round-trips work, logs stream, disable/enable restarts it cleanly, and the sandbox dir shows the new cadences (heartbeat mtime every ~5 s; input/output stats quiet at idle).
- Manually verify an unresponsive extension (e.g. busy-loop in extension code) is still killed by the watchdog within ~25 s.

## Docs and release impact

- [x] No docs changes needed

No user-facing docs impact — internal IPC cadence; no configuration surface changed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved extension communication reliability when messages arrive together or are split across reads.
  - Prevented oversized messages from disrupting communication.
  - Improved startup and shutdown handling, including final output delivery and more tolerant cleanup.
  - Reduced unnecessary background activity while maintaining responsive input and output handling.
  - Improved heartbeat and watchdog behavior for more reliable process monitoring.

- **Tests**
  - Added comprehensive regression coverage for communication, polling, message limits, and complete extension workflows.
  - Expanded the regression suite to include extension communication checks.


kolacheee:
@coderabbitai review

kolacheee:
@coderabbitai review

kolacheee:
@coderabbitai review

PR #4736, opened 2026-08-07
fix: address assigned issue sweep

### LINKED ISSUE #4718: [Feature]: Per-chat image style override for custom image agents (follow-up to #4682/#4686)

Follow-up to #4682 / PR #4686 (per-chat image connection override + snapshot button for custom image agents), requested by the same user after using the shipped feature:

> "This works great, i can change image connection and create a snapshot with no issues — thanks again. If you ever get back to this (what is here already is more than enough) do you think it'd be possible to change the image **style** too? Like the illustrator has. I tried to change it via the agent prompt but it isn't always consistent."

Today the chat-level **Image Style** select applies to custom-agent images — but it's **chat-global**: one style profile for the Illustrator *and* every custom image agent in that chat. There's no way to render one custom agent in a different style than the chat's default (e.g. a "polaroid snapshots" agent alongside a painterly Illustrator), and steering style through the agent's prompt template is unreliable, as the requester found.

## Proposed solution

Extend the per-chat, per-agent map that #4686 introduced — the shape was designed for exactly this kind of growth:

- `ChatMetadata.customAgentImageSettings[agentType]` gains `styleProfileId?: string | null` alongside the existing `imageConnectionId`.
- The same custom-agent card in **Chat Settings → Agents** gains an **Image Style** select (reusing the Illustrator style select pattern, with **Chat default** as the empty option).
- Resolution: when set, the per-agent `styleProfileId` wins over the chat-wide `imageStyleProfileId` for that agent's `image_prompt` results, then falls through the existing chain (chat-wide → connection's style profile → default). Both handler twins (main `generate.routes.ts` and the retry route) need the same read, mirroring how the connection override was wired.
- A deleted style profile behaves like the chat-wide key does today: resolve-time fallback down the chain, no cleanup machinery needed. (Note: unlike `*ConnectionId` keys there is no auto-sweeper for style ids — a stale entry is inert and falls through, same as `chatMeta.imageStyleProfileId` already does.)

Scope: same shape and size as the merged #4686 connection override (map extension + one select + a read at the two style-resolution sites) — comfortably within the "keep it small-scale" guidance from the parent request.

## Alternatives considered

- **Prompt-template styling** (status quo): the requester tried it; models apply it inconsistently, and it can't leverage the style-profile system (positive/negative prompt fragments, provider-specific handling).
- **Duplicating agents per style**: the exact copy-paste churn #4682 was filed to eliminate — reintroducing it for styles would be a regression of intent.
- **Per-agent global (not per-chat) style in agent settings**: doesn't cover the "same agent, different style per chat" case, which is the point of the per-chat map.

## Additional context

Requested by the original #4682 reporter after validating PR #4686 in real use ("what's here is more than enough" — this is optional polish, explicitly low-priority). Queued behind current priority work (#4703, #4707–#4709); assigning to track ownership.

## Template check

### LINKED ISSUE #4720: [Feature]: Attributed fetch + per-extension traffic visibility for full-page personal extensions

In #4703, a phone-hosted instance served ~204 requests/min (multi-GB/hour) for two hours-plus, and the source was only identifiable by downloading and dissecting a 266 MB HAR: a **full-page personal extension** (a memo/notes board) was re-scanning every chat's messages (`GET /chats/<id>/messages?limit=100` × 18 chats, back-to-back sweeps with ~0.15 s pause between sweeps) and its traffic was indistinguishable from Marinara's own.

The full-page tier is an explicit "fully trust" grant, and that's the right model — but trust covers *intent*, not *bugs*. The memo extension's author almost certainly never intended an unthrottled full-table rescan; the user had no way to see that an extension was responsible (nothing in the UI attributes traffic), and the symptom presented as "Marinara drains my battery / eats my data plan," landing on this issue tracker.

Mechanically, the gap is visible in the injector: full-page extensions receive a frozen host API (`__marinaraRunFullPageExtension` → `FullPageExtensionApi` in `packages/client/src/components/layout/PersonalExtensionInjector.tsx`) that provides attributed **storage, logging, and managed timers** — but no fetch. Extensions reach the API through the page's raw `window.fetch`, so nothing counts, attributes, or bounds their requests.

## Proposed solution

Layered visibility first, enforcement only where it doesn't break the trust model:

1. **Attributed `api.fetch` on the full-page host API.** Extend the existing frozen `FullPageExtensionApi` with a `fetch` that wraps `window.fetch`, tagging and counting requests (count + transferred bytes) per extension id — same pattern as the API's managed `setTimeout`/`setInterval`. Extension-authoring docs should present it as the expected way to call the local API (raw `window.fetch` keeps working; full-page code can always reach it — this is visibility, not a jail).
2. **Per-extension traffic counters in the UI.** Show request count / bytes (session totals and a recent-rate figure) per extension in Settings → Personal Extensions, next to the existing enable toggles. A sustained-high-rate badge (e.g. > 60 req/min for several minutes) turns "my phone is warm" into "the memo extension made 12,000 requests this hour" — the diagnosis in #4703 took a 266 MB HAR; this makes it one glance.
3. **Server-side anomaly notice (stretch, catches raw-fetch extensions too).** The server already knows request rates per route; a cheap counter on the heavy read endpoints (message lists) that logs a warning — and optionally surfaces a one-time client toast — when a single client sustains a pathological read rate (e.g. > 100 message-list reads/min for 5+ minutes), with wording that points at personal extensions as a likely cause. This is attribution-free, so it works even for extensions that bypass `api.fetch`, and it is a diagnosis aid rather than a limiter.

Explicitly **not** proposed: hard rate limits on full-page extensions. Their fetches are same-origin page fetches — indistinguishable from the app's own without cooperation — and the tier's consent screen already frames it as full trust. The #4703 failure was invisibility, not malice; bounding a trusted-but-buggy extension is the extension author's fix once the user can see the problem.

## Alternatives considered

- **Hard per-extension budgets enforced by the host**: only enforceable for code that cooperates (uses `api.fetch`); a non-cooperating script escapes to `window.fetch`, so enforcement would give false confidence while breaking legitimate bursty extensions. Visibility gets ~all of the value with none of the false promises.
- **Moving chat/message reads into a capability-gated host API** (like the sandboxed tier's context snapshots): architecturally cleanest, but it's a breaking change for every existing full-page extension and a much bigger project than this problem needs.
- **Doing nothing** (status quo): the next runaway extension lands as another multi-week "Marinara drains my battery" investigation. #4629 (the runtime's own 2 s poll) and #4703 (this) suggest extension-adjacent traffic already has a track record.

## Additional context

- Root-cause analysis with the HAR breakdown: https://github.com/Pasta-Devs/Marinara-Engine/issues/4703#issuecomment-5219132238 (24 sweeps × 18 chats in 125 s; bracketed by `GET/PATCH /api/personal-extensions/<id>/storage`).
- The sandboxed browser tier is unaffected (opaque-origin iframe, no network); the server tier is OS-sandboxed without network. This is specifically about the full-page tier.
- Related history: #4629 (extension runtime polling), #4628/#4630 (battery investigation the reporter's symptom initially landed in).

## Template check

### LINKED ISSUE #4724: [Issue]: Tracker panel's hardcoded pageSize-20 observer hijacks the shared chat-messages query when messages-per-page differs (same class as #4721)
## Summary

The tracker panel's sprite-expression lookup mounts a `useChatMessages` observer with a hardcoded `pageSize: 20` on the **same React Query key** the chat transcript uses with the user's messages-per-page — the same options mis-share fixed for the Chat Settings drawer in #4721, from a different observer. When messages-per-page is set to anything other than 20, whichever observer rendered last owns the shared query's fetch closures, and the transcript's pagination runs with the wrong page size.

## Expected behavior

The tracker panel's message reads have no effect on the transcript's pagination: refetches fetch messages-per-page-sized pages and "Load More" reflects the real history.

## Actual behavior

With messages-per-page ≠ 20 and the tracker panel mounted with sprite-expression lookup enabled:

- When the tracker's observer wins the options race, an invalidation-driven refetch fetches the newest page at `limit=20` instead of the user's setting.
- `ChatArea`'s `hasNextPage` is computed from **its** observer options, so `getNextPageParam` evaluates a 20-row page against `pageSize = 50` (for example): `20 < 50` → `undefined` → **"Load More" disappears** with older history still present. The defensive newest-page trim doesn't compensate — it only trims pages *longer* than the transcript's page size.

Code locations (current staging):

- `packages/client/src/features/tracker-panel/hooks/use-tracker-panel-model.ts:74` — `useChatMessages(activeChatId, 20, spriteExpressionLookupEnabled)`
- `packages/client/src/components/chat/ChatArea.tsx:751` — transcript observer at the user's `messagesPerPage` (0–500 in Settings; default 20 — which is why the default configuration masks the bug)
- `packages/client/src/hooks/use-chats.ts` — `useChatMessages` keys by chat id only; page size lives in the `queryFn`/`getNextPageParam` closures

## Steps to reproduce

1. Set Messages-per-page to 50; open a chat with well over 50 messages and sprite-expression lookup active (tracker panel mounted).
2. Trigger a transcript refetch (complete a generation, or any messages invalidation).
3. Depending on which observer set options last, the refetch fetches `limit=20` and "Load More" can vanish although older history exists.

## Environment

- Marinara Engine version: staging (c4f15f373)
- Install type: source
- OS + version: any (React Query v5 semantics)

Found by the adversarial review sweep run on the #4721 fix (the sweep lens explicitly looked for remaining instances of the class). Confirmed against `@tanstack/query-core@5.100.9`: `QueryObserver.setOptions` writes the shared query's options with no `enabled` guard, so the tracker's observer participates even while its `enabled` gate is false.

## Additional context

Same fix shape as #4721 (merged fix pending in its PR): the sprite-expression lookup is a read-only newest-window consumer — it should read through the limit-keyed `useChatMessagePeek` (or an equivalent dedicated key) rather than observing the shared transcript query. When fixed, extend the #4721 source-scan regression pin (currently guarding only `ChatSettingsDrawer.tsx`) to cover `use-tracker-panel-model.ts`, or generalize it to allow only the sanctioned transcript observers on the shared key.

Kept out of the #4721 PR deliberately — pre-existing, different surface, and the tracker's consumer shape (`MessageWithSwipes` usage for sprite lookups) needs its own look before swapping the data source.

## Template check

### LINKED ISSUE #4726: [Issue]: Character messages lose speaker attribution in individual group chats
## Summary
In Group Chat Conversations using individual mode, character chat history omits speaker names and timestamps. This causes characters to interpret all prior messages as if they were written by the user.

## Expected behavior
Each chat history entry should include its timestamp and speaker name, such as `[12:16] Angie: just drowning in civics homework lol you?`, so responding characters can identify who said each message.

## Actual behavior
Character messages appear as unattributed text in Chat History, causing characters to treat every message as a user message and respond to the user instead of the appropriate character.

## Steps to reproduce
1. Start a Group Chat Conversation in individual mode with multiple characters.
2. Send a user message and allow multiple characters to reply.
3. Inspect the Chat History supplied to a later character, such as through the peek prompt.
4. Observe that prior character replies lack timestamps and speaker names and are interpreted as user messages.

## Environment
- Marinara Engine version: Not provided
- Install type: Not provided
- OS + version: Not provided
- Browser or app shell: Not provided

No attachments provided. The reported history contains bare character messages such as `just drowning in civics homework lol you?` instead of attributed entries such as `[12:16] Angie: just drowning in civics homework lol you?`.

## Additional context
A second user reported noticing similar speaker-attribution behavior in a single direct message.

Discord thread: Group Chat Conversations - Character Text Context
Discord thread link: https://discord.com/channels/1417099416812392641/1535322819573850202
Requested by: Marinara (marinara_spaghetti, Discord ID 169561011928432653)

<details>
<summary>Discord thread transcript (3 messages)</summary>

```text
[2026-08-07T16:24:45.105Z] AgentX: When using Group Chat Conversations in individual mode, character Chat History doesn't contain who said what, so the characters think everything everybody said came from User. Recommend inserting the character names to make the AI understand the context.

For example, it is current like:
` ` `Chat History - User
[12:15] User: hey all, anything going on today?
` ` `

Then for character replies (via peek prompt)

` ` `Chat History - User 
just drowning in civics homework lol you?

civics is the worse, too much reading

it's just the civics paperwork, it's not too bad

...
` ` `

None of the characters know which character said what. The Character Chat History should be like the User one with `[Time] Name: text`.

So something like 
` ` `Chat History - User 
[12:16] Angie: just drowning in civics homework lol you?

[12:17] Tom: civics is the worse, too much reading

[12:17] Bob: it's just the civics paperwork, it's not too bad

...
` ` `

When when character 4 responds, they know that Angie is the one who is drowning in civics homework, not user. Right now, everyone thinks user posted everything, so they reply as if replying to user instead of the other characters.
[Discord message: https://discord.com/channels/1417099416812392641/1535322819573850202/1535322819573850202]

[2026-08-07T16:53:12.486Z] DSBwizzard: Dang nice catch
[Discord message: https://discord.com/channels/1417099416812392641/1535322819573850202/1535329980848930886]

[2026-08-07T17:02:42.634Z] Indigo: this is a good catch I've noticed something like this too in a single DM before
[Discord message: https://discord.com/channels/1417099416812392641/1535322819573850202/1535332372222840863]
```

</details>

## Template check

### LINKED ISSUE #4730: Server crashes with "JavaScript heap out of memory" on Android/Termux every few messages in large installs
From a user report (Android, Termux, staging): sending a message shows "Failed to fetch" in the chat UI, and the Termux terminal shows the backend dying with:

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

It happens every few messages. Their GC trace is the important clue — about 100 seconds after a fresh restart, a full Mark-Compact still had **~1010 MB live**, against a heap limit of ~1 GB:

```
[14798:...] 100259 ms: Mark-Compact 1036.4 (1100.0) -> 1010.7 (1090.6) MB ... allocation failure; scavenge might not succeed
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

So this is not a leak — the install's data no longer fits in the heap V8 picked on that phone, and normal per-message allocation spikes push it over the edge.

## Three things stack up

1. **The file-native store keeps every table in memory.** All of `data/storage/tables/*.json` is loaded at boot and stays resident (`packages/server/src/db/file-backed-store.ts`). A long-running roleplay install (large message history, swipes, summary history, long-term memory) produces a permanent live set of many hundreds of MB.

2. **No launcher sets a heap size.** There is no `--max-old-space-size` / `NODE_OPTIONS` anywhere in `start.sh`, `start.bat`, `start-termux.sh`, or the Android docs. Node falls back to V8's physical-memory heuristic, which on phones lands around **1 GB** — far less than desktops get, on exactly the platform where installs are hardest to debug.

3. **Per-message allocation spikes are proportional to total table size, not message size.**
   - Every insert/update shallow-copies the **entire table array**: `const nextRows = target.map(cloneRow)` (`file-backed-store.ts:1100`, also the update path at `:1144`). On a messages table with tens of thousands of rows, that's a full-array copy several times per send. (Pattern introduced in `37bb8f758`, "Enforce file-native unique keys".)
   - The debounced flush serializes the **whole dirty table into one contiguous string**: `JSON.stringify(rows)` (`file-backed-store.ts:1445` in `saveFileSnapshots`). For a big messages table that is a single allocation of hundreds of MB.

   With a ~800 MB baseline and a ~1 GB cap, whichever send happens to coincide with a flush dies. That's why it fails "every few messages" instead of deterministically.

## Try it yourself

On any machine, artificially shrink the heap to phone scale and use a large profile (or generate one — a chat with tens of thousands of messages):

```bash
NODE_OPTIONS="--max-old-space-size=1024" pnpm start
```

Send messages in a large roleplay chat. Within a few sends, the flush after a generation OOMs the server the same way the Termux trace shows.

## Suggested directions

1. **Cheapest, biggest win: set an explicit heap size in the launchers.** `start.sh` / `start.bat` / `start-termux.sh` can export `NODE_OPTIONS="--max-old-space-size=..."` (respecting a user-provided value), sized from available RAM. Even a flat 2048 on Android would have absorbed this user's workload.
2. **Stream the flush.** Write table snapshots as JSON Lines (row-by-row through the write stream) instead of one `JSON.stringify` of the whole table. The profile-ZIP exporter already writes JSONL per table (`backup.routes.ts`), so there's precedent and a reader for the format.
3. **Revisit the full-table clone per write.** Copy-on-write of the whole array on every insert/update is O(table) allocation on hot per-send paths. Cloning only the affected rows (or sharding messages per chat) would bound it.

## Workaround for affected users (worth adding to docs/TROUBLESHOOTING.md)

In Termux, before launching:

```bash
NODE_OPTIONS="--max-old-space-size=2048" PORT=7869 bash ./start.sh
```

(Use 3072 on phones with 8 GB+ RAM. Add the `export` to `~/.bashrc` to make it permanent.)

### LINKED ISSUE #4731: [Issue]: Prompt shown twice after API 403 error
## Summary
In Roleplay mode, a sent prompt is duplicated in chat history when the API returns a 403 error. The duplicate entries appear linked because deleting the new copy removes both.

## Expected behavior
The sent prompt should appear only once in chat history when the API returns a 403 error, and deleting one message should not unexpectedly remove another entry.

## Actual behavior
After the API returns a 403 error, the sent prompt appears twice in chat history. Deleting the new copy deletes both messages, while the writing box remains empty.

## Steps to reproduce
1. Open a chat in Roleplay mode.
2. Send a prompt that results in an API 403 error.
3. Observe that the sent prompt appears twice in chat history.
4. Delete the new copy and observe that both entries are removed.

## Environment
- Marinara Engine version: 2.4.1 (stable)
- Install type: termux
- OS + version: Android 16
- Browser or app shell: Not provided

Not provided in the Discord context.

## Additional context
A maintainer asked whether the duplication also appears in logs, but no log details were provided.

Discord thread: Bug report - Prompt shown twice on API error
Discord thread link: https://discord.com/channels/1417099416812392641/1535365745775484998
Requested by: Marinara (marinara_spaghetti, Discord ID 169561011928432653)

<details>
<summary>Discord thread transcript (3 messages)</summary>

```text
[2026-08-07T19:15:19.509Z] Fussel: ME version: 2.4.1 (stable)
Installation method: Termux (Android 16)
Chatting mode: Roleplay

I noticed that when I send my prompt and the API returns a 403 error, the already sent message then appears twice in the chat history. Deleting the new copy deletes both messages from the chat history. The writing box stays empty though.
[Discord message: https://discord.com/channels/1417099416812392641/1535365745775484998/1535365745775484998]

[2026-08-07T19:15:37.224Z] Fussel: Bug report - Prompt shown twice on API error
[Discord message: https://discord.com/channels/1417099416812392641/1535365745775484998/1535365820077842524]

[2026-08-07T19:59:33.157Z] Marinara: In logs?
[Discord message: https://discord.com/channels/1417099416812392641/1535365745775484998/1535376875981897778]
```

</details>

## Template check

### LINKED ISSUE #4733: [Issue]: Echo Chamber size does not persist across chat navigation or restart
## Description

In Roleplay mode, the Echo Chamber size selected for a chat is not remembered. The size resets after restarting Marinara Engine or after switching to another chat and returning.

## Steps to reproduce

1. Open a Roleplay chat with Echo Chamber available.
2. Change the Echo Chamber size.
3. Either switch to another chat and return, or restart Marinara Engine and reopen the chat.
4. Observe that the selected size has been lost.

## Expected behavior

The Roleplay chat should persist its selected Echo Chamber size and restore it after navigating between chats or restarting Marinara Engine.

## Actual behavior

The selected Echo Chamber size resets and must be configured again.

### LINKED ISSUE #4734: [Issue]: Narrative Director push-story instruction is not applied in individual group chats
## Description

In Roleplay mode, when a group chat uses Individual response mode and has a Narrative Director agent added, selecting **Push Story Naturally** or **Push Story Randomly** does not append the instruction to the next message. This fails whether the next message is sent manually or generated through the queue.

The behavior is also incorrectly affected regardless of whether **Secret Plot** is enabled or disabled.

## Steps to reproduce

1. Open a Roleplay group chat.
2. Set the group response mode to **Individual**.
3. Add a Narrative Director agent.
4. Select **Push Story Naturally** or **Push Story Randomly**.
5. Send the next message manually or let the queue produce it.
6. Observe that the selected push-story instruction is not appended or applied.
7. Repeat with **Secret Plot** enabled and disabled.

## Expected behavior

The selected Narrative Director push-story instruction should be appended to and applied to the next message in an Individual-mode group chat, for both manual sends and queued generations. It should work whether Secret Plot is enabled or disabled.

## Actual behavior

The next message is produced without the selected push-story instruction.

SpicyMarinara:
## Why

This sweep addresses every currently open issue assigned to SpicyMarinara. The fixes keep each concern narrow: extend existing settings and prompt paths, isolate one query observer, add one launcher guard, and preserve existing trust/runtime behavior.

## What changed

- Added per-chat style overrides for custom image agents with chat/default fallback in main and retry generation.
- Added attributed `marinara.fetch`, session traffic totals, recent rate, and sustained high-traffic visibility for full-page extensions.
- Moved tracker sprite lookups onto the limit-keyed message peek query.
- Preserved speaker attribution in individual Conversation group history.
- Added a user-overridable 2 GB Termux Node heap default for large profiles.
- Locked failed-generation prompt reconciliation to durable submission IDs.
- Committed Echo Chamber size when pointer capture is lost.
- Preserved Narrative Director push-story instructions at each individual responder boundary.
- Fixed server-extension `onCleanup` storage writes exposed by the extension security proof.

## Issues

Closes #4718
Closes #4720
Closes #4724
Closes #4726
Closes #4730
Closes #4731
Closes #4733
Closes #4734

## Automated evidence

- `pnpm check`
- `pnpm regression:prompt`
- Open-issue and Roleplay streaming regression scripts
- `pnpm regression:extensions-security`
- `pnpm regression:docs`
- `pnpm version:check`
- `bash -n start-termux.sh`
- Full Playwright smoke run: 200 passed, 99 skipped, one unrelated saturated-run mobile composer timeout; its isolated rerun passed in 5 seconds. The desktop Echo Chamber persistence case passed in the full run.

## Manual verification

- [ ] Manually verify custom image-agent style selection and deleted-profile fallback in Chat Settings.
- [ ] Manually verify attributed extension traffic totals and the sustained-rate badge.
- [ ] Manually verify Narrative Director natural/random pushes for manual and queued individual group responses with Secret Plot on and off.
- [ ] Manually verify Termux startup and a large-profile generation on an affected Android device.

## Notes

Validation boxes remain unchecked for human verification. English extension documentation was updated; #4737 tracks the required docs-i18n mirror.

## Summary by CodeRabbit

- **New Features**
  - Custom image agents support per-chat image connection and style overrides.
  - Full-page extensions support tracked fetch requests, traffic metrics, and sustained high-rate warnings.
  - Conversation groups retain speaker labels and receive Narrative Director guidance.
  - Termux startup applies a safe memory limit while preserving custom settings.

- **Bug Fixes**
  - Improved tracker pagination, failed-generation recovery, resize persistence, prompt deduplication, instruction retention, and shutdown storage writes.
  - Custom image generation now honors configured style profiles.

- **Documentation**
  - Documented extension fetch usage and traffic attribution.


SpicyMarinara:
@coderabbitai review

Issue #898, opened 2026-05-15
[Issue]: Professor Mari tool calls can appear successful without completing


cha1latte:
### Summary

Professor Mari tool calls can appear to run, or Mari may say she will run them, without the requested card fetch/update actually completing. Reports include card fetches stalling after manual saves, fetch snackbars appearing without follow-up progress, and update commands landing as plain conversation text when command payloads contain square brackets.

### Expected behavior

Professor Mari should reliably execute requested card/persona fetch and update actions, then continue the conversation with the fetched or updated data available.

### Actual behavior

Professor Mari may say she will fetch or update a card, or the UI may show that a card was fetched, but the conversation then stalls or the requested data/action does not complete. In another reported case, update commands containing square brackets are emitted as plain conversation text instead of being parsed as commands.

### Steps to reproduce

Known reported cases:

1. Open a Conversation mode chat with Professor Mari.
2. Ask Professor Mari to fetch or work with a character card/persona.
3. Manually edit and save that card.
4. In the same Professor Mari conversation, ask her to fetch the card/persona again.
5. Observe that the action appears to happen, but the card/persona data does not load.
6. Open a fresh Professor Mari conversation and try the same fetch.
7. Observe that the fresh conversation can fetch until another manual save happens.

Additional reports from the same Discord thread:

- Asking Professor Mari to copy and modify a character card can show a fetched-card snackbar and then hang.
- Character update commands containing square brackets in quoted field values can fail parsing and appear as normal conversation output.
- The original reporter described the broader symptom as Professor Mari saying she will fetch/update a card but not actually doing it.

### Marinara Engine version

1.5.9

### Install type

Docker

### OS + version

Docker environment; one additional reporter mentioned Windows 11.

### Browser or app shell

Not specified.

### Screenshots or video

Not a visual/UI issue.

### Logs

Not provided.

### Additional context

Discord source: https://discord.com/channels/1417099416812392641/1480379733173731458/threads/1504204832263901264

Reporter used Conversation mode through NanoGPT and tried GLM 5.1 and Deepseek V4 Pro.

Related external patch shared in the Discord thread: https://github.com/timwu/Marinara-Engine/commit/605f6d5473a94cc81a996522adba2b4ebd2fe50d

That patch proposes two changes: quote-aware parsing for character update commands containing nested square brackets, and an autonomous follow-up generation turn after Professor Mari executes a fetch command.

### Template check

[...]

PR #4771, opened 2026-08-08
feat(mari): stable, capped Home-Mari name lists (relocate out of the system prefix) (#4768)

### LINKED ISSUE #4768: [Feature]: Replace flat entity name injection with a stable, searchable mechanism — reduce noise and enable prompt caching

Professor Mari's conversation mode has two interconnected problems with the `<available_names>` block in her system message:

1. **Noise at scale:** Every generation injects ALL entity names (characters, personas, lorebooks, chats, presets) as a flat, unfiltered list. For users with large collections, this becomes thousands of tokens of context noise that the model cannot meaningfully process.
  
2. **Cache invalidation:** The name list is rebuilt from the DB on every generation (`professor-mari-prompt-context.ts:30-73`). Even when no entities change, DB query ordering differences and re-serialization produce a different string, invalidating provider-side prompt caching.
  

Both problems stem from the same design: a flat, dynamically-rebuilt name list injected into the system message on every generation.

### Scale of the name list problem

| Collection size | Name list tokens (est.) | % of 128K context |
| --- | --- | --- |
| 50 characters | ~240 tokens | 0.2% |
| 200 characters | ~960 tokens | 0.7% |
| 1,000 characters | ~4,740 tokens | 3.7% |
| 1,000 chars + personas + lorebooks + chats + presets | ~5,500+ tokens | 4.3%+ |

These are comma-separated name lists wrapped in `<available_names>` XML blocks — no descriptions, no metadata, just names. The model receives them regardless of whether the user's current conversation involves any of these entities.

### Cache impact

| Provider | Cache granularity | Cost of invalidation |
| --- | --- | --- |
| Anthropic | Prompt prefix (min 1,024 tokens for Claude) | Full re-prompt cost on cache miss |
| OpenAI | Prompt prefix (automatic for repeated prefixes) | Full re-prompt cost on cache miss |
| Google | Context caching (explicit cache creation) | N/A (manual caching) |

For a user with 1,000 characters, the system message is ~13,740 tokens. On Anthropic, this means ~13,740 tokens are re-priced on every generation instead of cache-hitting at the reduced rate. Over a conversation with 50 generations, that's ~687,000 tokens of unnecessary full-price billing.

The system message changes between generations because:

- `DEFAULT_CONVERSATION_PROMPT` and `MARI_ASSISTANT_PROMPT` are stable
- `<available_names>` blocks are **rebuilt from DB every generation** (lines 30-73)
- `<loaded_context>` changes when new data is fetched via `[fetch:]`
- DB queries may return rows in different orders across calls
- Any entity change (add, edit, delete) between generations changes the entire block

### Why this is a problem

1. **The model already has a fetch mechanism.** Mari's `<data_access>` instructions already tell her to use `[fetch: type="character", name="Luna"]` to retrieve full card data. The name list exists so Mari knows what's *available* to fetch — but a flat list of 1,000 names is not a useful lookup mechanism.
  
2. **No filtering or search.** The model receives ALL names regardless of relevance. If the user asks about Luna, the model still sees 999 other character names it doesn't need.
  
3. **Compounding cost.** The noise problem and the caching problem compound each other — large name lists are both wasteful to inject AND wasteful to rebuild.
  

## Proposed solution

**Option A: Replace with searchable tool (recommended)**

Replace the flat `<available_names>` injection with a lightweight search tool that Mari can call when she needs to discover entities:

```
[list_characters: filter=""] → returns first 20 character names
[list_characters: filter="luna"] → returns matching characters
[search_characters: query="fortune teller"] → searches descriptions/tags
[list_lorebooks: filter=""] → returns lorebook names
```

This eliminates both problems: no names in the system message (no noise, no cache invalidation). The existing `[fetch:]` command and `<data_access>` instructions provide the retrieval mechanism; this adds discovery.

What already exists and can be reused:

- The `[fetch:]` command already retrieves full entity data by name
- The `<data_access>` instruction block already teaches Mari the fetch pattern
- The workspace assistant's `app_data` tool schema covers CRUD operations
- The DB query infrastructure (`chars.list()`, `lorebooksStore.list()`, etc.) already exists

What would need to be added:

- A `[list_characters]` / `[search_characters]` command handler in `character-commands.ts`
- A reduced default name list (e.g., first 20 characters, or recently used) instead of all entities
- Instructions in MARI_ASSISTANT_PROMPT teaching Mari to use the search tool

**Option B: Cap + cache (minimal change)**

Keep the `<available_names>` blocks but:

1. Cap at 50 names per category with a note: "and 950 more — use [list_characters: filter='...'] to search"
2. Cache the serialized string in memory; invalidate on entity CRUD
3. Sort alphabetically before joining for deterministic serialization

This reduces noise from ~4,740 tokens to ~240 tokens and ensures cache hits when no entities change. Lower blast radius than Option A.

**Option C: Stable prefix isolation**

Move `<available_names>` and `<loaded_context>` to separate messages at stable positions. This isolates changing content from stable content (`MARI_ASSISTANT_PROMPT`, ~9,000 tokens), allowing the provider to cache the stable prefix. The changing blocks would still miss cache, but the stable prefix (~9K tokens) would cache consistently.

**Recommended: Option A + B combined.** Replace the flat list with a searchable tool for new users, with a capped fallback for backward compatibility. Cache the reduced list deterministically.

## Alternatives considered

- **Keep injecting all names** — works for small collections (<100 entities) but doesn't scale. A user with 1,000 characters sends ~4,740 tokens of noise every generation.
- **Inject only recently used entities** — reduces noise but loses discoverability. Mari wouldn't know about characters the user hasn't chatted with recently.
- **Inject names in a separate system message** — doesn't reduce token usage, just reorganizes the prompt. The core problem (flat list of 1,000 names) remains. (See Option C for a variant that does help with caching.)
- **Truncate the name list** — hides the problem without solving it. Mari would silently lose access to entities beyond the truncation point.
- **Accept the cache miss** — works for small collections but wastes money at scale. The cost compounds with every generation.

## Related issues

- **#4240 (Pre-processing agent context omitted from first generation):** User reported that "Memory Recall or Secret Plot get injected right in the middle, right before the chat history. Since these are dynamic blocks, the cache breaks immediately after the system instructions." Same class of problem — dynamic blocks in the system message invalidating prompt caching.
- Confirms that dynamic prompt content directly breaks provider-side caching.
- **#2302 (Full prompt assembly runs twice every turn):** `assembleGenerationPrompt` invoked twice per turn, redoing all expensive work including storage reads and DB queries. The name list rebuild is one instance of this broader redundancy problem.
- **#2778 (Helper Functions for Professor Mari):** LLM providers struggle formatting Mari's DB commands, racking up tokens. Shows the pattern of Mari's conversation mode needing better tooling for entity operations.
- **#3377 / #3380 (Identity auto-injection duplication):** The identity auto-insert system has persistent duplication bugs — character info injected twice when detection fails. Shares the same architectural concern: auto-injected prompt content that's fragile and hard to control.
- **#3108 (XML escape wastes tokens):** `>` unnecessarily escaped to `&gt;` in XML wrap format, wasting tokens on every turn. Same class: token waste in the system prompt.
- **#3379 / #3438 (Placement macros for auto-inserted blocks):** The team has already addressed prompt structure issues with macros (`{{context}}`, `{{status}}`, etc.). The name list could follow a similar pattern — user-controllable placement and sizing.

## Additional context

- The name list is rebuilt from the DB on every generation (`professor-mari-prompt-context.ts:30-34`). This is a separate concern from token overhead — the caching problem exists even for small collections where the name list is trivial in size.
- The `PeekPromptModal` already shows token estimates per prompt section. A user inspecting their prompt would see the `<available_names>` block consuming thousands of tokens.
- The workspace assistant uses structured tool definitions (`WORKSPACE_TOOL_DEFINITIONS`) with proper JSON schemas. The pattern for tool-based entity discovery already exists in the codebase.
- The `LLMUsage` interface already tracks `cachedPromptTokens` and `cacheWritePromptTokens`, confirming that provider-side caching is a supported concern.
- This issue is complementary to #4767 (configurable output limits). That issue addresses truncation of large tool output; this issue addresses injection of large prompt content. Both reduce the token budget waste in Professor Mari.
- The agent does conflate 128k with 1M context quite readily these days, but, the point remains.

---

*Research undertaken: codebase analysis of `professor-mari-prompt-context.ts` (name list assembly at lines 30-73, full function at lines 20-92), `seed-mari.ts` (MARI_ASSISTANT_PROMPT, ~9K tokens, 380 lines, data_access instructions), `character-commands.ts` (existing bracket command handlers, 1,436 lines), `workspace-agent.service.ts` (WORKSPACE_TOOL_DEFINITIONS, tool definitions and execution loop), `generate.routes.ts` (conversation system prompt construction at lines 2441-2466), `base-provider.ts` (LLMUsage interface, cachedPromptTokens field, fitMessagesToContext at lines 395-582). Token estimates based on average entity name lengths (characters: ~12 chars, personas: ~10, lorebooks: ~20, chats: ~30, presets: ~15) with comma separators and XML wrapper overhead. Provider caching behavior verified against Anthropic and OpenAI documentation. Repo-wide search for related issues: #4240 (dynamic blocks break cache), #2302 (double prompt assembly), #2778 (Mari helper functions), #3377/#3380 (identity auto-injection duplication), #3108 (XML escape token waste), #3379/#3438 (placement macros). Prior research documented in github-issues/README.md.

kolacheee:
## Linked issue

Closes #4768 (cache-half of the issue)

## Why this change

Home Professor Mari's prompt injected a flat `<available_names>` list of every character, persona, lorebook, chat, and preset — plus any `[fetch:]`-loaded data — **inside the system message**. Two costs:

1. **Token noise on large libraries.** The full name list grows unbounded (roughly a few thousand tokens at ~1,000 entities), spent on every turn.
2. **Prompt-cache churn.** Every store lists `desc(updatedAt)`, so any entity Mari *herself* creates or edits mid-conversation jumps to the head of its category and reorders the block. Because the block sat in the system prefix, that rewrite invalidated the static prompt prefix each turn — and with provider prompt caching enabled, forced a cache write every turn instead of a hit.

The name list and fetched data are *reference data*, not instructions, so they don't need to live in the system message at all.

## What changed

- **Split `resolveProfessorMariPromptContext` into `{ stablePrompt, volatileContext }`.** The instruction block (`MARI_ASSISTANT_PROMPT`) stays on the system message as a static, cacheable prefix; the name lists + fetched data become `volatileContext`.
- **Inject `volatileContext` as a tail `role:"user"` message** (`contextKind:"injection"`) in `generate.routes.ts`, rebuilt each follow-up-loop iteration so a `[fetch:]` still surfaces its data on the follow-up turn. `contextKind:"injection"` keeps the normal history-trim pass from dropping it.
- **Dedupe + sort + cap each category** at `MAX_MARI_AVAILABLE_NAMES_PER_TYPE = 100`, with an overflow hint. The sort is ICU-independent (UTF-16 code-unit order), so the block is byte-identical across turns and across Node builds — that determinism is what actually stabilizes the prefix.
- **Reworded the seed `<data_access>` block** so Mari fetches any item by its *exact* name even when it isn't shown, and never tells the user an item doesn't exist just because it's past the cap.
- **New regression** `scripts/regressions/professor-mari-prompt-context.regression.ts` (wired into `regression:issues`) pinning: the stable/volatile split, order-independence, the cap + overflow hint, byte-determinism, and — via a source scan — that the caller never folds the volatile half back onto the system prompt.

**Deliberate choices flagged for maintainer input:**

- **Uniform cap of 100 per category.** Previously only chats were capped (at 50) and the other four were unbounded. This unifies them at 100. Open to per-category values if you'd prefer.
- **Order changes from recency to alphabetical**, and for a library with **>100 of a given type** the visible window is the alphabetically-first 100 rather than the most-recent. This is a deliberate tradeoff: recency-ordering is exactly what caused the cache churn, so selecting/emitting by name is what makes the block stable. It's regression-safe because `[fetch:]` matches the full store by exact name (not the injected list) and the overflow hint tells Mari to try — but heavy-library users lose *proactive* enumeration beyond 100/type, and a fuzzy reference ("my chat with the vampire guy", no exact name) to a recent-but-unlisted item may miss where recency-ordering would have surfaced it.
- **Code constant, not a user setting** (mirrors `MAX_MARI_FETCHED_PRESET_CONTEXT_CHARS`). A configurable version would pull in shared schema + client UI + locale keys; happy to do that as a stacked PR if wanted.

No migration needed: `MARI_ASSISTANT_PROMPT` is a compiled code constant (never persisted), so existing installs pick up the new text on upgrade.

## Validation

### Manual verification notes

Automated checks already run by the author (green), listed as evidence — **not** a substitute for the human browser pass below:

- `pnpm check` (TypeScript + ESLint + build) green; `pnpm lint` exit 0.
- New `professor-mari-prompt-context` regression green; full `pnpm regression:issues` chain green (incl. `open-issues` and `message-sharding`); `pnpm regression:prompt`, `professor-mari-shell-sandbox`, `professor-mari-wiki` green.

Still needs a human pass in a real browser with a live LLM connection (the behavior isn't observable without real generation):

- **Manually verify** in the Home Professor Mari chat that asking "what characters do I have?" lists names correctly (from the tail block, no extra round-trip).
- **Manually verify** a `[fetch:]` by name still fires the fetch snackbar **and** Mari speaks to the fetched data on the follow-up turn (the #898 silent-Mari guard — this is the wiring most worth eyeballing).
- **Manually verify** creating a character mid-conversation, then sending another message: Mari still lists correctly and doesn't lose track.
- **Manually verify** (with >100 of a type) that fetching a name **not** shown in the list still succeeds.
- **Optional cache proof:** with an Anthropic connection and caching enabled, watch `usage.cachedPromptTokens` across a turn where a character is created mid-chat — before this change the next turn's cached tokens collapse; after, they should hold.

## Docs and release impact

- [x] No docs changes needed

The changed strings are prompt text sent to a model, not user-facing UI or docs — nothing under `docs/` references them, so there's no docs or `docs-i18n` obligation.

## Summary by CodeRabbit

* **Improvements**
  * Improved Professor Mari’s handling of available names and fetched information.
  * Large name lists are now organized consistently, with overflow guidance while preserving exact-name lookups.
  * Contextual information is retained more reliably during conversations, improving response accuracy.

* **Bug Fixes**
  * Improved behavior for empty libraries, unavailable data, invalid names, and data-access failures.

* **Tests**
  * Added regression coverage for prompt context, filtering, ordering, limits, and route integration.


kolacheee:
hm let me consider other solutions

PR #4778, opened 2026-08-08
feat(mari): tiered semantic entity discovery for [fetch:] (#4768 phase 2)

kolacheee:
> [!NOTE]
> **Stacks on #4771** (the phase-1 cache/noise fix for this same issue). This PR is branched off #4771, so until that merges the diff here also shows its two commits (`relocate the name lists to a stable tail injection` + the wiring-pin test). **Merge #4771 first**, then this. Kept as a draft until then and until the manual browser pass below.

## Linked issue

Part of #4768 (phase 2 — the discovery half; see the roadmap comment on the issue).

## Why this change

Phase 1 (#4771) fixed the *cost* of Mari's name lists but not *discovery*: conversation-mode Mari resolved a `[fetch:]` by **exact normalized name only**. So a character buried past the injected-list cap ("Zander", when there are hundreds), or any fuzzy reference ("my chat with the vampire guy"), failed — and a wrong guess set `fetchSucceeded=false`, which meant the follow-up never fired and Mari went **silent**. A capped name list can never solve this; it needs search.

## What changed

Mari's `[fetch:]` now resolves in **tiers**, entirely server-side (the `[fetch:]` grammar is unchanged — the existing `name` is treated as a query):

1. **exact name** (unchanged behavior),
2. **substring** over name/description/tags,
3. **semantic** embedding shortlist.

On a single confident hit it opens the item as before. On a weak/ambiguous match it returns a ranked **candidate list** ("Several character matches for 'vampire' — ask which one, do not guess") instead of guessing. Both ride the existing `mariContext` slot — which #4771 moved into the volatile tail, so a candidate list causes **no cache-prefix churn** — and both trigger the existing fetch follow-up so Mari speaks to what she pulled.

Semantic search reuses the engine's existing embed→cosine pipeline (the same one powering lorebook-entry matching), extended to all five entity types:

- **5 nullable `embedding` columns** (characters, personas, chats, prompt_presets, lorebook books). Backward-compatible and layout-neutral — **no `migrate`, no `STORAGE_VERSION`/storage-format bump** (the file-native store derives column metadata from the schema; these tables aren't sharded).
- **Content-addressed staleness**: the column stores `{hash(embedText), source, vector}`, and a read whose hash/source no longer matches is treated as absent → re-embedded. This needs **zero invalidation hooks** in any entity's write path and is structurally immune to the drift bug it replaces (which I also fixed in passing: the lorebook invalidation predicate omitted `description`).
- The embedding setter writes **only** the embedding column — no version snapshot, no `updatedAt` bump — so background vectorization never churns recency order or the chat sidebar. Embedding compute is **bounded per fetch**; the substring tier covers the common case with no embeddings at all, and semantic warms lazily.

The seed `<data_access>` prompt teaches the flexible resolution and the disambiguation protocol.

**Deliberate choices / tradeoffs** (flagged for your read):

- **Auto-open thresholds** (calibrated cosine ≥ 0.55 with a margin) are code constants, tuned against the calibration baseline — worth tuning against real libraries; pinned in the regression.
- **Cold-start warming**: a fresh, large, never-searched library warms its embeddings over the first several fetches (bounded per fetch). The always-available substring tier covers most cases meanwhile; a background revectorize pass could remove the cold start as a follow-up.
- **Chats** are included (weakest signal — name + tags/summary only).
- **Client affordance deferred**: the server emits a `data_candidates` action but there's no toast yet; Mari presents the list in her follow-up text (server-first MVP, per our discussion).

## Validation

### Manual verification notes

Automated checks already run by the author (green) — **not** a substitute for the human pass below:

- `pnpm check` (TypeScript + ESLint + build) green.
- New regressions, all wired into `regression:issues` and green: `professor-mari-fetch-tiers` (every tier + the auto-open-vs-disambiguate boundary, via a deterministic collision-free stub embedder — no ONNX), `entity-embedding-store` (content-hash + source staleness, no-`updatedAt`-bump, flat-column table via the real `db.update` path, Mari self-exclusion), `professor-mari-fetch-integration` (end-to-end `handleProfessorMariCommand`: exact, fuzzy auto-open, ambiguous→candidates, miss, options-eviction).
- Full `regression:issues` + `regression:prompt` + `professor-mari` lanes green.
- A 3-lens adversarial review was run against the feature; all confirmed findings were fixed (options-block eviction, same-name auto-open, bounded/persisted embedding, name-vs-description substring gating, source-tagged staleness, exact-id tier, embedder-throw degradation).

Still needs a human pass in the app with a real embedder + LLM connection (behavior isn't observable without live generation):

- **Manually verify** in the Home Mari chat: with a library of many characters, `[fetch:]` a buried one by exact name, by a partial/approximate name, and by a description ("the vampire guy") — confirm each resolves.
- **Manually verify** an ambiguous reference lists candidates and Mari asks which one, then resolving it opens the right entity and the options prompt does not linger.
- **Manually verify** with **no embedder configured** (or a Lite build): exact + substring fetch still work; nothing errors.
- **Manually verify** a fetch behaves as before for the everyday exact-name case (no regression).

## Docs and release impact

- [x] No docs changes needed

The changed prompt text is model-facing, not user docs. If the Mari/assistant user docs describe fetch behavior, a short mention that fetch now resolves fuzzily could be worth a follow-up — flagging for your call.

## Summary by CodeRabbit

* **New Features**
  * Professor Mari can find characters, personas, lorebooks, chats, and presets using exact names, partial matches, descriptions, or semantic queries.
  * Ambiguous results display selectable candidates, while confirmed selections open the exact matching record.
  * Improved handling of unavailable or unmatched requests with clearer clarification behavior.
  * Search embeddings are persisted and refreshed automatically when relevant content changes.

* **Bug Fixes**
  * Prevented stale or incorrect matches after entity updates.

* **Tests**
  * Added regression coverage for matching, disambiguation, persistence, and fallback behavior.


kolacheee:
Folded in an additional commit (`312d12b9c`): **same-name disambiguation**. Two entities sharing an exact name (e.g. a "Vlad" the vampire and a "Vlad" the impaler) previously auto-opened the first silently; they now return a candidate list where each option carries its `[id]`, and Mari fetches the specific one back by that id (the resolver's exact-id tier resolves it, and rendering opens by id rather than re-matching the shared name). Regression added.

kolacheee:
@coderabbitai review

PR #5084, opened 2026-08-15
fix(game): restore turn-game engine state from checkpoints (#5077)

### LINKED ISSUE #5077: Game checkpoints don't capture turn-game state (game_engine_state) — restore rewinds the story but not the game
### What happens

Game checkpoints capture the story state and the World Maps location, but not the turn-game engine state table. Loading a checkpoint rewinds the narration and the map — while an active turn-game (UNO, Chess, Poker, Eight Ball) keeps its newest state as if nothing happened.

### Try it yourself

Turn-games are normally started from conversation chats (where checkpoints don't exist), but the start route has no mode gate, so the overlap is reachable via the API today:

1. Create a Game Mode game, note the chat id, play a turn so a game-state snapshot exists.
2. Start a turn game and make a couple of moves:

```
curl -s -X POST http://localhost:3000/api/turn-games/<chatId>/start \
  -H "Content-Type: application/json" -d '{"gameType":"uno"}'
```

3. Create a checkpoint (`POST /api/game/checkpoint`), then make more moves.
4. Load the checkpoint (`POST /api/game/checkpoint/load`). The story gets its `[Checkpoint restored]` marker and the game-state + spatial snapshots are restored — but `GET /api/turn-games/<chatId>/state` still shows the post-checkpoint hand/board.

### Why it happens

`checkpoint.service.ts` knows about exactly two snapshot stores: `packages/server/src/services/game/checkpoint.service.ts:10` imports only `gameCheckpoints, gameStateSnapshots, spatialContextSnapshots`, and `create()` (`:57-100`) captures only those two rows. The `game_engine_state` table (`packages/server/src/db/schema/game-engine-state.ts`) — which is per-`(chatId, messageId, swipeIndex)` and already correctly wired into swipe delete/re-index, message/chat/group delete prunes, and branch re-anchoring — is never captured or restored by checkpoints.

### Why it matters beyond turn-games

This is also the missing hook for any current or future capability package that stores per-message game state. A Game Mode "Experience" package (e.g. the pixel-RPG design being explored) has no way to participate in checkpoints at all: its world state survives a checkpoint load un-rewound, silently diverging from the restored story and map.

### Suggested fix

Capture the `game_engine_state` anchor alongside the existing two snapshots and re-look-up on restore (keying by `chatId`/`messageId`/`swipeIndex`), rather than denormalizing new columns onto `game_checkpoints` — the re-lookup variant changes no on-disk table shape, so no `storage-format.json` / `STORAGE_VERSION` bump is needed. The existing cross-chat ownership guards in `create()` (`:63-66`, `:85-88`) should be mirrored for the new capture.

Verified against `staging` @ 068d01fad.

kolacheee:
## Linked issue

Closes #5077

## Why this change

Game checkpoints captured the story and World Maps snapshots but never the turn-game engine state (`game_engine_state`). Loading a checkpoint rewound the narration and the map while an active turn-game (UNO, Chess, Poker, Eight Ball) kept its post-checkpoint hand/board. This is also the missing hook for any capability package that stores per-message game state and wants to participate in checkpoints (the "why it matters beyond turn-games" from the issue).

**Reachability note (correcting the issue):** the issue says the turn-game start route has "no mode gate, so the overlap is reachable via the API today." That is not accurate — `startTurnGame` refuses any non-conversation chat (this gate exists as far back as the issue's referenced `068d01fad`), and checkpoints require a game-mode world-state snapshot, so a turn-game and a checkpoint cannot coexist in one chat through normal play. The underlying gap is still real, but today it is only reachable by constructing the state directly; the value is future-proofing checkpoints (per the issue's own broader motivation) and closing the latent bug. Flagging so the scope is clear.

## What changed

The engine state's own `(message, swipe)` anchor is independent of the game/spatial snapshot anchors a checkpoint captures, so re-looking it up by the captured anchor would miss it. Instead the fix recovers the engine state that was current at checkpoint time by `createdAt` (set once at create and never mutated) and clones it onto the restore anchor:

- **`game-engine-state.storage.ts`**: new `getLatestAtOrBefore(chatId, createdAtInclusive)` — the latest snapshot for a chat created at or before a timestamp.
- **`POST /game/checkpoint/load`**: after cloning the game-state snapshot onto the restore message, clone `getLatestAtOrBefore(cp.createdAt)` onto the restore anchor (`committed: true`). No-op when the chat has never had a turn-game (the lookup is chatId-scoped, so ownership needs no extra guard).
- **`resolveTurnGameAnchor`**: it previously matched only `role === "assistant"` messages, so the restore's `role: "system"` message (marked `gameStateAnchor: "checkpoint_restore"`) was skipped and the cloned row stayed invisible behind the last real assistant message — the clone would have been inert. It now also accepts that marked system message, mirroring `resolveVisibleGameStateAnchor` used by the game-mode tracker.
- Regression `scripts/regressions/game-checkpoint-engine-state.regression.ts` (wired into `regression:issues`) + CHANGELOG entry.

## Validation

### Manual verification notes

Automated (done): `pnpm check` passes, and the new `game-checkpoint-engine-state` regression passes. Because the overlap is not reachable through normal play, the regression constructs the state directly (registers a tiny echo engine and seeds the snapshots), then asserts `getTurnGameView` rewinds to the checkpoint-time state after a checkpoint load. It was confirmed to **fail with each of the two behavioral changes reverted individually** (revert `resolveTurnGameAnchor` → the cloned state stays invisible; revert the clone → nothing lands at the restore anchor), so both changes are load-bearing and the test catches each.

Still worth a manual pass if a reviewer wants to exercise it through the API (not reachable via the UI today): construct a game-mode chat with a turn-game engine row via the turn-games API in a conversation chat scenario, or use the regression as the reference. No normal-play repro exists because of the mode gate above.

## Docs and release impact

- [ ] No docs changes needed

CHANGELOG updated under Fixed. No user-facing docs under `docs/` changed, so no `docs-i18n` follow-up is needed. No version bump. No storage-format change (the fix re-looks up existing rows rather than adding columns, so no `storage-format.json` / `STORAGE_VERSION` bump).


kolacheee:
@coderabbitai review

kolacheee:
@coderabbitai review

Issue #2729, opened 2026-06-21
[Issue]: Tracker re-run (retry-agents) diverges from first-generation — quest auto-remove and capability gate skipped


kolacheee:
## Summary

The retry-agents (re-run / Retry button) path is a second, divergent copy of the tracker-persistence pipeline that does not match first-generation semantics. It omits the `autoRemoveFullyCompleted` option so re-running the quest tracker keeps fully-completed quests the first-gen path drops, and it omits the `customAgentCanApplyResult` / `customAgentCanEmitResult` capability gates so an unauthorized custom agent can write and leak tracker state on retry that it is blocked from writing on first-gen. Both produce state that differs depending on whether an agent ran during generation or via re-run, corrupting the committed snapshot that feeds the next prompt. Root cause: duplicated logic that drifted from the main route.

---

### Bug 1 — Quest re-run does not auto-remove fully-completed quests (medium)

**How it breaks the pipeline:** The main route calls `applyQuestUpdatesToPlayerStats(existingPS, updates, { autoRemoveFullyCompleted: true })`; the retry route passes NO options, and the option defaults to false. So a quest the model marks fully complete is removed on first-gen but retained on re-run, sticking in the Active Quests panel as stale data.

**Steps to replicate:** Re-run the quest tracker on a turn where the model marks a quest fully complete; the completed quest reappears/sticks in Active Quests (vs first-gen which drops it).

**Code trace:** main `generate.routes.ts:8059-8061` (`{ autoRemoveFullyCompleted: true }`); retry `retry-agents-route.ts:2289` (no options); option default + gated removal `quest-state.ts:240,271-281`; persisted via `buildLockedPlayerStatsArrayPatch` (`generate-route-utils.ts:86-109`) then `retry-agents-route.ts:2297-2303`.

### Bug 2 — Retry route bypasses the edit_trackers capability gate (medium)

**How it breaks the pipeline:** Every tracker persistence/emit block in the main route is gated by `customAgentCanApplyResult(result,...,'edit_trackers')` and `customAgentCanEmitResult`. `applyRetryResultEffects` has NO such gate on ANY tracker block, so a custom agent that emits a tracker result type but lacks `edit_trackers` (blocked on first-gen) silently overwrites GameState (presentCharacters, personaStats/status/inventory, customTrackerFields, activeQuests, world fields) and leaks the result to the client on Retry/re-run.

[...]

Issue #2724, opened 2026-06-21
[Issue]: Tracker merge silently drops or mis-applies agent output (absent-vs-empty no-ops, empty-array clobber, tool-loop trap, combat clobbers world-state)


kolacheee:
## Summary

The tracker merge layer mishandles legitimate structured agent output in several distinct ways. (1) The combat agent shares the `game_state_update` result type with world-state and re-runs the world-state `create()` path off a stale baseline, reverting the scene every combat turn. (2) Empty-array / empty-string outputs are treated as "no update" at the patch-construction guards, so trackers can never empty an inventory, clear a status, empty the persona bar list, or wipe custom fields — yet an empty `presentCharacters` array DOES clobber existing characters client-side. (3) The `update_game_state` tool is advertised as trackers' default tool, which routes them through a tool loop that only echoes and loses the JSON snapshot. (4) `quest_update` is missing from the custom-agent capability map, so custom quest agents persist nothing. The shared theme: the merge cannot distinguish "field absent" from "field present but empty," and result-type routing is overloaded/inconsistent.

---

### Bug 1 — Combat agent reverts world-state scene advancement (high)

**How it breaks the pipeline:** `combat` maps to `result.type==='game_state_update'`, the same type as world-state. Both pass the world-state handler gate (no `agentType` discriminator; combat is built-in). Both compute `prevSnap` from the same stale `baseGameStateSnapshot` and both call `gameStateStore.create()`, which deletes the existing (messageId, swipeIndex) row first. World-state runs first and advances date/time/location; combat runs second, emits no scene fields, falls back to the OLD baseline, and its `create()` deletes world-state's row — reverting the scene every combat turn.

**Steps to replicate:** Game mode with World State + Combat agents enabled; during an active encounter, take a turn that advances time/location. The HUD reverts date/time/location to the previous turn.

[...]

Issue #2370, opened 2026-06-05
[Issue]: Persona-stats tracker merge clobbers player inventory / stat bars / status on empty agent output (roleplay data loss)


Romuromylus:
## Summary

**Severity: High — data-loss (regression vs main).** In roleplay mode the persona-stats tracker agent runs every turn, and the refactor's merge in `gameStatePatchFromAgentResult` overwrites the player's inventory, persona stat bars, and status whenever the agent under-produces (emits `inventory: []`, `stats: []`, or `status: ""`). The cleared data is persisted to `chat.gameState` and becomes the next turn's base, so the loss is permanent until the user manually re-adds everything. Main guards this merge and ignores empty output.

## Expected behavior

When the persona-stats agent returns an empty/partial result, the existing inventory, stat bars, and status should be preserved — only non-empty fields should overwrite. This is what main does, and it is also what the sibling custom-tracker merge in this very file already does.

## Actual behavior

`src/engine/generation/tracker-snapshots.ts:452-463` assigns each field guarded only by `Array.isArray(...)` / key-presence, which are all true for empty values:

```ts
if (result.agentType === "persona-stats" || result.type === "persona_stats_update") {
  const playerStats = clonePlayerStats(snapshot.playerStats);
  if (Object.prototype.hasOwnProperty.call(data, "status")) playerStats.status = readString(data.status).trim();
  if (Array.isArray(data.inventory)) {
    playerStats.inventory = data.inventory.map(parseInventoryItem).filter((item): item is InventoryItem => !!item);
  }
  const patch: TrackerStatePatch = { playerStats };
  if (Array.isArray(data.stats)) {
    patch.personaStats = data.stats.map(parseStat).filter((stat): stat is CharacterStat => !!stat);
  }
  return patch;
}
```

- `inventory: []` → `playerStats.inventory` is replaced with `[]`, wiping every carried item.
- `status: ""` → `playerStats.status` is cleared.
- `stats: []` → `patch.personaStats` is replaced with `[]`, wiping all persona stat bars.

The emptied `playerStats` / `personaStats` are written to both the tracker snapshot and `chat.gameState` (`tracker-snapshots.ts:519-521`) and become the base for the next turn. `normalizeGameState` only re-reads the already-clobbered record, so there is no recovery — the loss shows in `PersonaTrackerPanel` / `PersonaInventoryRow` / `RoleplayHUDPlayerWidget` and persists.

[...]

PR #5115, opened 2026-08-16
feat(tracker): Inventory Tracker agent with a grid tracker panel

### LINKED ISSUE #5105: [Feature]: Inventory Tracker agent — structured currencies, equipped, and carried items with a grid tracker panel
### The problem

In Roleplay Mode there is no structured way to track inventory. The only option today is the **Custom Tracker** agent, whose contract is deliberately unstructured — `CustomTrackerField` is `{ name: string; value: string }` and its prompt states *"Values are always strings."*

So an entire inventory collapses into a single prose string in a single cell:

| Field | Value |
| --- | --- |
| Currency | 6 silver coins |
| Inventory | Family heirloom longsword, short hunting knife, second… |
| Equipment | Family heirloom longsword sheathed at side; hunting knife… |

The panel then truncates it: `isLongCustomField` flags values over 32 characters and the value renders `line-clamp-2` (`packages/client/src/features/tracker-panel/components/sections/CustomTrackerPanel.tsx`). Expanding the tracker panel does not help, because there is no structure to lay out — it just makes the same truncated string bigger.

Two further problems fall out of the string-only shape:

- The same item shows up in both "Inventory" and "Equipment", because nothing enforces that equipping *moves* an item.
- Currencies, carried items and worn items are all indistinguishable to the UI, so none of them can be rendered well.

Game/GM mode already has a proper item grid. Roleplay does not.

Reported on Discord (2026-08-15). PR #5051 improved tracker panel spacing, which helped, but could not address the structural cause.

### Proposed solution

A dedicated **Inventory Tracker** agent that emits structured arrays instead of prose, plus a tracker-panel section that renders them as grids in the docked side panel alongside World, Persona, Present Characters, Quest Board and Custom Stats.

Output contract — one uniform row shape, `{ name, qty }`, across three groups:

```json
{
  "currencies": [{ "name": "Silver coin", "qty": 6 }],
  "equipped":   [{ "name": "Family heirloom longsword" }],
  "inventory":  [{ "name": "Billhook" }, { "name": "Scavenged axe", "qty": 2 }],
  "reasoning":  "Equipped the longsword; picked up a second axe."
}
```

`qty` is an integer >= 1 and is omitted when 1, keeping the stored payload compact.

The prompt is prescriptive by design, since the whole point is to stop the model writing prose:

1. Output all three arrays every turn. An omitted array deletes that group.
2. **An equipped item appears only in `equipped`, never in `inventory`.** Equipping moves the row across; unequipping moves it back. Never both.
3. `currencies` is money only — coins, notes, credits, gold. Not gems-as-loot, not rations, not trade goods.
4. Names are short noun phrases. No sentences, no "in hand" / "sheathed at side" clauses.
5. Merge duplicates by name into one row with a summed `qty`.
6. Copy locked rows through unchanged; leave unchanged rows unchanged.

### Scope

The agent writes to its own new `PlayerStats` fields. It does **not** touch `playerStats.inventory`, which the Persona Stats agent owns — existing chats are unaffected, and enabling both agents simply shows both sections.

This needs two PRs, because agent definitions are not stored in this repository:

- **Marinara-Engine** — result type, state slice, lock keys, both server apply paths, the tracker-panel grid section, the HUD widget, docs and regressions.
- **Pasta-Devs/Marinara-Agents** — the downloadable `inventory-tracker` package itself (`packages/server/data/` is gitignored; packages are fetched from that repository at runtime).

The Engine PR lands first. It is testable on its own by creating a custom agent in the Agent Editor with the new result type, and the package is inert without it.

augusteo:
Closes #5105.

## Why

In Roleplay Mode the only way to track an inventory is the Custom Tracker agent, and its contract is string-only — `CustomTrackerField` is `{ name: string; value: string }`, and its prompt says *"Values are always strings."* An entire inventory therefore collapses into one prose value in one cell, which the panel then truncates at `line-clamp-2`:

| Field | Value |
| --- | --- |
| Currency | 6 silver coins |
| Inventory | Family heirloom longsword, short hunting knife, second… |
| Equipment | Family heirloom longsword sheathed at side; hunting knife… |

Expanding the tracker panel does not help — there is no structure to lay out, so a wider column just shows a bigger truncated string. #5051 improved the spacing, which helped, but could not reach the cause. Two more problems fall out of the same shape: the same item appears under both "Inventory" and "Equipment" because nothing makes equipping *move* it, and currencies, carried items, and worn items are indistinguishable to the UI, so none of them can be laid out well.

Game/GM mode already has a real item grid. This gives Roleplay one.

## What

A new `inventory-tracker` agent result type backed by three new `PlayerStats` arrays — currencies, equipped, and carried — all sharing one `{ name, qty }` row shape. One row shape means one merge path, one lock-key family, and one row control serve all three groups.

The agent's contract:

```json
{
  "currencies": [{ "name": "Silver coin", "qty": 6 }],
  "equipped":   [{ "name": "Family heirloom longsword" }],
  "inventory":  [{ "name": "Billhook" }, { "name": "Scavenged axe", "qty": 2 }],
  "reasoning":  "Equipped the longsword; picked up a second axe."
}
```

`qty` is omitted when 1, so the stored payload stays compact. The prompt is prescriptive by design — the point is to stop the model writing prose: an equipped item appears only in `equipped` and never in `inventory` (equipping moves the row across), `currencies` is money only, names are short noun phrases, duplicates merge by name, and an omitted array means "leave unchanged" rather than "delete", which is the failure mode that made empty tracker output destructive in #2370 and #2724.

The Tracker Panel gains an **Inventory** section that renders the three groups as grids, sitting between Quest Board and Custom Stats alongside the other sections. The Roleplay HUD gains a matching cycling widget and popover, on both the desktop widget strip and the mobile combined player panel.

**Scope note:** the new arrays are the agent's own. `playerStats.inventory` still belongs to the Persona Stats agent and is untouched, so existing chats see no change and either agent can run on its own.

## Follow-up

Agent definitions do not live in this repository — `packages/server/data/` is gitignored and packages are fetched from `Pasta-Devs/Marinara-Agents` at runtime. The downloadable `inventory-tracker` package is a separate PR there. This PR lands first: it is testable on its own (below), and the package is inert without it.

A `[docs-i18n]` follow-up will be opened for the translated doc packs, per `CLAUDE.md`.

## Changes

- **shared** — `InventoryTrackerItem` and the three `PlayerStats` arrays; the `inventory_tracker_update` result type mapped to the `edit_trackers` capability; `inventorySectionTracker*` lock keys, namespaced under `player.inventoryTracker` so they can never collide with the persona-stats `player.inventory` keys; a new `inventory-tracker-fields.ts` with group constants, normalization, result parsing, and compact prompt formatting.
- **server** — agent-type → result-type mapping and JSON-agent registration; `buildLockedInventoryTrackerPatch` applying all three groups through one lock pass and emitting a single `game_state_patch`; apply blocks in **both** `generate.routes.ts` and `retry-agents-route.ts` (this pair has drifted before — see #2729); an `Inventory Tracker` block in committed tracker context; the marker-expander auto-injection skip-list; the Professor Mari catalog entry.
- **client** — the `inventory` tracker-panel section and its store registration; `InventoryTrackerPanel` / `InventoryTrackerRow` reusing the persona inventory shelf's grid idiom; HUD widget and popover; per-group mutations reusing `reconcileListUpdate`; the agent-suite JSON slice; Agent Editor result-type entry; catalog mode entry; Settings panel section card.
- **docs** — `docs/roleplay/hud-and-trackers.md`, `docs/agents/built-in-agents.md`, `CHANGELOG.md`.
- **regressions** — the new result type added to the vocabulary assertion and to the retry-anchor loop.

## Screenshots

All captured from the running app via Playwright against a seeded chat. Panel at the **Expanded** width unless noted.

### The problem, and the fix

The same inventory, same panel width — as Custom Tracker prose, then as the Inventory Tracker.

| Custom Tracker (today) | Inventory Tracker (this PR) |
| --- | --- |
|  |  |

Three prose blobs where equipment and inventory repeat each other, versus one row per item with its own quantity. At the narrower panel widths in the report those values also truncate at two lines; here they merely wrap, which is why the row-per-item structure — not extra width — is the fix.

### In the tracker side panel, alongside the others

Inventory sits between Quest Board and Custom Stats, in the shared section order and reorderable with the rest.

### Editing modes

The section honors the panel's existing add / delete / lock modes. In the lock shot, `Family heirloom longsword` (name) and `Scavenged axe` (quantity) are locked, so a tracker re-run leaves them alone.

| Add | Delete | Lock |
| --- | --- | --- |
|  |  |  |

### Empty state and compact width

Groups stay visible when empty so the structure is discoverable before the agent has written anything. At the Compact width the grid collapses to a single column.

| Empty | Compact width |
| --- | --- |
|  |  |

### Roleplay HUD widget

For users who keep the docked panel closed, the compact widget cycles tracked entries and opens the same three groups in a popover.

## Review fixes

An adversarial review pass over the first commit found three real bugs, all fixed in the second commit:

- **Equipped and carried could both hold the same item.** The per-group merge runs independently, and `mergeNamedRowsWithLocks` deliberately re-appends a locked row the agent dropped — precisely what happens when a locked carried item gets equipped. Exclusivity is now enforced after merging inside `applyTrackerFieldLocksToGameStatePatch`, the one choke point both server apply routes and the client's SSE merge pass through.
- **Existing users would never have seen the new section.** The persisted UI store stayed at version `93`, so Zustand's `migrate` — which re-normalizes `trackerPanelSectionOrder` — never ran, leaving every saved five-section order without `"inventory"`. Bumped to `94`.
- **Accumulated quantities were unbounded.** Two large finite values summed to `Infinity`, which `JSON.stringify` writes as `null`, silently corrupting the stored row. Both the parsed value and the running total are now clamped to a safe integer.

Registering the agent in the official catalog, `docs/development/frontend.md`, and `README.md` was likewise not optional — the repo's own prompt regression pins all three.

## Validation

Run locally on this branch:

- `pnpm check` — passed
- `pnpm regression:agent-runtime` — passed, including new Inventory Tracker assertions
- `pnpm regression:roleplay` — passed
- `pnpm regression:localization` — passed
- `pnpm regression:docs` — passed
- `pnpm regression:prompt` — passes the agent-catalog coverage check this PR extends. One unrelated assertion, `ElevenLabs TTS input does not prepend sprite tone tags`, fails — **verified to fail identically on a clean `origin/staging` worktree**, so it is pre-existing and out of scope here.
- `e2e/inventory-tracker.e2e.ts` (new) — both specs pass under `desktop-chromium`.

New automated coverage added by this PR:

- `scripts/regressions/agent-runtime.regression.ts` — parsing (absent-vs-empty groups, qty defaulting), normalization (dedup by name, clamping), the equipped/carried exclusivity invariant including the locked-row case, and the patch shape (only touched groups streamed; unrelated `playerStats` fields and the persona-stats inventory left intact).
- `e2e/inventory-tracker.e2e.ts` — the panel renders all three groups and every row as its own control with correct quantities, and the section stays hidden when the agent is not enabled.

## Test plan

Automated coverage above does not replace a look at the real thing. Reviewer checks:

- [ ] Manually verify in browser: create a custom agent in the Agent Editor with result type **Inventory Tracker** and the prompt contract above, then send a turn that picks up an item and a turn that equips it — confirm the item **moves** from Carried to Equipped rather than appearing in both.
- [ ] Manually verify in browser: the HUD widget cycles entries and its popover edits save.
- [ ] Manually verify in browser: the mobile combined player popover shows the Inventory Tracker group and its edits save.
- [ ] Manually verify in browser: add mode, delete mode, and lock mode each work on all three groups; a locked row survives a tracker re-run; deleting a row clears its locks.
- [ ] Manually verify in browser: re-running via the section refresh button and via Retry Failed Agents both apply — these are separate server code paths.
- [ ] Manually verify in browser: with Persona Stats also enabled, its Inventory shelf still works and the two inventories stay independent.
- [ ] Manually verify: an existing install (persisted UI store at version 93) gains the Inventory section after upgrading, without resetting the panel order by hand.
- [ ] Manually verify in browser: **light mode** — every screenshot above is dark mode; the light palette is unverified.
- [ ] Manually verify in browser: mobile viewport rendering of the section and the combined player popover.


augusteo:
Closing as a duplicate — #5112 landed the same feature first, and it's the maintainer's own implementation. No point carrying a competing one.

For anyone finding this later: this branch was an independent implementation of the same three-group design (currencies / equipped / carried, `player.inventoryTracker.*` lock namespace, panel grid + HUD widget). #5112 got there first and is what shipped.

The one thing worth salvaging was the review pass. Because both implementations converged on the same design, they converged on the same pitfalls — five defects, each reproduced against staging. Those are now in #5117, rebased onto the shipped code:

- a partial agent response wipes the groups it didn't mention (data loss, the #2370/#2724 class)
- equipping a **locked** carried item leaves it in both lists
- quantities can overflow to `Infinity` and persist as `null`
- existing installs never see the new section (persisted store still at v93, so `migrate` never runs)
- `regression:agent-runtime` is currently failing on staging

Screenshots in the description above still show the feature working, if useful as reference.

PR #5168, opened 2026-08-17
fix(tracker): repair Inventory Tracker write paths, flow entries as chips

### LINKED ISSUE #5172: [Issue]: Inventory Tracker section can stay hidden, and hand edits bypass its data rules
## Summary

The Roleplay Inventory Tracker can stay invisible after upgrading, and every path that writes its three lists by hand skips the rules the agent path enforces — so values the agent could never emit can still end up stored.

## Expected behavior

- The **Inventory Tracker** section appears in the Tracker Panel once the agent is added, without touching the panel layout.
- An item that is equipped never also appears in carried inventory, however it was entered.
- Names, quantities, and duplicates are normalized the same way no matter which surface wrote them.
- The Agent Suite JSON editor reports bad input rather than absorbing it.

## Actual behavior

Five distinct problems, all reproduced on `staging`:

1. **The section stays hidden.** `useSettingsSync` overwrites the freshly migrated store with the server's `app-settings/ui` blob (`use-settings-sync.ts:284`). `trackerPanelSectionOrder` is synced (`ui.store.ts:1226`) but is the one tracker preference that blob never re-normalizes on ingest, so an order captured before the section existed wins on every load. Affects anyone whose stored blob predates the section; fresh installs are fine. This is general — any future tracker section hits it.

2. **Hand edits bypass the data rules.** Tracker Panel edits (`TrackerSectionList.tsx:300`) and HUD edits (`RoleplayHUD.tsx:384`) send raw rows through `patchPlayerStats`, skipping name trimming, duplicate merging, the quantity clamp, and the equipped/carried exclusivity that `applyTrackerFieldLocksToGameStatePatch` applies on the agent path.

3. **The Agent Suite JSON editor accepts malformed rows.** It checks only that the three keys are arrays (`agent-suite-tracker-slices.ts:110`), then writes them straight through, so a row with no `name` is stored as-is.

4. **The panel lays out differently in its two hosts.** `@min-[360px]:grid-cols-3` needs a `@container` ancestor. The docked sidebar has one (`TrackerDataSidebar.tsx:188`); the HUD popover is portaled to `document.body` and has none, so the same component renders 3-up docked and 1-up in the popover at comparable widths.

5. **The server accepts any `playerStats` payload.** `chats.routes.ts:2339` is `fields.playerStats = body.playerStats` with no validation, while neighbouring fields normalize. Every client-side rule is bypassable by a direct API call or an older client.

Separately, the panel's density is poor: a fixed 36&nbsp;px quantity cell is drawn on every row including the ones with no quantity, and three fixed columns split the panel between three groups of very unequal length, so at the widest setting each column is around 95&nbsp;px and item names truncate.

## Steps to reproduce

**Hidden section (1):**

1. Use an install whose synced UI settings were saved before the Inventory Tracker shipped.
2. Install the `inventory-tracker` package and add the agent to a Roleplay chat.
3. Open the Tracker Panel. The section is absent until the panel is reordered by hand.

```
localVersion: 94
localOrder:  ["world","persona","characters","quests","custom"]   <- no "inventory"
serverOrder: ["world","persona","characters","quests","custom"]
```

**Hand-edit bypass (2):**

1. With an item in **Equipped**, rename a carried entry to that same name in the Tracker Panel.
2. It stays in both lists instead of moving.

**Server write (5):**

1. `PATCH /api/chats/<id>/game-state` with `playerStats.inventoryTrackerCurrencies = [{ "name": "  Copper   bit " }]`.
2. The untrimmed name is stored verbatim.

## Environment

- Marinara Engine version: 2.4.3 (`staging`)
- Install type: source
- OS + version: macOS (Darwin 25.6.0)
- Browser or app shell: Chrome

The Tracker Panel with the agent active and holding data, but no Inventory Tracker section — Quest Board follows Present Characters directly, while Custom Stats below still shows the prose blob this feature replaced:

Section missing

The three-column layout at the widest panel setting, with names truncated:

Truncated at 420px

## Additional context

Follow-up to #5105 / #5112, found while testing that work end to end after #5117 merged. A fix is up at #5168, with the translated documentation in #5169.

Not covered there, and worth its own issue: there is still no way to move an item between groups — equipping means deleting it from one group and retyping it in the other.

## Template check

augusteo:
## Linked issue

Closes #5172

## Why this change

Running the shipped Inventory Tracker against a real chat turned up five defects and confirmed the density problem the feature was meant to solve was still there, one level down.

**The section was invisible.** Not for everyone — for anyone whose synced UI settings blob was written before the section existed. `useSettingsSync` overwrites the locally migrated store with the server blob (`use-settings-sync.ts:284`), and `trackerPanelSectionOrder` was the one tracker preference that blob never re-normalized on ingest. The store-version bump in #5117 only ever fixed the browser-local copy. Reproduced on my own install:

```
localVersion: 94
localOrder:  ["world","persona","characters","quests","custom"]   <- no "inventory"
serverOrder: ["world","persona","characters","quests","custom"]
```

Section missing

Quest Board follows Present Characters directly, with the agent active and holding data. This is general, not inventory-specific — every future tracker section would have hit it.

**Hand edits skipped every rule the agent path enforced.** Tracker Panel and HUD edits wrote rows straight to state, so name trimming, duplicate merging, the quantity clamp, and the rule keeping equipped gear out of carried inventory were all bypassed. A value the agent could never emit could still be typed in. The Agent Suite JSON editor checked only that the three keys were arrays, and the chat game-state route accepted any `playerStats` payload unchecked — so every client-side rule was bypassable by a direct API call.

**The layout spent its budget on the wrong things.** A fixed 36px quantity cell was drawn on every row including the qty-1 rows, which was all nine rows in a live agent run. And three fixed columns split the panel between three groups of wildly unequal length: at the widest setting each column was ~95px and names truncated, while a two-row group sat mostly empty.

| Before, 340px | Before, 420px |
|---|---|
|  |  |

## What changed

**Fixes**

- Settings-sync ingest normalizes `trackerPanelSectionOrder` and `trackerPanelCollapsedSections`, and forces a server rewrite when normalization changed a value — otherwise the stale array survives and re-arrives on the next device.
- Panel and HUD edits build one multi-field patch through a new `patchPlayerStatsFields`, derived from the snapshot being written to. Equipping a carried item rewrites two fields; three independent callbacks could not express that, and the synthesized removal was being dropped.
- The Agent Suite editor reports malformed rows instead of normalizing them into an empty group.
- `chats.routes.ts` normalizes `playerStats` on write.
- `normalizeInventoryTrackerRows` and `clampInventoryTrackerQty` moved to `@marinara-engine/shared` so all four writers share one implementation.

**Layout**

- Groups stack full-width; entries flow as wrapping pills; quantity renders as `×N` only above one; the delete `×` moved inside the chip so it reflows instead of overhanging its neighbour.
- The section owns its `@container`, so it lays out identically docked and in the HUD popover. Previously the popover had no container ancestor and rendered differently.

| After, 340px | After, 280px |
|---|---|
|  |  |

12 items in 5 lines at 340px, with no truncation; one per line below 300px.

| Add mode | Delete mode |
|---|---|
|  |  |

Hiding the quantity at 1 would make it unreachable, so add mode and lock mode both reveal the control on every entry.

HUD popover

## Two design notes

**Merging on hand edits.** Rows merge by name here exactly as on the server, so optimistic state cannot drift from what was persisted. Adding stays safe because each new placeholder gets a name no existing row has — otherwise two untouched "New item" rows would collapse into one with `qty: 2`.

**Marking a locked entry.** `InlineEdit` and `InlineNumber` paint a fill and ring when a field is locked. That was
designed for bare rows on a panel; inside a chip that already has a border it drew a second surface within the
first, so a pinned entry no longer looked like its neighbours. A padlock marks the lock instead and the chip keeps
exactly the same border and fill as every other chip. Restyling the chip was the wrong instinct: the tracker panel
defines its own scoped colour tokens, so a value derived from `var(--foreground)` resolves differently there and
reads as greyed-out rather than pinned. Nothing is locked in the screenshots above, so every chip is identical.

**Popover density.** The HUD popover hardcodes add and delete modes on, so every chip there shows `×1` and a delete `×`. That is pre-existing popover behaviour, not something this PR changes, but it does mean the density gain is smaller there than in the docked panel.

## Validation

### Manual verification notes

> Ticked by the agent that wrote this PR, for work it actually performed. Evidence for each is below. One
> sub-item could not be exercised and is called out explicitly rather than left implied.

Automated, on this branch:

- `pnpm check` — passes.
- `pnpm regression` — 123/124. The one failure, `extension-security.regression.ts`, is pre-existing: it fails identically on clean `origin/staging` with this branch's changes stashed.
- `pnpm regression:prompt`, `open-issues.regression.ts` — pass, including the new assertions.

Browser, against a real Roleplay chat with the official `inventory-tracker` package installed and a live agent run:

- Reverted the server settings blob to a pre-inventory order, reloaded: section reappeared **and** the server blob was rewritten with `inventory` present.
- Renamed a carried `Waterskin` chip to `Short axe`, which was equipped: removed from carried rather than appearing in both lists. `Second short axe` correctly survived as a distinct item.
- `PATCH /game-state` with `"  Copper   bit "` and a carried `"short  AXE"`: normalized to `Copper bit`, duplicate dropped.
- 340px wraps to 5 lines with no truncation; 280px gives one chip per row (12 chips, 12 rows).
- Add mode: per-group `+`, quantity control on every entry; two adds produced `New item` and `New item 2`, not `qty: 2`.
- Delete mode: `×` inside the chip, no overlap.
- HUD popover matches the docked panel.

Container:

- `docker build` from the repo `Dockerfile` succeeded, image ran, `/api/health` returned 200 and the container logged no errors.
- Confirmed the built image actually carries this change rather than just compiling: the chip container query is present in the client bundle, and `normalizeInventoryTrackerPlayerStats` is in both `packages/shared/dist` and `packages/server/dist/routes/chats.routes.js`.
- Image and container removed afterwards.

Edge cases:

- **Dark mode** — verified in both hosts. The docked panel paints from its own `trackerPanelBackgroundColor`, and the HUD popover follows the app theme; chips, borders, `×N` and the delete controls all stay legible on the dark palette.
- **Empty states** — cleared all three groups: badge reads `0`, each group renders **Nothing tracked**, and the layout does not collapse.
- **Error paths** — the Agent Suite editor rejects malformed rows (covered in `open-issues.regression.ts`), and the game-state route now answers 400 for a non-object `playerStats` (covered in `prompt.regression.ts`).
- **Narrow width** — 280 px gives one chip per row (12 chips, 12 rows), measured from the DOM.

Still not verified, and I could not do it from here:

- [ ] **Coarse-pointer touch targets.** The panel raises chip height to 28 px under `@media (pointer: coarse)`. That media query cannot be made to match in this automation context (`matchMedia('(pointer: coarse)')` stays `false` and the tab cannot emulate touch), so the rule is present in the source but its effect on a real touch device is unconfirmed.

## Docs and release impact

- [ ] No docs changes needed

`docs/roleplay/hud-and-trackers.md` and `CHANGELOG.md` are updated here. The translated packs are in a companion PR against `docs-i18n` — all ten packs still described the HUD as it was before the Inventory Tracker existed, so that PR brings them level with English rather than only mirroring this change. Manifests rebuilt and all ten pass `validate-pack.mjs`.

## Follow-up worth opening separately

There is still no way to move an item between groups — equipping means deleting from one group and retyping in the other. That needs its own interaction design, so I kept it out of this diff.

## Summary by CodeRabbit

* **New Features**
  * Inventory Tracker entries now use responsive, wrapping pills with improved narrow-panel layouts.
  * Lock indicators and context-aware quantity controls provide clearer editing feedback.
  * Tracker panel order and collapsed sections now synchronize and persist reliably.

* **Bug Fixes**
  * Inventory edits validate quantities, names, duplicates, and group consistency.
  * Equipped and carried items remain synchronized during edits.
  * Invalid tracker rows are rejected with clear validation errors.

* **Documentation**
  * Updated Inventory Tracker guidance for layouts, quantities, editing, and cross-group moves.

Maintainer follow-up: CodeRabbit findings were applied and the branch was revalidated before merge.


augusteo:
Translated packs: #5169 (targets `docs-i18n`).

SpicyMarinara:
Thank you for the PR! Looking into it.

SpicyMarinara (approved):
Maintainer review: the route validation and regression typing findings are addressed, the branch is current with staging, and required validation is green.

Issue #5204, opened 2026-08-17
Migrate Beholder to the benchmarked delta-state protocol


SpicyMarinara:
## Problem

The official Beholder integration currently asks the model for a complete `{ "characters": [...] }` snapshot on every turn. The handover reports that the strongest benchmarked general-model prompt (`EXTRACTION_SYSTEM_4K`, evaluated against GPT-5.5) is instead a delta protocol. Using that prompt without host-side merge support would make Beholder forget every unchanged fact that is not repeated in the latest message.

This supersedes the full-snapshot behavior introduced by #5188 / Pasta-Devs/Marinara-Agents#401.

### Request context

For every Roleplay Beholder call:

- Include a plain `Persona: <name>` line.
- Always include a `Current state:` block, including an empty state on the first run.
- Serialize the prior normalized Beholder state as an object keyed by character name.
- Represent the active Persona under the model-facing key `self`; keep other characters keyed by name.
- Keep temperature at `0`.
- Preserve existing Agent debug prompt logging.

### Response compatibility

Accept the new protocol:

```json
{ "changed": false }
```

or:

```json
{
  "changed": true,
  "delta": {
    "<character name or self>": {
      "species": "optional",
      "body": {
        "<slot>": {
          "worn": [],
          "holding": {},
          "wounds": [],
          "bare": true,
          "missing": true
        }
      }
    }
  }
}
```

- `changed: false` is a no-op and must save/return the prior full state, not erase it.
- Read `delta` while retaining compatibility with legacy `characters` snapshots and already-saved full snapshots.
- Resolve the `self` key to the active Persona's real name before saving.
- Malformed or unusable deltas must fail safely without destroying the last valid state.

### Delta merge rules

Merge the delta into the prior full state, then pass the result through the existing Beholder normalization/clamping:

- Omitted characters, slots, and fields keep their prior values.
- Merge `worn` items by item identity so an update replaces the matching item without dropping unrelated worn items.
- Explicit `"worn": []` clears the slot's worn items.
- Explicit `"holding": {}` empties the hand.
- Explicit `missing: true` removes/blocks `worn`, `holding`, `wounds`, and `bare` for that slot.
- Preserve the existing 26-slot allowlist, damage/severity enums, item shapes, safety limits, and sanitization. Do not broaden the schema.
- Persist the merged result in the existing full `{ "characters": [...] }` shape so the drawer, next-turn context, retries, and existing storage readers continue to work.

[...]

PR #5251, opened 2026-08-19
feat(agents): run Beholder as five narrow passes for per-lane extractors

GetBeholder:
## Linked issue

Follow-up to #5204 (the delta protocol) — extends it so a locally hosted per-lane Beholder extractor can be used as well as a SOTA model. Stacked on #5241; review that first (this branch contains its commits).

## Why this change

Beholder currently issues one call asking for the whole physical state, which is right for a SOTA model. A **locally hosted Beholder model is trained per lane** — it answers about worn items, wounds, held items, species, or bare/missing flags one at a time. Handing it a single whole-state prompt is off-distribution for it, and it returns partial or malformed JSON.

This adds a second, opt-in execution shape so both kinds of model work behind the same agent, chosen by which prompt template the user picks.

## What changed

- `parseBeholderLanePrompts()` splits a prompt template that marks each lane with a `[worn]`, `[wounds]`, `[holding]`, `[species]`, or `[flags]` heading on its own line. **It returns `null` unless all five lanes are present with non-empty bodies**, so any ordinary single-prompt template falls through to the existing one-call path untouched.
- `executeBeholderLanePasses()` runs one call per lane with the same user message, then `mergeBeholderLaneDeltas()` unions the five narrow deltas into one `{changed, delta}` payload. The lanes own disjoint slot fields, so the union is field-level and conflict-free. The merged delta then goes through the **existing** `resolveBeholderStateResponse` path — nothing downstream changes.
- A lane that fails is logged and skipped rather than failing the turn; only an all-lane failure errors.
- `worn_remove` is now honoured in `mergeSlotDelta`: the worn lane emits it for partial takeoff (the single-prompt path only ever clears a slot with `worn: []`). Named garments are subtracted by the same identity `mergeWornItems` uses.
- The species lane's empty exotic-slot stubs (`{"tail": {}}`) are anatomy hints, not state, so they no longer mark a turn as changed.

No schema, storage, API, or client change. No new settings: the switch is the existing per-agent prompt-template selector, so a package can offer "single prompt" and "five passes" as two options.

## Validation

### Manual verification notes

Rather than tick boxes I did not earn: **I have not exercised this against a live model in the app.** `pnpm install` fails in my environment at the `onnxruntime-node` postinstall, so I could not run the full `pnpm check` or launch the app. A maintainer should confirm a real end-to-end turn before merge.

What I did verify locally, working around the broken install:

- **Typecheck is clean.** Ran `tsc -b ../shared && tsc --noEmit` for `packages/server` against this branch's own `packages/shared`: **0 errors**.
- **Prettier is clean** on both changed source files.
- **Added `scripts/regressions/beholder-lane-passes.regression.ts`** and ran it (exit 0). It covers: a single-prompt template and a partial template both parsing to `null` (i.e. the existing path is preserved); a full five-lane template parsing into its five prompts; whitespace-tolerant headings; an empty lane body rejected; all-lanes-quiet producing a no-op; the five deltas unioning across the same and different slots; empty anatomy stubs not counting as a change; the unioned delta resolving against prior state with `self` mapped back to the persona; and `worn_remove` dropping a named garment while keeping the others.
- The existing `beholder-wound-merge.regression.ts` from #5241 still passes.

Note that `pull-request-checks.yml` does not run the regression suite (only `regression:launcher-update` and `regression:android-local-auth`), so neither regression file executes in CI as things stand.

### Sequencing note

A package that offers a five-pass template must require an Engine version containing this change — otherwise the whole multi-lane template is sent as one system prompt. Happy to follow up in `Marinara-Agents` once this lands and the minimum version is known.

## Summary by CodeRabbit

* **New Features**
  * Improved Beholder state extraction across equipment, wounds, inventory, species, and flags.
  * Multiple state sections can now be processed concurrently and combined into one result.
  * Selective removal of individual worn items is supported without clearing other equipment.
* **Bug Fixes**
  * Partial processing failures no longer discard successful state updates; errors occur only when all sections fail.
  * Empty, duplicate, invalid, or unchanged state sections are handled more reliably.

### DIFF
--- a/packages/server/src/services/agents/agent-executor.ts
+++ b/packages/server/src/services/agents/agent-executor.ts
@@ -44,6 +44,7 @@ import { getAssetManifest } from "../game/asset-manifest.service.js";
 import {
   BEHOLDER_PASS_LANES,
   formatBeholderRequestContext,
+  isBeholderLaneResponse,
   mergeBeholderLaneDeltas,
   parseBeholderLanePrompts,
   resolveBeholderStateResponse,
@@ -1007,8 +1008,8 @@ async function executeBeholderLanePasses(args: {
     // parseError marker instead of throwing, and treating that as a usable
     // lane would let five broken replies look like a clean no-change turn.
     const laneData = parseAgentResponse(config, outcome.value.laneText).data;
-    if (shouldFailInvalidJsonResult(config, laneData)) {
-      logger.warn("[agent] %s pass %s returned unparseable output; skipping lane", config.type, lane);
+    if (shouldFailInvalidJsonResult(config, laneData) || !isBeholderLaneResponse(laneData)) {
+      logger.warn("[agent] %s pass %s did not answer in the extraction contract; skipping lane", config.type, lane);
       continue;
     }
     laneResponses.push(laneData);
--- a/packages/server/src/services/agents/agent-executor.ts
+++ b/packages/server/src/services/agents/agent-executor.ts
@@ -937,9 +937,19 @@ async function executeBeholderLanePasses(args: {
 
   logger.info(`[agent] ${config.type} (${config.name}) — ${model} — ${BEHOLDER_PASS_LANES.length} passes`);
 
-  const settled = await Promise.allSettled(
-    BEHOLDER_PASS_LANES.map(async (lane) => {
+  // Bounded dispatch, not Promise.all: each lane's timeout budget starts when
+  // agentCallSignal builds it, so launching all five at once would let a lane
+  // queued behind a rate limit burn its budget before the provider starts work.
+  const settled = await settleAgentJobsWithConcurrencyLimit(
+    [...BEHOLDER_PASS_LANES],
+    AGENT_BATCH_FALLBACK_MAX_CONCURRENT,
+    async (lane) => {
       const messages = buildStandardAgentMessages(config, lanePrompts[lane], context);
+      logger.debug(`[agent] ═══ ${config.type} [${lane}] PROMPT ═══`);
+      for (const msg of messages) {
+        logger.debug(`[agent] [${msg.role}] ${msg.content}`);
+      }
+      logger.debug(`[agent] ═══ END ${lane} PROMPT — temperature=${temperature} maxTokens=${maxTokens} ═══\n`);
       emitAgentDebug(context, {
         stage: "request",
         ...agentDebugBase(config, model, temperature, maxTokens),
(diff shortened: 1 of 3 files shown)


GetBeholder:
All six addressed in f964038ac — every one was valid. Verified each against the code rather than applying blind:

**Lane timeouts (`agent-executor.ts`)** — confirmed: `agentCallSignal` builds `AbortSignal.timeout(...)`, which starts counting immediately, and all five were constructed synchronously in the `.map`. Now dispatched through `settleAgentJobsWithConcurrencyLimit` at `AGENT_BATCH_FALLBACK_MAX_CONCURRENT`, so a lane's budget starts when its worker starts.

**Unparseable lanes counted as success** — the important one. `parseAgentResponse` returns a `parseError` marker rather than throwing, so a lane of garbage still incremented `laneResponses.length`, the all-failed guard never fired, and `mergeBeholderLaneDeltas` skipped everything to produce a *successful* `changed: false`. A wholly broken model would have looked like a quiet turn. Lanes are now screened with `shouldFailInvalidJsonResult` and skipped, so five bad replies hit the error path.

**Prompt logging** — added the per-lane `logger.debug` prompt dump and END PROMPT marker, matching the single-call path.

**`worn_remove` matching nothing** — correct, and it mattered: the removal is the *only* content of a takeoff turn, so leaving `used` false made `resolveBeholderStateResponse` report an unusable delta. It now marks the delta handled whether or not a garment was actually on the slot.

**Both test gaps** — added a repeated-heading case and a `worn_remove` concatenation case, plus one for the no-op removal above.

Verification: `tsc -b ../shared && tsc --noEmit` on `packages/server` → **0 errors**; prettier clean; both regressions pass. I also checked the new no-op-removal assertion actually fails against the previous guard (`A no-op removal must not invalidate the turn`) rather than passing vacuously.

GetBeholder:
Valid, and the same failure mode one level up from the previous round — fixed in the latest push.

Screening only for `parseError` still let a lane answer with valid JSON of the wrong shape (`{}`, a bare array, `{changed: true}` with no delta) and count as a usable reply. With every lane doing that, the merge skipped them all and the turn resolved as a successful `changed: false` rather than reaching the all-lanes-failure path.

Added `isBeholderLaneResponse()` next to the merger: a lane counts only when it answers `{changed: false}` or `{changed: true, delta: {...}}`. The executor now requires it alongside the existing `shouldFailInvalidJsonResult` check, and the regression covers the guard directly (valid shapes, empty object, delta-less `changed: true`, array, non-JSON, null).

Also rebased onto `staging` now that #5241 has merged, so this branch carries only its own two commits. Re-verified after the rebase: `tsc -b ../shared && tsc --noEmit` → 0 errors, prettier clean, both regressions pass.

SpicyMarinara (approved):

PR #5281, opened 2026-08-19
feat(agents): validate Beholder extractions before they reach state

GetBeholder:
## Linked issue

No separate issue — continues the Beholder parity work behind #5251 and #5280. **Stacked on #5280**; review that first, and this diff is validator-only once #5280 lands.

## Why this change

Beholder has no runtime validation. Output is schema-clamped — unknown slots dropped, enums coerced — and then merged. So anything *well-formed* is accepted, and well-formed is not the same as *possible*:

- a concussion recorded on a `back`
- a boot worn on a `left_hand`
- clothing on an amputated limb
- `"pureblood"` stored as a species

All four survive schema clamping today. And they do not stay contained: tracked state is fed back into the next extraction *and* into the next roleplay response, so one impossible entry keeps being restated until a user hand-edits it.

This matters more here than for a typical agent because the extractor is deliberately small. The reference implementation treats runtime validation as mandatory for exactly that reason.

## What changed

- **`beholder-validator.ts`** (new) — port of the validator from the reference extractor, itself a mirror of the validator that produced Beholder's training labels. 25 rules across anatomy, species, multi-slot garments, mutually exclusive clothing, missing-limb dominance and cascade, plus prose-aware severity and damage checks. Error-severity findings strip the offending path (deepest first, then descending list index so same-list strips don't shift each other); warnings and suggestions are reported only.
- **`beholder-validator-data.ts`** (new) — the constant tables, **vendored verbatim** from the generated source so they stay diffable and cannot drift by hand-editing.
- **`agent-executor.ts`** — validation runs at `resolveStructuredAgentResult`, the single point every Beholder path already funnels through, so the single-prompt and per-lane shapes are both covered. Findings are logged at debug with rule and path.
- **`beholder-state.ts`** — the persona-as-`self` keying is extracted so the prompt and the validator see the same shape (the validator compares a delta against prior state, so it must be keyed like the delta).

No schema, storage, API, or client change.

## Validation

- [ ] `node scripts/validate-catalog.mjs` passes locally
- [ ] `git diff --check` passes locally
- [ ] Rebuilt every affected manifest, payload, artifact, and catalog entry

### Manual verification notes

Those boxes are the package template's and do not apply — this is Engine code, no package touched. `pnpm install` also fails in my environment at the `onnxruntime-node` postinstall (it rejects CUDA 13), so I cannot run the full `pnpm check` or the app.

What I did run:

- **Parity gated, not asserted.** `scripts/regressions/beholder-validator-parity.regression.ts` (new) runs the port against the fixture generated from the source validator, vendored unchanged, and compares **both** the findings (rule id, path, severity) and the **full stripped delta**: **60/60 cases match**. Comparison is scoped to the ported rule set, so a rule the port has not taken on shows as a gap rather than a false mismatch.
- **Typecheck clean**: `tsc -b ../shared && tsc --noEmit` for `packages/server` → 0 errors. Prettier clean.
- The existing `beholder-normalizer`, `beholder-user-message`, `beholder-lane-passes` and `beholder-wound-merge` regressions still pass.

Not verified: behaviour against a live model. The judgement worth a maintainer's eye is that error-severity findings **strip silently** (reported at debug level, not surfaced in the UI) — that is the reference behaviour, but a user seeing an item quietly not appear has no in-app explanation yet. Surfacing findings is a natural follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added validation for generated character state, including anatomy, equipment, injuries, species, and consistency checks.
  * Invalid details are automatically removed while valid state is preserved.
  * Added support for character-specific prior state and locked equipment slots.

* **Bug Fixes**
  * Prevented invalid findings from being persisted in tracked character state.
  * Improved handling of narration-derived state and conflicting equipment details.

* **Tests**
  * Added comprehensive regression coverage for validation and state-cleanup scenarios.


SpicyMarinara (approved):

GetBeholder:
@coderabbitai review

SpicyMarinara:
@coderabbitai review

GetBeholder:
Closing as stale.

Same reasoning as #5284 — behind `staging` since 19 Aug. Worth recording what this would have bought, so the trade-off is explicit rather than forgotten:

it carried a garment→slot map (`boots: ["left_foot", "right_foot"]`) with error-severity findings, which would catch a real extraction failure I reproduced today. Given the prose *"hangs the cloak on a hook and pulls on black boots"*, the extractor drops the removal and puts the boots on `chest` and `back`. The add half is a mis-slotting this validator would have rejected.

That is being addressed where it belongs — as an extraction quality problem in the model's training data, not as a validation layer added to the Engine. If runtime validation is wanted later, it should be weighed as its own decision rather than carried in on the back of this branch.

PR #5284, opened 2026-08-19
feat(agents): match the reference extractor's merge semantics and guards

GetBeholder:
## Linked issue

No separate issue — continues the Beholder parity work behind #5251, #5280 and #5281. **Stacked on #5281**; review that first.

## Why this change

Five ways tracked state drifted from what the story actually said. Each survives schema validation, so none of them errors — the state just quietly stops matching the fiction, and because it is fed back into the next extraction and the next response, it stays wrong.

1. **An update erased what was already known.** A delta re-describing a garment replaced the stored entry outright, so `{"item":"blazer","damage":"damaged"}` — "the blazer is torn" — dropped the navy the blazer was already known to be. Same class as the wounds bug in #5241, on the other field.
2. **Plurals stacked.** Garment identity was the raw name, so `"boots"` appended a second entry beside a stored `"boot"` and the character was wearing two.
3. **A slot could be bare *and* clothed**, with nothing to resolve the contradiction.
4. **`missing` was trusted.** Amputation is the field the extractor misfires on most, and a wrong one cascades and keeps being restated.
5. **Characters could be invented.** On a cold start or a first-person turn the extractor can attribute state to a name appearing nowhere in the story, and once written it persists.
6. **The validator ate takeoffs.** An empty `worn` list is how "took it all off" is expressed; pruning removed it along with lists that were only emptied by stripping an invalid item.

## What changed

- **Worn and wounds update in place** — incoming fields win, fields the delta did not re-emit are kept, and the first-seen surface form is kept so a label does not flicker between "boot" and "boots".
- **Canonical garment identity**, so singular and plural are one garment. `worn_remove` matches canonically too: removing "boots" clears a stored "boot".
- **Bare/worn reconciliation** — a delta asserting `bare` empties the stack; a garment added clears a stale `bare`. Slots the delta did not touch are left alone, because only a fresh assertion resolves the contradiction.
- **`stripModelMissing`** — model-emitted `missing` is dropped before anything else sees it, matching the reference, which makes it manual-only. The slot editor and directives still set it.
- **`dropHallucinatedCharacters`** — a character is kept when it is the persona, already tracked, or named in the narration; anything else is dropped and logged. A tracked character survives a turn that never names it, since a pronoun-only sentence attributes by context.
- **`keepOriginallyEmpty`** in the validator — the model's own clears survive pruning, while a list emptied by *stripping* still collapses to a no-op so a bad emission cannot wipe a stack.

No schema, storage, API, or client change.

## Validation

### Manual verification notes

Boxes left for a human — `pnpm install` fails in my environment at the `onnxruntime-node` postinstall (it rejects CUDA 13), so I cannot run the full `pnpm check` or the app.

What I did run:

- **`scripts/regressions/beholder-merge-parity.regression.ts`** (new, passing) — covers all five: colour surviving a damage update, co-located garments surviving, plural folding onto the stored singular with the surface form kept, canonical `worn_remove`, bare emptying the stack and a garment clearing a stale bare, `missing` stripped with the rest of the slot intact, an explicit clear surviving the validator, a strip-emptied list *not* surviving, and the invented-character guard (whole-word matching, regex-significant names, persona and tracked names kept).
- **Each assertion was checked against the old behaviour**, not just the new: reverting the field-merge fails with `an update keeps fields the delta did not re-emit`; reverting canonical identity fails with `"boots" updates the stored "boot" instead of stacking`; removing the reconciliation fails with `going bare empties the worn stack`.
- **The validator parity fixture still passes 60/60** — `keepOriginallyEmpty` changes pruning without moving any oracle case.
- Typecheck clean (`tsc -b ../shared && tsc --noEmit` → 0 errors), prettier clean, and the four existing Beholder regressions still pass.

Not verified: behaviour against a live model. The judgement worth a maintainer's eye is `stripModelMissing` — it is a deliberate capability *removal* (the model can no longer report amputation at all), taken from the reference because the false-positive rate there is worse than the missing feature.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of extracted character state by filtering hallucinated characters and invalid fields.
  * Preserved garment, wound, and clothing details more consistently during updates and removals.
  * Prevented invalid anatomy, item placement, severity, and holding-state changes from corrupting saved state.
  * Preserved intentional empty-list updates while removing unsupported model-generated values.
  * Improved handling of conflicting worn and bare clothing states.

* **Tests**
  * Added comprehensive regression coverage for state merging, validation, filtering, and cleanup across 38 scenarios.


SpicyMarinara (approved):

GetBeholder:
@coderabbitai review

GetBeholder:
@coderabbitai review

GetBeholder:
Both Major findings were real and are fixed in `aaee1516a`. Thanks — the first one contradicted this PR's own claim.

**Normalizer defaults overwriting omitted fields.** Correct, and it undercut the whole point of the change: `damage`, `severity` and `bleeding` were always emitted with schema defaults, then spread over the stored entry, so "the blazer is torn" reset a garment known to be `broken` back to `pristine`, and re-describing a wound without a severity quietly downgraded it. `material` and `color` escaped only because they are spread conditionally.

Fixed at the root you identified: defaulting now happens when normalizing **stored state**, not when normalizing a **delta** (`normalizeWornItem`/`normalizeWound` take an `applyDefaults` flag, false on the delta path). An entry appearing for the first time is new state, so it still gets the defaults.

**Structured paths for character keys.** Also correct. `cleanText` keeps `.` and `[`, so a name like `Dr. Vance` or `A[B]` tokenized *inside the name*, the lookup missed, and the offending field was reported but never stripped — validation that looks like it works and silently does not.

Rather than change the finding shape (the parity fixture compares `path` strings), `removePath` now resolves the character segment against the delta's **actual keys**, longest match first, and parses only what follows. Slots and fields are canonical, so the character key is the only ambiguous segment.

Verification: the validator fixture still passes **60/60**, typecheck is clean, and all six Beholder regressions pass. New cases cover an omitted `damage` keeping `broken`, an omitted `severity`/`bleeding` not downgrading a wound, a first-seen garment still defaulting to `pristine`, and stripping through `Dr. Vance`, `A[B]` and `St. John-Smith`. I checked each new assertion fails against the previous behaviour rather than passing vacuously.

I also rebuilt on top of the `staging` merge that was pushed to this branch rather than force-pushing over it.

GetBeholder:
Closing as stale rather than rebasing it in.

This has been sitting behind `staging` since 19 Aug, and the parts of it that mattered most are already in: `beholder-state.ts` merges wounds by identity and handles `worn_remove` on the current branch. Landing the remainder now would mean re-validating a three-day-old diff against a moved target for a benefit we can no longer point at.

The branch stays if any specific guard here turns out to be wanted; it can be reopened as a narrow change with a concrete failure behind it.

PR #5305, opened 2026-08-20
feat(chat): render the Beholder paper doll and open it from the toolbar

GetBeholder:
## Linked issue

No separate issue — continues the Beholder work behind #5251, #5280, #5281 and #5284.

## Why this change

The panel rendered tracked state as a text list. The reference extractor draws a **body**: a silhouette with a region per slot, tinted so an **outline** means the armour there is damaged and a **fill** means the body itself is hurt, the slot cards arranged around the figure, and a front/back flip for wounds on the back. Reading a character was meant to be looking at a figure, not parsing a list — and the distinction that matters in play (which side, how bad, covered or not) is exactly what a list flattens away.

My first attempt at this replaced that with a flat grid of my own design. That was the wrong call: there is no technical reason the original cannot be used, and it is a better design than what I substituted.

## What changed

- **`paperdoll.ts`** — the reference renderer, ported. It is the source file with type annotations added and **nothing else changed**, except: two functions dead in the original (`formatSlotList`, `rowSide`) dropped so it compiles under `noUnusedLocals`, one unused parameter renamed, and `escapeHtml` rewritten as an equivalent lookup table. Species families, off-body slots, the missing-limb cascade, damage tiers, wound severity encoding and the digest all come across as they are.
- **`paperdoll.css`** — its stylesheet, plus the design tokens it depends on, scoped to `.beholder-panel` so nothing outside the panel is affected.
- **`colors.ts`, `garment-data.ts`, `state.ts`** — the small modules it imports.
- **`BeholderDoll.tsx`** — a thin component that feeds it state and delegates the controls the markup already draws: front/back, character tabs, layout switch.
- The panel now renders that instead of a list.

Keeping it as markup rather than rebuilding it in JSX is deliberate: it stays diffable against the implementation it came from, so the two cannot quietly drift.

## Validation

### Manual verification notes

Boxes left for a human, and for a UI change that matters: **I have not seen this render.** `pnpm install` fails in my environment at the `onnxruntime-node` postinstall (it rejects CUDA 13), so I cannot run the app. It needs eyes in both themes and at drawer width before it is trusted.

What I could verify without that:

- **The port is faithful, and I checked rather than assumed.** Diffing the ported file against its source with type annotations stripped leaves only the changes listed above — no logic differs.
- **Formatting changed nothing.** Prettier reformatted the ported file to satisfy `format:check`; rendering the same state across all six layout/view combinations before and after produced a **byte-identical 100 KB of markup** (same SHA-256).
- **It renders.** A headless run produces the silhouette (`<svg class="bh-silhouette">`, `viewBox`), per-slot regions (`data-slot="chest"`), the missing hatch (`bh-part-missing`), and worn/wound/held content. A character name containing `<script>` comes out escaped.
- **The client builds.** Vite bundles it; the doll and the dock are code-split with their own CSS chunks.
- Client typecheck **0 errors**; prettier clean; `check-locales.mjs` clean.

### Note on appearance

The doll brings its own palette and spacing tokens rather than adopting the host theme, because the point was to render as designed. If it should instead follow Marinara's light/dark tokens, that is a straightforward change to the token block — but it is a design decision, not a port decision, so I have not made it unilaterally.

## Opening it

The doll previously rendered inside the chat settings drawer, under the agent's row — findable only if you knew it was there, and impossible to keep open while playing, which is the only time it is useful.

It now opens from a **button in the chat toolbar**, beside the other panel controls, into a **dock on the right** — which is where the reference extractor puts it in this host. It docks rather than floats because Marinara already knows how to make room for a right-side panel: opening it sets the same variable the tracker panel uses, so the chat shifts instead of being covered. The dock measures the app bar for its top offset and whatever already occupies the right edge for its horizontal offset, so it never parks on top of another panel. A narrow viewport gets the compact list layout and the dock becomes a bottom sheet.

The button appears only when Beholder is tracking the chat — the state it shows comes from that agent, so otherwise it would open empty while taking a toolbar slot from everyone who does not use it.

### Not here yet

Per-slot manual editing and locks, and the directive bar.

## Summary by CodeRabbit

- **New Features**
  - Added a Beholder paper-doll display for equipment, body parts, wounds, damage, missing limbs, and held items.
  - Added access through the conversation toolbar, with responsive desktop and mobile dock views.
  - Added front/back views, character selection, and paired, column, and list layouts.
  - Layout preferences now persist between sessions.
  - Added responsive rendering and clearer indicators for missing or damaged areas.

- **Bug Fixes**
  - Improved handling of alternate garment names and color descriptions.
  - Added clearer loading, error, empty, and unavailable-character states.
  - Added derived status for dependent missing body parts.


SpicyMarinara (approved):

GetBeholder:
Closing this in favour of Pasta-Devs/Marinara-Agents#496, which ships the same UI from the Beholder package instead of the Engine client.

This PR hardcoded the paper doll into the client (`components/chat/beholder/*` plus a mount in the chat surface). That was the wrong shape: Beholder is an optional agent, so its panel should appear only for people who installed it and switched it on, and the Engine should not carry UI for a package that most installs never add.

The capability path already does exactly that — the package declares `entrypoints.client` and the `conversation-toolbar` slot, and `CapabilityElement` mounts it only when the agent is active on that chat. Same doll, same dock behaviour, no Engine-side Beholder code.

The server-side work stays where it is: #5281 (validation) and #5284 (merge semantics) are unaffected by this.

PR #5602, opened 2026-08-28
feat(server): add optional outbound proxy support (MARINARA_PROXY)

### LINKED ISSUE #5588: [Feature]: Add optional outbound proxy support (`MARINARA_PROXY`)

On restricted networks (for example mainland China), `raw.githubusercontent.com` is often unreachable, which breaks agent and capability-package installation:

- `package-manager.service.ts`'s `fetchBytes()` calls `safeFetch()`, which is backed by undici's `fetch` plus a self-built `Agent` — not Node's native `http` or `axios`.
- The download target is fixed to `raw.githubusercontent.com/Pasta-Devs/Marinara-Agents`.

undici ignores both the OS/system proxy and the `HTTP_PROXY` / `HTTPS_PROXY` environment variables, so a direct connection to GitHub times out. Marinara Engine currently has no proxy path at all, and manually downloading [Marinara-Agents](https://github.com/Pasta-Devs/Marinara-Agents) does not import correctly either (there may be a correct manual-import workflow I have not found).

It would be much more convenient if outbound requests could optionally route through a local proxy, so users on restricted networks can pull agents and artifacts — and, when needed, send model requests through the same proxy.

## Proposed solution

Add an environment variable, tentatively named **MARINARA_PROXY**. When it is set, `safeFetch` swaps its self-built `Agent` for undici's `Socks5ProxyAgent` (SOCKS5) or `ProxyAgent` (HTTP/HTTPS) as the dispatcher; when it is unset, behavior is identical to today.

- Supported schemes: `socks5://`, `socks5h://`, `http://`, `https://`
- Read order: `MARINARA_PROXY` → `HTTPS_PROXY` → `ALL_PROXY`
- Read lazily: `process.env` is checked per request, so `.env` changes take effect within ~2s without a restart
- `policy.allowLocal` requests (Ollama / LM Studio / loopback) never go through the proxy

```diff
- dispatcher?: Agent,
+ dispatcher?: Dispatcher,
```

## Alternatives considered

I tried several alternative approaches with Claude's help — TUN proxying, injection/hooking, and so on — but each introduced additional side effects. A source-level change that adds one optional proxy branch is the only approach with no side effects and is user-friendly.

## Additional context

1. **Unchanged when no proxy is configured**: `resolveOutboundProxyUrl()` returns `null`, so the existing `new Agent({...})` direct path runs exactly as before.
2. **Local requests unaffected**: when `policy.allowLocal` is true (Ollama / LM Studio / loopback), the original direct branch runs and never touches the proxy.
3. **SSRF protection unchanged**: protocol allowlists, hostname allowlists, and the private/loopback/reserved-IP DNS checks (`validateOutboundUrl` + `validateResolvedAddresses`) all run before the proxy branch.
4. **Timeouts still guaranteed**: the proxy branch no longer passes `agentOptions.bodyTimeout`/`headersTimeout`, but every caller already supplies `signal: AbortSignal.timeout(...)`.
5. **New failure modes are limited**: only an explicitly configured malformed or unsupported proxy URL raises a descriptive error (for example `Unsupported outbound proxy protocol ...`).

## Template check

stbanana:
# PR title

feat(server): add optional outbound proxy support (MARINARA_PROXY)

---

> [!IMPORTANT]

## Linked issue

Closes #5588

## Why this change

On networks where direct GitHub (or provider) access is blocked, Marinara Engine cannot download agents or capability packages, and provider requests can time out: `safeFetch()` is backed by undici's `fetch` plus a self-built `Agent`, and undici ignores both the OS proxy and the `HTTP_PROXY`/`HTTPS_PROXY` environment variables.

This adds an opt-in outbound proxy so restricted-network installs can route non-local outbound requests through a local SOCKS5 or HTTP(S) proxy. Behavior is unchanged when no proxy is configured.

## What changed

- `packages/server/src/utils/security.ts`
  - Import `ProxyAgent`, `Socks5ProxyAgent`, and `type Dispatcher` from `undici`.
  - Add `resolveOutboundProxyUrl()` (reads `MARINARA_PROXY` → `HTTPS_PROXY` → `ALL_PROXY`) and `createOutboundProxyDispatcher()` (SOCKS5/SOCKS5H/HTTP/HTTPS).
  - In `validateOutboundUrlForFetch()`, after SSRF validation, return a proxy dispatcher when a proxy is configured; otherwise keep the existing direct `Agent` path unchanged.
  - Widen the dispatcher parameter of `readCappedResponse()` / `capStreamingResponse()` from `Agent` to `Dispatcher`.
- `.env.example` — document the new `MARINARA_PROXY` setting.
- `docs/CONFIGURATION.md` — document the setting and its scope.

Reviewer notes:

- SSRF validation (protocol/hostname allowlists plus DNS and private/loopback/reserved-IP checks) still runs before the proxy branch, and `policy.allowLocal` requests (Ollama, LM Studio, loopback) never use the proxy.
- The proxy applies to **all** non-local outbound fetches — provider calls, media, wiki, docs — not just GitHub downloads. This is intentional.
- When `MARINARA_PROXY` is unset, the standard `HTTPS_PROXY` / `ALL_PROXY` env vars are honored, so an ambient host proxy is picked up automatically.

## Validation

### Manual verification notes

- [x] Verify `Download Agents` lists packages when `MARINARA_PROXY=socks5://...` is set on a network that blocks direct GitHub access.
- [x] Verify an agent/package download succeeds through the proxy.
- [x] Verify behavior is unchanged (direct path) when `MARINARA_PROXY` is empty or unset.
- [ ] Verify a local provider (e.g. Ollama) still connects directly and is not routed through the proxy.
- [ ] Verify an unsupported or malformed proxy URL raises the descriptive error.

## Docs and release impact

No UI changes.

## Summary by CodeRabbit

* **New Features**
  * Added optional outbound proxy configuration through `MARINARA_PROXY`.
  * Supports SOCKS5, SOCKS5H, HTTP, and HTTPS proxies for external requests, including provider calls and downloads.
  * Automatically falls back to `HTTPS_PROXY` or `ALL_PROXY` when the new setting is not configured.
* **Documentation**
  * Documented proxy configuration, supported formats, local-network bypass behavior, and configuration error reporting.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -248,3 +248,15 @@ GIPHY_API_KEY=
 #     or set SPOTIFY_REDIRECT_URI to your HTTPS URL).
 # Must exactly match a Redirect URI registered in your Spotify Developer Dashboard app.
 SPOTIFY_REDIRECT_URI=
+
+# Optional outbound proxy for non-local requests: agent and capability package
+# downloads, AI provider calls, media generation, and other HTTPS fetches. When
+# set, Marinara routes those requests through a SOCKS5, SOCKS5H, HTTP, or HTTPS
+# proxy instead of connecting directly. Useful when outbound GitHub or provider
+# access is blocked. Leave empty to connect directly. Examples:
+#   MARINARA_PROXY=socks5://127.0.0.1:1080
+#   MARINARA_PROXY=http://127.0.0.1:7890
+# Loopback and private-network provider requests always connect directly and
+# never use the proxy. When this is empty, the server also honors the standard
+# HTTPS_PROXY and ALL_PROXY environment variables.
+MARINARA_PROXY=
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 - Character cards now support editable metadata summaries, AI-generated summary drafts, and Character Library previews that use the saved summary when available.
 - Character card Conversation profiles now provide controls to generate About Me and Conversation behavior text from the card's available information.
 - The Home Character of the Day widget now uses saved character summaries and offers direct chat-start and character-view actions.
+- Outbound requests can now route through an optional local proxy via the new `MARINARA_PROXY` setting (SOCKS5/SOCKS5H/HTTP/HTTPS, with `HTTPS_PROXY`/`ALL_PROXY` fallback), so agent and capability-package downloads and provider calls work on networks that block direct GitHub access. Loopback and private-network provider requests always connect directly (#5588).
 
 ### Fixed
 
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -316,6 +316,7 @@ Scene video providers are set up as connections inside the app, not as environme
 | Variable                          | Default                                    | What it does                                                                      |
 | --------------------------------- | ------------------------------------------ | --------------------------------------------------------------------------------- |
 | `DOCS_I18N_BASE_URL`              | official `docs-i18n` branch                | Where translated documentation packs download from (Settings → General → Documentation Language). Must be a public `https://` host; forks and mirrors can point it at their own copy of the `docs-i18n` branch. |
+| `MARINARA_PROXY`                  | empty                                      | Optional outbound proxy for non-local requests. Accepts `socks5://`, `socks5h://`, `http://`, or `https://`; local and private-network provider requests never use it. See below. |
 | `GIPHY_API_KEY`                   | empty                                      | Giphy key for GIF search in Conversation mode. Search is off when unset.          |
 | `INTIFACE_URL`                    | `ws://127.0.0.1:12345`                     | Default address for the Intiface haptic app.                                      |
 | `SPOTIFY_REDIRECT_URI`            | derived from request                       | Override for the Spotify login callback URL. Set it when TLS is handled upstream. |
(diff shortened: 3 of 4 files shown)


SpicyMarinara:
Maintainer review: thank you for tackling a real restricted-network problem. The direction makes sense, but this PR needs changes before owner approval or merge.

Blocking correctness and security issues:

1. **The proxied request is not bound to the SSRF-validated destination.** The direct path pins Undici to an address returned by `validateResolvedAddresses`; the proxy branch discards those addresses and sends the hostname to the proxy for separate resolution. Split-horizon or proxy-side DNS can therefore reach a private destination after server-side validation passes.
2. **Local/public routing is inferred from the wrong value.** `policy.allowLocal` is a permission flag, not a destination classification. With it enabled, public provider calls bypass the proxy. Meanwhile, loopback or mDNS destinations allowed through `allowLoopback` / `allowMdns` while `allowLocal` is false can be sent to the proxy, potentially targeting the proxy host local services instead of Ollama or LM Studio.
3. **The proxy branch drops dispatcher behavior.** `agentOptions` and `keepAliveInitialDelayMs` are ignored. LLM requests rely on header/body timeouts and do not universally provide an abort deadline, so the claim that all timeouts remain guaranteed is not correct.
4. **The documented opt-in behavior is contradictory.** Leaving `MARINARA_PROXY` empty can still activate ambient `HTTPS_PROXY` or `ALL_PROXY`, while `.env.example` says an empty value connects directly. There is no `NO_PROXY` handling.
5. **Invalid proxy URLs can leak credentials** because the full configured value is included in the error. The unsupported-protocol message also omits accepted `socks5h://`.

Required before re-review:

- Add focused automated regression coverage for the cases above, including differing server/proxy DNS results and redirect hops.
- Preserve supported dispatcher timeouts/options, or reject unsupported combinations explicitly.
- Base local bypass on the validated destination and keep SSRF enforcement effective through the proxy.
- Rebase onto current `staging` and resolve the existing conflicts.
- Update the `docs-i18n` branch or open the required `[docs-i18n] docs/CONFIGURATION.md` follow-up.
- Leave AI-generated validation/test-plan checkboxes unchecked and describe evidence in prose, per repository policy.

To keep this small and safe, I recommend scoping the first iteration to the official Marinara-Agents catalog and artifact downloads through explicit `MARINARA_PROXY`, restricted to exact trusted GitHub origins. Provider-wide proxying can be handled separately once its broader trust-boundary behavior is covered.

Issue #5618, opened 2026-08-29
Two smoke tests still flake after #5595: art-scale sliders (chromium) and game-widget Chroma visibility (webkit)


kolacheee:
**What happens**

#5595 fixed the eight webkit-only smoke failures and staging's Release Regression Matrix went green for the first time in the retained history (run 33223048649 on `ec38ff7d5`). But at least two *other* tests still fail intermittently, so PR branches keep inheriting occasional red mobile lanes that have nothing to do with their diffs — the #5589 problem at a smaller scale:

- `e2e/core-flows.e2e.ts:398` — "Art scale sliders stay interactive at the largest display size" (**mobile-chromium**)
- `e2e/core-flows.e2e.ts:8608` — "Game widget editing and log deletion follow Chroma while weather effects render" (**mobile-webkit**, `expect(locator).toBeVisible()` timeout)

**Evidence these are the tests, not the PRs**

On PR #5614 (server-storage-only diff), mobile-chromium **passed** the full suite on `48cb34603`, then **failed `:398`** on `7f998e025` — whose only delta is a query-param guard and a comment. Same run: mobile-chromium *passed* `:8608` while mobile-webkit failed it — a server bug would fail both engines, an engine-specific visibility timeout won't. PR #5617 (config/docs only, zero code) ran both mobile lanes green minutes later.

**Try it yourself** (both reproduce on current `staging`, no PR changes involved)

```bash
pnpm exec playwright test -c playwright.config.ts --project=mobile-chromium -g "Art scale sliders stay interactive" --retries=0
```

```bash
pnpm exec playwright test -c playwright.config.ts --project=mobile-webkit -g "Game widget editing and log deletion" --retries=0
```

Both failed on a Windows 11 checkout of staging `672c8d156` (run each a few times on Linux to see the intermittency CI shows — in CI they pass more often than they fail).

**Notes for whoever picks this up**

`:8608` layers weather-effect animation on top of a visibility assert, which is the classic shape of the webkit timeouts #5595 just fixed elsewhere — the same stabilization approach probably transfers. `:398` drives sliders at the largest display size on the mobile-chromium viewport; the 19s duration before failure suggests a wait expiring rather than a wrong assert.

Follow-up to #5589 (closed by #5595's merge).

kolacheee:
Taking this one (kolacheee, agent-assisted) — both tests reproduce locally, debugging now. PR incoming.

kolacheee:
[...]

Issue #5589, opened 2026-08-28
[Issue]: Release Regression Matrix has been red on every staging push since at least 2026-08-25


Romuromylus:
## Summary

The Release Regression Matrix workflow (the Playwright smoke lanes: `desktop-chromium`, `mobile-chromium`, `mobile-webkit`) has not had a single green run on `staging` in all of the retained run history — 14 failures and 6 cancellations since 2026-08-25, zero successes. Every PR inherits the red matrix, which trains us to merge over failing checks and buries any *real* e2e regression a PR introduces.

## Expected behavior

The smoke lanes pass on the `staging` head, so a red lane on a PR is meaningful signal.

## Actual behavior

Every staging push since at least `c47dedc8a` (2026-08-25T13:48Z) fails the matrix. The failing test set rotates between runs, which points at a mix of at least one hard breakage plus flaky cases:

- Consistently failing right now (all three browser lanes, staging head `e390e3619` and PR branches alike, across reruns): `e2e/core-flows.e2e.ts:14754` — "Professor Mari shows the latest context budget when token usage is enabled" fails with `expect(locator).toContainText` → element(s) not found.
- Rotating/flaky examples from other runs: `core-flows.e2e.ts:1765` "Conversation message actions follow their messages" (toHaveCSS on the hover-gated `mari-message-actions` container), `:3299` avatar-action balance, `:7172` chat Help overlay labels, `:13756` Illustrator subsections, `:13966` World Maps placement, `:16580` Character of the Day centering, `:18209` mobile composer history focus.

## Steps to reproduce

1. Open the Actions history for the Release Regression Matrix workflow filtered to `staging`.
2. Observe 0 green runs in the retained history (2026-08-25 onward).
3. Re-run the failed lanes on any recent run: `core-flows.e2e.ts:14754` fails again in all three browser projects, while other cases rotate.

## Environment

- Marinara Engine version: staging `e390e3619` (post-v2.4.4)
- Install type: source (GitHub Actions CI)
- OS + version: ubuntu-latest runners
- Browser or app shell: Playwright desktop-chromium / mobile-chromium / mobile-webkit projects

## Logs, screenshots, or video

Example runs:

- staging `e390e3619`: https://github.com/Pasta-Devs/Marinara-Engine/actions/runs/33129778519 (all three lanes fail; `14754` among the failures)
[...]

PR #5630, opened 2026-08-29
ci: shard the browser matrix, cache browsers, and gate unaffected PRs

### LINKED ISSUE #5622: Speed up the browser CI matrix: shard the lanes, build once, and skip on unrelated PRs
**What happens**

The Release Regression Matrix (playwright.yml) takes 15–21 minutes per browser lane on every PR — mobile-chromium ~15 min, mobile-webkit ~20 min — regardless of what the PR touches. A docs-only or installer-only PR waits the same 20 minutes as a client rewrite.

**Why it's slow today** (verified against the current workflow/config)

- Each lane runs the full ~103-test `core-flows.e2e.ts` suite **serially in a single worker**: `fullyParallel: false` in playwright.config.ts (deliberate — tests share one app server's mutable fixtures within a run), one job per browser.
- The workflow has **no path filtering**: `on: pull_request / push` to staging, nothing else.
- Each of the three jobs independently does its own install + client/server build before testing.
- WebKit-on-Linux is inherently slower, and the two known flaky tests (#5618) add retry time — a 19-second test retried five times is ~1.6 wasted minutes per affected run.

**Proposed work, in order of impact**

1. **Shard each browser lane** (`--shard=1/3` or `1/4` as extra matrix dimensions). This sidesteps the `fullyParallel: false` constraint cleanly: that setting guards test isolation *within one run's shared server*, but each shard job is its own runner booting its own server, so shards need no test-level isolation. 3 browsers × 3 shards ≈ **~6 minute wall time**; ×4 approaches 5. Cost: more total runner-minutes (parallel), plus a blob-report merge step if unified reports are wanted.
2. **Build once, reuse**: a single build job uploads the built client/server as an artifact; browser jobs download instead of rebuilding. Compounds with sharding (9–12 jobs would otherwise each rebuild).
3. **Cache Playwright browsers** (`~/.cache/ms-playwright`, keyed on the lockfile) if not already cached — the WebKit download alone is worth 1–2 min/job. (pnpm is already cached.)
4. **Skip gate for genuinely unrelated PRs**: a cheap change-detection job (`dorny/paths-filter` or `git diff --name-only`) that the browser jobs `needs:`/`if:` on. Two constraints:
   - **Do not use a workflow-level `paths:` filter** if these are required checks — a skipped workflow leaves required checks stuck as "Expected" and blocks merging. The in-workflow gate keeps the jobs reporting green in seconds instead.
   - The safe skip-list is **narrow**, because the suite boots the real server and client, so server changes can break it. Genuinely skippable: `docs/**`, `**/*.md`, `CHANGELOG.md`, `win/installer/**`, `android/**` (WebView shell), `scripts/regressions/**` (has its own lane), and workflow files outside the matrix itself. That covers the docs-i18n mirrors, installer, and launcher-adjacent PRs where the wait is most pointless.

**Optional policy variant** (separate decision): tag a fast smoke subset for PRs (~3–4 min) and keep the full matrix on staging pushes (already running) plus a nightly. Full coverage would still gate what ships, but red lanes would surface post-merge — which cuts against the never-merge-over-red discipline #5589 restored, so this one deserves an explicit maintainer call rather than riding along.

**Also worth doing first for free minutes**: fixing the two #5618 flakes trims the retry waste out of every affected run.

Scope: contained to `.github/workflows/playwright.yml` plus a report-merge step; no test-content changes required for items 1–4.

kolacheee:
Closes #5622 — items 1–4. The smoke-subset policy variant is deliberately **not** included, per maintainer decision.

## What changed

`playwright.yml` is restructured around a shared composite action (`.github/actions/playwright-shard`); the node lane is untouched.

**1. Sharding (the big one).** Each browser lane now runs as 3 parallel shard jobs (`--shard=k/3`), taking wall time from 15–21 min toward the length of one third of the suite. One discovery en route: with the config's `fullyParallel: false`, Playwright's shard unit is the *file* — and this suite is one 19,800-line file, so plain `--shard` put all 197 tests in shard 1 and zero elsewhere. The fix is `--fully-parallel --workers=1` on the CI invocation only: `--fully-parallel` changes just the sharding granularity to individual tests, while `--workers=1` keeps execution inside each shard strictly serial in file order — the no-concurrent-tests-against-one-server property that `fullyParallel: false` exists to protect. Each shard job is its own runner booting its own servers with a fresh data dir. Distribution verified: 66/66/65.

**2. "Build once" — implemented as its honest version.** The issue's premise that each job rebuilds the app turned out to overstate: the Playwright webServer builds only the shared package (~40s) and then runs *dev* servers; the client is never production-built in this workflow. A serial prebuild job would add more gate latency than nine parallel 40-second shared builds cost, so there is no artifact hand-off — the per-job work that actually dominated (browser downloads, suite runtime) is what items 1 and 3 remove. Prebuilding the client for tests would mean testing `vite preview` instead of the dev server the suite was written against — out of scope.

**3. Playwright browser cache.** `~/.cache/ms-playwright`, keyed on OS + browser + the installed `@playwright/test` version. On a hit, `playwright install` skips the download (the WebKit bundle is the chunky one); `--with-deps`' OS packages still install, as they must.

**4. Skip gate for unaffected PRs.** A fast `browser-lane-gate` job diffs the PR's merge commit against its first parent (exactly the PR's change set) and skips the shard jobs when **every** changed file matches the narrow skip list: `docs/**`, `**/*.md`, `.gitattributes`, `win/**`, `android/**`, `scripts/regressions/**` — files the running app can never load. Anything else (server, client, e2e, workflows, lockfile, launcher scripts) runs the lanes; so does any push, `workflow_dispatch`, or a diff the gate can't determine (conservative fallback). This is an in-workflow gate, **not** a workflow-level `paths:` filter, for the required-checks reason documented in the issue.

**Branch-protection compatibility:** the shard jobs have new names (`desktop-chromium 1/3`, …), but three fan-in jobs carry the exact pre-existing check names — `desktop-chromium`, `mobile-chromium`, `mobile-webkit`. Each passes when its shards all passed, passes with a "skipped: cannot affect the browser lanes" note when the gate proved it, and fails otherwise — including when the gate job itself fails, so a broken gate can never green the matrix.

## The honest risk, and how it was validated

Mid-file shards start against a fresh server with none of the earlier tests' accumulated state, so any test that silently depended on a predecessor's side effects would fail in shards 2/3. All three desktop-chromium shards were run locally from cold starts: **shard 2 — 64/64 passed, shard 3 — 54/54 passed** (remainder skipped by project gates), **shard 1 — 63 passed with one failure of the pre-existing "Echo Chamber size persistence" pixel-tolerance flake, which passes 3/3 solo** — an independent intermittent, not order coupling (worth its own #5618-style look if it recurs in CI). Two earlier local shard-2 runs failed en masse with `ECONNREFUSED` — that turned out to be this Windows machine's known consecutive-run server-startup artifact, not the sharding (the clean re-runs above are decisive). The mobile lanes run the same single file with the same ordering profile, and this PR's own CI run exercises all nine shard jobs end-to-end on the runners that matter.

## Validation done

- YAML parse-validated; gate pattern logic unit-tested against 10 change-set shapes (docs-only/installer/regressions-only skip; server/client/e2e/workflow/lockfile/launcher run)
- Shard distribution verified 66/66/65 per project
- All 3 desktop-chromium shards run locally to completion
- This PR's CI run is itself the end-to-end proof: it exercises the gate (this PR touches workflows → lanes run), all nine shards, the cache (cold then warm on re-run), and the fan-ins

## Manually verify

- [ ] Confirm the fan-in checks (`desktop-chromium`, `mobile-chromium`, `mobile-webkit`) appear on this PR and match whatever branch protection requires
- [ ] After merge, open a docs-only PR and confirm the shard jobs skip while the fan-ins report green with the skip note
- [ ] Compare wall time of this PR's matrix against a recent pre-change run
- [ ] Re-run a lane once to confirm the browser cache hits (the install step should skip the download)

## Summary by CodeRabbit

* **New Features**
  * Added parallelized browser testing across multiple shards for faster validation.
  * Added automatic browser setup, caching, and failure-evidence uploads.
  * Added change detection to skip browser checks for documentation-only and other non-impacting changes.
  * Preserved clear browser check results through consolidated success or failure reporting.

### DIFF
--- a/.github/actions/playwright-shard/action.yml
+++ b/.github/actions/playwright-shard/action.yml
@@ -1,0 +1,68 @@
+name: Run one Playwright shard
+description: >-
+  Set up pnpm/Node, restore the Playwright browser cache, and run one shard of
+  one browser project. Each shard job runs on its own runner with its own app
+  servers, which is what makes sharding safe despite fullyParallel: false —
+  that setting guards test isolation within ONE run's shared servers, and
+  shards never share a run (#5622).
+inputs:
+  project:
+    description: Playwright project name (desktop-chromium, mobile-chromium, mobile-webkit)
+    required: true
+  browser:
+    description: Browser bundle to install (chromium, webkit)
+    required: true
+  shard:
+    description: 1-based shard index
+    required: true
+  total-shards:
+    description: Total shard count
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Set up pnpm
+      uses: pnpm/action-setup@0977fd99725f1db4007ccb2928dbb4e90d06cc86 # v6.0.10
+      with:
+        run_install: false
+
+    - name: Set up Node.js
+      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+      with:
+        node-version: 24
+        cache: pnpm
+
+    - name: Install dependencies
+      shell: bash
+      run: pnpm install --frozen-lockfile
+
+    - name: Resolve Playwright version
+      id: playwright-version
+      shell: bash
+      run: echo "version=$(node -p "require('@playwright/test/package.json').version")" >> "$GITHUB_OUTPUT"
+
+    - name: Cache Playwright browsers
+      uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+      with:
+        path: ~/.cache/ms-playwright
+        key: playwright-${{ runner.os }}-${{ inputs.browser }}-${{ steps.playwright-version.outputs.version }}
+
+    - name: Install ${{ inputs.browser }}
+      # On a cache hit the download is a no-op; --with-deps still installs the
+      # OS packages, which are not cacheable.
+      shell: bash
+      run: pnpm exec playwright install --with-deps ${{ inputs.browser }}
+
+    - name: Run ${{ inputs.project }} shard ${{ inputs.shard }}/${{ inputs.total-shards }}
+      # --fully-parallel changes only the SHARDING granularity: with the
+      # config's fullyParallel:false the whole single-file suite is one
+      # indivisible group and every test lands in shard 1. Pairing it with
+      # --workers=1 keeps execution inside each shard strictly serial in file
+      # order, so tests never run concurrently against the shard's servers —
+      # the property fullyParallel:false exists to protect.
+      shell: bash
+      run: >-
+        pnpm exec playwright test -c playwright.config.ts
+        --project=${{ inputs.project }}
+        --fully-parallel --workers=1
+        --shard=${{ inputs.shard }}/${{ inputs.total-shards }}
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -43,10 +43,69 @@ jobs:
       - name: Run complete Node regression suite
         run: pnpm regression
 
-  browser:
-    name: ${{ matrix.project }}
+  # The browser suite boots the real server and client, so almost any code
+  # change can affect it — the skip list is deliberately narrow (#5622): only
+  # files the running app can never load. This is an in-workflow gate, not a
+  # workflow-level paths filter, so the fan-in jobs below still report green
+  # for branch protection instead of hanging as "Expected".
+  changes:
+    name: browser-lane-gate
     runs-on: ubuntu-latest
-    timeout-minutes: 45
+    timeout-minutes: 10
+
+    permissions:
+      contents: read
+
+    outputs:
+      browser: ${{ steps.decide.outputs.browser }}
+
+    steps:
+      - name: Checkout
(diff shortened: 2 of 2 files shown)


kolacheee:
For the record on the one red lane in the first run: `mobile-webkit 1/3` failed on **`page.reload: WebKit encountered an internal error`** in "mobile transcript swipes navigate Conversation and Roleplay alternatives" — a WebKit engine crash, not a test-logic or shard-ordering failure, and that test was in the old #5589 rotation. The fan-in failing 2s later was the new plumbing doing its job. Rerunning only the failed job went green, and that's a side benefit worth naming: pre-sharding, this flake would have cost a full ~20-minute lane rerun — now it cost one ~7-minute shard. All 9 shards + 3 fan-ins are green as of the rerun. If this test's engine crash recurs, it deserves its own #5618-style issue (it can likely be stabilized by avoiding the reload-under-animation pattern, though an engine crash can never be fully fixed test-side).

PR #5674, opened 2026-08-30
fix(agents): make Beholder's bare and missing flags manual-only

GetBeholder:
# Pull request

## Linked issue

Closes #

## Why this change

- The extractor may propose `missing` and `bare`; Beholder should not apply either on its own. Both are destructive when wrong — `missing` takes over a slot outright and cascades distally, `bare` contradicts whatever is worn there.
- The model is not reliable at them. Measured against the held-out eval set (85 per-character rows, frozen scorer):

| flag | tp | fp | fn |
|---|---|---|---|
| `bare` | 3 | **5** | 12 |
| `missing` | 0 | 0 | 1 |

  `bare` is more wrong than right, and `missing` was never emitted at all — so this costs three true positives and saves five false ones.
- The reference extractor has refused model-set `missing` for the same reason since its first release, and now refuses `bare` alongside it. This brings the agent in line.

## What changed

- `beholder-state.ts`: both flags are stripped in `resolveBeholderStateResponse`, the single point where a model reply becomes state, so the delta path and the full-snapshot path are covered together. Nothing else in the reply is touched — `worn_remove` in particular still works, and there is a test for it.
- A delta whose entire content was those flags now resolves as a **valid no-op** instead of `"empty or unusable state delta"`. The agent refused it on purpose; surfacing that as an extraction failure would hand the operator an error they can neither act on nor reproduce.
- Nothing outside the Beholder agent's own files is touched.

- Affected package IDs: `beholder` (agent behaviour only; no manifest or catalog change)
- Engine compatibility impact: none — no schema, route or contract change
- Restart, storage, update, or uninstall impact: none. Existing stored state is untouched; only newly proposed flags are refused.

## Validation

- [x] `node ./scripts/run-regressions.mjs --filter scripts/regressions/beholder-manual-only-flags.regression.ts` passes
- [x] All four pre-existing `beholder-*` regressions still pass
- [x] `tsc --noEmit -p packages/server/tsconfig.json` clean for the changed file

### Manual verification notes

The two unticked boxes are honest: this is a server-side behaviour change with no package payload, so there is nothing to install, and I verified it through the regression harness rather than a live chat.

New regression covers: a delta proposing `missing` does not take the slot; a delta proposing `bare` does not take the slot and does not disturb what is worn there; the flags are dropped while wounds and held items in the same slot survive; a full-snapshot reply is stripped the same way; and `worn_remove` still removes the named garment.

## Documentation impact

None — no UI change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Beholder state processing by ignoring unsupported `missing` and `bare` flags in generated responses.
  * Responses containing only these flags are now handled as valid no-ops, while other invalid responses remain errors.
  * Existing manually authored state and garment-removal behavior remain unaffected.
  * Reported worn items and wounds continue to be preserved and merged correctly.

* **Tests**
  * Added regression coverage for full snapshots, deltas, wound merging, and garment removal.


GetBeholder:
Holding as a draft. Making these flags manual-only is only correct where there IS a manual path to set them: the reference extension has a slot editor, the Marinara package does not yet. Merging as-is would turn `bare` and `missing` from sometimes-wrong into unsettable. This lands after the slot editor is ported into the package.

GetBeholder:
Fixed in 86fadfa, and you were right that this was the important one.

The full-snapshot branch stripped `missing` and `bare` and then replaced `prior` wholesale, so the operator's flags were erased on the next snapshot. That undoes every correction — the exact outcome this change exists to prevent, so the PR was defeating itself on the path most installs actually take.

Both cases you named are handled: a snapshot that omits the flagged slots keeps them, and a snapshot that contradicts them does not get to overrule them, because a flag the model is not allowed to set is not a flag it is allowed to clear. Everything else in the snapshot still wins.

Regressions added for both, and I confirmed each fails against the previous code.

GetBeholder:
Both fixed.

**Case-insensitive identity (Minor).** Correct — delta resolution matches through `sameCharacterName` and the carry-forward keyed on the exact string, so a snapshot that merely recased a name dropped its flags silently. It matches the same way now.

**Normalization fixed point (Major).** This was the better catch. Carrying `missing` onto a snapshot slot that still had `worn` returned a state that this function's own prior-normalization would change on the next turn — the garment would disappear a turn late, which reads as a bug rather than as the flag working. The carried state is re-normalized, so a missing limb is not left wearing a sleeve, and the result now survives a round trip through prior-normalization unchanged (asserted).

Worth noting: the contradiction case I had added earlier asserted the *opposite* of this — that the snapshot's garment survived on a slot flagged missing. That assertion was wrong and is corrected. The regression now checks the flag wins, the garment on the missing slot goes with it, and slots the flags do not touch still take the snapshot's value.

PR #5676, opened 2026-08-30
fix(agents): recover the removal a compound take-off loses

GetBeholder:
# Pull request

## Linked issue

Closes #

## Why this change

When one sentence both takes a garment off and puts another on, the extractor reports the addition and drops the removal. Measured against the held-out set:

| prose | removal emitted |
|---|---|
| take-off alone — *"hangs the cloak on a hook"* | **3/3** |
| compound — *"hangs the cloak on a hook **and** pulls on black boots"* | **2/8** |

The garment it failed to remove stays in state and is fed back into every later turn, so one miss compounds for the rest of the scene. This reproduces on ordinary prose, not just constructed cases — a six-turn scene hit it on *"trades the ruined sweater for a clean linen shirt… and pulls the coat back on"*.

## What changed

- `beholder-state.ts` gains three pure helpers: `beholderTakeoffClause` (split a compound sentence down to the take-off half), `beholderDeltaLacksRemoval`, and `mergeBeholderWornRemovals`.
- `agent-executor.ts` runs one extra worn-lane call after the five passes **only** when the reply carried no removal, the prose shows something coming off, and the sentence is compound. Only `worn_remove` is merged from the answer — it never adds or replaces a garment. A failed repair is logged and dropped rather than losing the turn.
- `buildBeholderMessages` takes an optional narration override; the narration derivation moved to `beholderNarration` so both callers share it.

**The route matters.** Re-asking the *same compound sentence* with removal-only framing recovers **0 of 6** — it is the compound sentence itself that blinds the model, not the framing. That is why this splits the input rather than rewording the prompt. Five prompt variants were measured first (an explicit rule, a worked example, clause-splitting instructions, a terse imperative, and reordering the existing bullet); every one left the number at 2/8.

- Affected package IDs: `beholder` (agent behaviour only; no manifest or catalog change)
- Engine compatibility impact: none — no schema, route or contract change
- Restart, storage, update, or uninstall impact: none. Cost is one extra provider call on turns that show a garment coming off and reported no removal; ordinary turns are unaffected.

## Validation

- [x] `scripts/regressions/beholder-takeoff-repair.regression.ts` passes
- [x] All pre-existing `beholder-*` regressions still pass
- [x] `tsc --noEmit -p packages/server/tsconfig.json` clean

### Manual verification notes

Measured end to end through these helpers against the trained extractor (koboldcpp, Beholder-Q8_0):

| group | before | after | extra calls |
|---|---|---|---|
| compound | 2/8 | **8/8** | 5 |
| removal only | 3/3 | 3/3 | 0 |
| addition only | 4/4 | 4/4 | 0 |

The two unticked boxes are honest: this is server-side behaviour with no package payload, so there is nothing to install, and it was verified through the regression harness and a live-model run rather than a chat session.

The unit regression pins what decides *when* it fires and *what* it asks — firing too eagerly costs a call on every turn, not firing costs the removal — plus that the merge takes only `worn_remove` and never duplicates.

## Documentation impact

None — no UI change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved clothing removal detection for compound narration.
  * Corrected cases where worn-item removals were missed during scene updates.
  * Prevented duplicate removals and unrelated item changes from being applied.
  * Improved handling of take-off actions spanning multiple clauses or sentences.

* **Tests**
  * Added regression coverage for compound take-off actions, subject preservation, no-op scenarios, and repaired removal results.

### DIFF
--- a/packages/server/src/services/agents/agent-executor.ts
+++ b/packages/server/src/services/agents/agent-executor.ts
@@ -1088,6 +1088,15 @@ async function executeBeholderLanePasses(args: {
       const repairMessages = prepareAgentProviderMessages(
         buildBeholderMessages(config, lanePrompts.worn, context, takeoffClause),
       );
+      // Through the debug path like every other provider call. This one is easy to
+      // miss precisely because it is conditional, and it is the call you most want to
+      // see when a removal did not come back.
+      emitAgentDebug(context, {
+        stage: "request",
+        ...agentDebugBase(config, model, temperature, maxTokens),
+        messageCount: repairMessages.length,
+        messages: debugMessages(repairMessages),
+      });
       const repair = await provider.chatComplete(repairMessages, {
         model,
         temperature,
--- a/packages/server/src/services/agents/agent-executor.ts
+++ b/packages/server/src/services/agents/agent-executor.ts
@@ -47,6 +47,9 @@ import { normalizeCyoaChoiceOutput } from "./cyoa-choice-normalization.js";
 import { getAssetManifest } from "../game/asset-manifest.service.js";
 import { normalizeBeholderProse } from "./beholder-normalizer.js";
 import {
+  beholderDeltaLacksRemoval,
+  beholderTakeoffClause,
+  mergeBeholderWornRemovals,
   BEHOLDER_PASS_LANES,
   buildBeholderUserMessage,
   formatBeholderRequestContext,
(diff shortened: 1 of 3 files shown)


GetBeholder:
Both findings were real and are fixed in 5eba638.

**Wrong subject (Major).** Confirmed with your exact example. The narration handed to the lane can span several messages, and the subject was taken from the first clause of all of it — so `Tim waits.\nMaggie ties a scarf and takes off her boots.` produced `Tim takes off her boots.`, and the repair removed Tim's garment and persisted it. That is worse than the missing removal the repair exists to recover.

The clause is now scoped to the sentence that actually shows the take-off, most recent one wins. Regression covers the newline- and sentence-separated forms; I checked both fail against the previous code and pass against the fix.

**Debug path (Minor).** Agreed — it is the call you most want to read when a removal did not come back, and it was the one call you could not see. It now emits a `request` event through `emitAgentDebug` like every other provider call.

PR #5677, opened 2026-08-30
feat(agents): let an operator correct Beholder's physical state

GetBeholder:
# Pull request

## Linked issue

Closes #

## Why this change

Beholder shows what the extractor believes about each character. When it is wrong there is currently no way to fix it.

The state a chat carries forward is the `resultData` of the agent's last successful run, and `loadPriorBeholderState` builds the next prompt from that same record. So a correction has to land there or it means nothing — anything stored elsewhere would look right in the panel and then be narrated away on the following turn.

The existing `PATCH /runs/:runId` cannot serve this. Both the listing and the patch reach runs through an inner join on `agentConfigs`, which a package agent's runs do not satisfy, so the client cannot discover the run id in the first place. Verified against a running Engine: `GET /runs/:chatId/custom` returns zero rows for a chat with Beholder state.

## What changed

One route, beside the `GET` that already reads this state:

- `PUT /api/agents/beholder-state/:chatId` — normalizes the submitted state, refuses it with 400 if it does not survive normalization, 404s when there is no run to correct yet, and otherwise updates the last successful Beholder run's `resultData`.

Nothing else is touched, and nothing outside the Beholder agent's own area of the routes file.

- Affected package IDs: `beholder`
- Engine compatibility impact: additive route; no schema or contract change
- Restart, storage, update, or uninstall impact: none. It writes the same record the agent already writes, through the same storage call the existing run-patch route uses.

The body is normalized with `normalizeBeholderState` before storage, so a malformed or hostile payload is rejected rather than written into the state the prompt is built from.

## Validation

- [x] `tsc --noEmit -p packages/server/tsconfig.json` clean
- [x] Exercised against a running Engine: `PUT` returns 200 and the following `GET` reflects the change
- [x] Driven end to end from the package UI — 29/29 UI checks pass, including "the edit is written to the stored state" and the edited garment then appearing in the panel

### Manual verification notes

Tested by pointing an isolated Engine (own port, own DATA_DIR) at a build of this branch with the Beholder package installed, then driving the panel in a browser: open a slot, edit it, Apply, and confirm both that `GET /beholder-state/:chatId` returns the edit and that the panel redraws with it. Also confirmed the 400 path by submitting a state that does not normalize.

## Documentation impact

The consuming UI ships in the Beholder package, not the Engine; screenshots belong with that change.

## Summary by CodeRabbit

* **New Features**
  * Added a privileged endpoint for operators to update a conversation’s Beholder state.
  * Validates and normalizes submitted state before applying changes.
  * Requires a latest successful Beholder run and returns the corrected state with run metadata.
  * Applies rate limiting to state updates.
  * Provides clear errors for invalid state, missing runs, or conflicting updates.

### DIFF
--- a/packages/server/src/middleware/rate-limit.ts
+++ b/packages/server/src/middleware/rate-limit.ts
@@ -39,6 +39,17 @@ export const ADMIN_RESTART_RATE_LIMIT = {
   timeWindow: 60_000,
 } as const satisfies MarinaraRouteRateLimit;
 
+/**
+ * Operator corrections to Beholder's physical state.
+ *
+ * A person fixing slots clicks Apply a handful of times a minute; this is far above
+ * that and still bounds an authorized write to the state the next prompt is built from.
+ */
+export const BEHOLDER_STATE_RATE_LIMIT = {
+  max: 60,
+  timeWindow: 60_000,
+} as const satisfies MarinaraRouteRateLimit;
+
 export const BACKUP_RATE_LIMIT = {
   max: 60,
   timeWindow: 60_000,
--- a/packages/server/src/routes/agents.routes.ts
+++ b/packages/server/src/routes/agents.routes.ts
@@ -24,6 +24,7 @@ import {
   type CustomAgentCapability,
 } from "@marinara-engine/shared";
 import { requirePrivilegedAccess } from "../middleware/privileged-gate.js";
+import { BEHOLDER_STATE_RATE_LIMIT } from "../middleware/rate-limit.js";
 import {
   getCustomAgentImportPolicy,
   setCustomAgentImportsEnabled,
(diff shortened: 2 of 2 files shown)


GetBeholder:
Both findings were real and are fixed in a7a6a03.

**Authorization (Major).** The write now takes `requirePrivilegedAccess`, the same guard this file already applies to Custom Agent imports. Loopback, the IP allowlist, trusted interfaces and basic-auth'd browsers all still satisfy it, so the editor is unaffected in normal setups.

I deliberately did **not** add it to the `GET`. That route predates this PR, and gating an existing read would change behaviour for current consumers rather than adding to them. Worth doing, but as its own change rather than smuggled into a write-path PR — flagging it here so it is not lost.

**Lost run (Minor).** `updateRunResultData` returns null when the run is deleted between the lookup and the write, and the handler returned 200 regardless — telling the operator their correction was saved when it had been discarded. That is now a 409.

PR #5678, opened 2026-08-30
feat(sidecar): add a second, independent local model slot for agents

GetBeholder:
## Linked issue

Closes #

## Why this change

Some agents are built around a small purpose-trained model rather than a general one — an extractor whose output is graded against a schema, where a 0.8B model tuned for the job beats a large general one and costs nothing to run.

There is nowhere to put such a model. The sidecar holds exactly one, and that one is the model the operator chose for generation, scene analysis and trackers. Installing an agent's model into it displaces their choice, so in practice those agents have no local path at all and must be pointed at a paid connection.

## What changed

A second, parallel slot. Everything in it is additive; nothing about the existing sidecar changes.

**Isolation.** Its own config file, its own model directory under `data/models/utility`, its own `llama-server` on its own dynamically allocated port. The one thing it shares is the installed llama.cpp runtime, read-only: it uses `sidecarRuntimeService.getCurrentInstall()` and refuses to start when the runtime is absent, rather than `ensureInstalled()`, which would download and rewrite shared state the main slot depends on. It never reads or writes `sidecar-config.json`, never touches models outside its own subdirectory, and never stops or restarts the main process.

**Routing.** When the slot is up and holding a model whose id matches an agent, it answers for that agent; otherwise the agent's configured connection is used exactly as before. The rule is one pure function (`utilitySlotServesAgent`) shared by both agent resolution paths rather than duplicated in each. A configured-but-down slot falls back rather than failing the run.

**Saying which model answered.** `GET /api/utility-sidecar/routing/:agentType` returns the verdict *and the reason for it*, so a UI can name the model that will actually run instead of inferring it from the connection list. This matters more than it looks: an agent answered by the wrong model fails in ways that read as a bad model rather than a wrong setting.

**Explicit selection.** The agent editor offers the slot as a connection, but only once a model for that agent is installed — an option that cannot answer is worse than no option. Chosen-but-not-serving is reported as unavailable rather than quietly answered by a different model, because the two need different prompts.

**Updates by version, not size.** A requantization can land on the same byte count, and the operator is being asked to spend a download, so comparison is by blob id. When the ids cannot be compared the answer is `indeterminate` rather than an implied "current".

**Settings expose hardware only** — context size, GPU layers, parallel slots, each clamped before it reaches `llama-server`. Sampling is deliberately absent: an extractor is graded against a schema, and a temperature dial on it turns a working install into a subtly broken one. Defaults are CPU-only so a first run cannot quietly contend with the main sidecar for VRAM.

All writes sit behind `requirePrivilegedAccess`.

## Validation

- [x] `scripts/regressions/utility-sidecar.regression.ts` added and passing
- [x] Full node regression suite: 166/168. The two failures (`extension-security`, `mari/db-execution-security`) reproduce identically on the unmodified base commit and are environmental
- [x] `tsc --noEmit` clean for `packages/server` and `packages/client`
- [x] `prettier --check "packages/**/*.{ts,tsx}"` clean
- [x] Ran a real engine against an isolated `DATA_DIR` and drove the whole feature end to end (below)
- [ ] Client UI clicked through in a browser (the agent-editor connection option)

### Manual verification notes

Driven against a real engine on an isolated `DATA_DIR`, seeded with a main sidecar holding a *different* model so displacement would show up:

- **Install** — real download from HuggingFace through the route: 811 MB, blob id recorded, landed only at `data/models/utility/beholder/`.
- **Input validation over HTTP** — malformed repo, `../../escape.gguf`, an absolute path, and a non-`.gguf` file each rejected with 400, and nothing was written outside the utility directory.
- **Update check** — against the live repo: ids matched, `updateAvailable: false`, `indeterminate: false`.
- **Start** — own `llama-server` on its own allocated port, launched with `-ngl 0`, so it does not contend with the main sidecar for VRAM by default.
- **A real Beholder agent run through `/api/generate/retry-agents`** — all five lane passes answered by the local model. The debug stream shows `utility-sidecar` as the only model used.
- **Cold start** — engine restarted with the slot selected but stopped; the next agent run started it on demand and again used only the local model.
- **The invariant** — the main sidecar's `sidecar-config.json`, its model file and its runtime record are byte-for-byte identical (sha256) after install, start, an agent run and a restart.

Three faults were found this way and are fixed in the second commit: the slot did not survive a restart (routing required an already-running process, so a restart silently returned the agent to its paid connection); the run warned that it was billing the default gpt-5.5 connection while the local model was actually answering; and the main generation path's built-in agent loop had no override at all. The first two are exactly the kind of thing the code reads correctly and the running system does not.

## Docs and release impact

- [x] No docs changes needed

Server-side feature; the client change is one conditional `<option>` in the agent editor, shown only when a model for that agent is installed.

## Summary by CodeRabbit

* **New Features**
  * Added an independently managed utility sidecar for local agent models.
  * Added model installation, removal, activation, update checks, hardware settings, and start/stop controls.
  * Eligible agents can automatically use the utility sidecar when a compatible model is available.
  * Added utility-sidecar status and routing information.
* **Improvements**
  * The agent editor displays eligible utility models and clearly labels unavailable selections.
  * Improved isolation between utility-sidecar and main local-sidecar settings and models.
  * Added installation progress, model validation, safeguards, and request limits.


GetBeholder:
Five findings, all real, fixed in e9e3eb2.

**Model id normalization (Major).** This was the bad one. `modelFilePath` normalized invalid ids instead of rejecting them, so `a/b` and `a_b` collapsed onto one directory and an empty id resolved to the utility root itself — `removeModel("")` would have deleted the entire utility directory. Ids are validated and refused now. The regression drops a canary file in the utility directory and requires it to survive every rejected id, since as you noted the old test accepted any rejection and would not have caught this.

**Process out of step with the active model (Major).** Correct on both counts. The child now records which model it loaded; `ensureRunning()` treats a mismatch as not-ready and stops it first, `setActiveModel` stops a process serving something else, and `removeModel` stops the active one before deleting it. Related: `setActiveModel("")` was slipping through as "clear the selection" — only `null` clears it now.

**Connection id not propagated (Minor).** This one mattered more than minor: reporting a paid connection for a run the local slot answered is the exact confusion this feature exists to remove. Both paths now report the connection that actually answered.

**Unavailable saved override (Minor).** Agreed. The option now stays when it is the saved selection and is labelled unavailable, rather than vanishing and leaving the select with no match.

GetBeholder:
Both fixed in the latest push — good catches, and both were load-bearing.

**stop() (Major).** Correct: `child.killed` records that the signal was sent, not that llama-server exited, so the SIGKILL fallback could never fire and `stop()` could return while the model file was still open. That is exactly the moment `removeModel` goes on to delete the directory. It now waits on the exit event, escalates to SIGKILL after a timeout, and waits again.

**ensureRunning() (Major).** Also correct. The identity check happened before joining an in-flight start, so a caller waiting on A's startup and then finding B selected got A's ready port while status reported B. The check is re-made after the wait now, and a mismatch starts the right model instead of returning the wrong one. Bounded so a flip-flopping selection cannot spin.

GetBeholder:
Both fixed in d2697f0.

**Concurrent start during shutdown (Major).** Right, and this is the one that bites hardest: `stop()` cleared `this.child` before the exit wait finished, so a concurrent `ensureRunning()` saw no child and started a second `llama-server` while the first was still exiting. Two processes competing for the same GPU and model slot is a worse failure than the one being avoided — on a box near its memory ceiling it is what takes something else down with it. Shutdown is a shared promise now: `stop()` joins an in-progress one, and `ensureRunning()` waits for it before it looks at the child or starts anything.

**Leaked timer (Minor).** Cleared.

PR #5703, opened 2026-08-31
fix(game): stop claiming the Game Master is writing after the tokens stop

### LINKED ISSUE #5692: [Issue]: Game Mode stays in writing state during idle gap before scene analysis
$## Summary

Game Mode can spend about 5-10 seconds with low CPU and GPU use after the main LLM finishes generating text but before sidecar scene analysis starts. During this interval, the status bar still says "The Game Master is writing the next segment...".

The delay is inconsistent in frequency and duration. With fast LLMs, it can represent most of the total turn latency.

## Expected behavior

After the main Game Master response is complete, Marinara should either begin scene analysis without an avoidable idle gap or show the actual processing phase. The status text should not say that the Game Master is writing when no main text generation is running.

## Actual behavior

The generated text is not immediately presented as the completed Game Mode narration. The UI continues to show "The Game Master is writing the next segment..." while CPU and GPU remain largely unused. Sidecar analysis starts later.

## Steps to reproduce

1. Configure a Game Mode chat with a fast main LLM and enable sidecar scene analysis.
2. Send a player action.
3. Observe the interval after the main model stops producing tokens and before the scene analysis request begins.
4. Compare the status text and system CPU/GPU activity during this interval.

The interval does not occur with the same duration on every turn.

## Environment

- Marinara Engine version: 2.4.4
- Install type: installer or other Windows installation (not specified in the report)
- OS + version: Windows 10.0.26100 (x64)
- Browser or app shell: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36

No logs, screenshots, or video were provided.

## Additional context

Reported diagnostics:

- GPU: ANGLE (Microsoft, Microsoft Basic Render Driver (0x0000008C) Direct3D11 vs_5_0 ps_5_0, D3D11)
- Active connection: G4 31B StyleTune
- Connection provider: custom
- LLM model: G4-MeroMero-v2-31B-Q5_K_M.gguf
- Server memory: heap 345.9 / 4288 MiB; RSS 543 MiB

Code inspection indicates that the server saves the assistant message before post-processing agents finish, but the client keeps Game Mode `isStreaming` active until the full generation request completes. The client then refreshes messages and starts scene processing. This may combine post-processing, refresh work, and scene-analysis startup into a visible idle interval. The Microsoft Basic Render Driver may affect local rendering or sidecar performance, but it does not explain the inaccurate writing status by itself.

Gunterlie:
Closes #5692

## Why

In Game Mode the status line said "The Game Master is writing the next segment..." for the whole generation request. But the request stays open well past the last narration token: the server saves the assistant message, then post-processing agents run, the client refreshes messages, and only then does sidecar scene analysis start. With a fast local model that gap is most of the turn, so users watched an idle CPU while the UI insisted text was being generated.

Unlike Roleplay, Game Mode cannot release `isStreaming` on `assistant_message_ready` — the segment-by-segment narration reveal depends on the stream staying live — so `isStreaming` alone cannot tell "the model is writing" apart from "the model is done and the turn is being assembled".

## What

The server already sends the exact signal: `message_saved` fires the moment the narration row is durable. Game generations now record that in a per-chat `narrationSavedChatIds` flag on the chat store, and the status line reads it — writing label before, the existing "Preparing scene…" label after.

The flag follows the same lifecycle as the neighbouring `backgroundIllustrationChatIds`: cleared when a new generation takes the chat's abort controller, and cleared by `clearPerChatState` on teardown.

No new locale keys.

## Scope note

This makes the status truthful. It does not shorten the gap itself — reducing the post-generation latency means reordering the server-side post-processing and scene-analysis handoff, which is a separate change.

## Test plan

- [ ] `pnpm check`
- [ ] `node scripts/run-regressions.mjs` (adds `game-narration-phase`, pinning that every teardown path clears the flag — a missed clear would strand the next turn on "Preparing scene…")
- [ ] Manually run a Game Mode turn with a fast model and scene analysis enabled; verify the label switches to "Preparing scene…" the moment the narration completes
- [ ] Manually verify the label stays "The Game Master is writing the next segment..." while tokens are still arriving
- [ ] Manually verify with streaming disabled that the writing label shows until the message is saved
- [ ] Manually run two turns in a row and verify the second turn starts on the writing label
- [ ] Manually verify the narration segment reveal and the input bar timing are unchanged

## Summary by CodeRabbit

* **New Features**
  * Added clearer game narration status updates while the Game Master prepares and writes the next segment.
  * Narration now indicates when text has been saved, even while additional game processing continues.

* **Bug Fixes**
  * Improved per-chat narration status tracking and cleanup when starting or ending generation.

* **Tests**
  * Added regression coverage for narration status transitions, chat isolation, and reset behavior.


Gunterlie:
Both findings applied in `ce0f2b1` (see latest commit).

1. **Group turns** — confirmed valid. `group_turn` fires once per character and each character's message is saved, so the first `message_saved` would have left the status reading "Preparing scene…" while the next character was still narrating. The handler now clears the flag at the start of each group turn.
2. **Superseded generations** — applied, matching the ownership guard `assistant_message_ready` already uses: the flag is only set while `abortControllers.get(chatId)` is still this request's controller.

`pnpm check` and the `game-narration-phase` regression pass.

Gunterlie:
Applied. The `group_turn` reset now carries the same `abortControllers.get(params.chatId) === abortController` guard as the `message_saved` set — guarding one and not the other was an asymmetry, thanks.

On the requested regression: the tracked `game-narration-phase` regression pins the store contract (set, per-chat scoping, and every clear path). Covering *queued SSE event ordering across two overlapping generations* would need a fake SSE transport and a driven `useGenerate` mutation, which the regression lane has no harness for today. I have left that uncovered rather than add a bespoke harness in this PR — flagging it explicitly rather than claiming coverage.

PR #5757, opened 2026-09-02
fix(server): distinguish staged workspace changes

### LINKED ISSUE #5756: Staged sensitive file changes count as applied mutations, so a read of the unchanged file can satisfy the verification guard
**In one sentence:** the workspace verification guard classifies every successful non-app_data mutation as *applied*, but `write`/`edit` on a sensitive file can succeed while only **staging** the change behind an approval — so a follow-up read of the (unchanged) file satisfies the guard and Mari may claim completion while the approval is still pending.

## The gap (flagged by CodeRabbit on #5755, pre-existing)

`isAppliedWorkspaceMutation` (workspace-agent.service.ts) treats any successful mutating command that isn't `app_data` as an applied change; only `app_data` results are additionally gated on `"saved": true`. But the sensitive-file path (`workspace-change-review.service.ts`) lets `commandWrite`/`commandEdit` return success with an output saying the file was *not* changed and an approval is pending. A subsequent successful read then flips the verification state to `verified`, and the completion-claim interception won't fire — even though nothing was applied.

Mitigations that keep this from being acute today: the read in that scenario returns the *old* content, so a model actually comparing what it reads against what it wrote sees the discrepancy; and the pending approval is prominently surfaced to the user. But the guard's own signal is wrong, and the #5754-era coaching ("a successful read after the write satisfies verification") leans on it harder.

## Possible directions (maintainer's call)

- Return an explicit applied-vs-staged outcome from mutation execution (structured on the command result, like the store read-back sentinel pattern from the #5754 follow-up) and require verification only for mutations that actually applied — a staged mutation would count as neither applied nor verified, and a completion claim would be intercepted with "the change is staged behind an approval; say so instead of claiming completion".
- Minimal version: gate `isAppliedWorkspaceMutation` for `write`/`edit` on the absence of the engine-written staging marker in the output (the staging text is engine-authored, so an anchored check is available).
- Coaching-only version: teach Mari that a staged sensitive change must be reported as pending, never as done — weakest, since the whole point of the guard is not trusting the model on this.

Related: the #5754 follow-up (store read-back for app_data mutations) established the pattern of engine-verified outcomes carried structurally on results — this is the same idea applied to the staging path.

mikemikimike:
Fixes #5756

## Why this change
Sensitive `write` and `edit` operations can return success while only staging a change for user approval. The workspace verification guard treated those results as applied mutations, so a later read of the unchanged file could incorrectly satisfy the guard and let the agent claim the change was applied.

## What changed
- recognize the sensitive-file approval response for `write` and `edit` as unapplied
- add regression coverage for staged `write` and `edit` results, while preserving ordinary applied-write behavior

## Validation
- [x] `node ./scripts/run-regressions.mjs --filter scripts/regressions/prompt.regression.ts` (1/1 passed)
- [x] `pnpm --filter @marinara-engine/server lint` (passed)
- [x] `pnpm build` (passed; existing bundle-size/circular-chunk warnings only)
- [x] `git diff --check` (passed)
- [ ] `pnpm regression:node` (153/158 passed; 5 existing Windows/baseline regressions)
- [ ] Prettier check on the two changed files (blocked by the repository's existing CRLF/full-file formatting baseline; added sections match Prettier output)

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive workspace write and edit actions awaiting user approval are no longer reported as completed changes.
  * Pending changes no longer trigger mutation verification or completion tracking.
  * Successfully applied changes continue to be tracked separately from staged, unapplied actions.

* **Tests**
  * Added regression coverage for staged sensitive file changes and genuinely applied workspace updates.


kolacheee:
Thank you for this fix - it correctly closes the staged-vs-applied gap from #5756, and the regression cases were a welcome touch.

Because this repository's required "Owner approval for outside contributors" check can only be satisfied by @SpicyMarinara (away for the next two weeks), and we didn't want the bug live in the meantime, your fix has been adopted into PR #5775 with your commit and authorship preserved exactly as you wrote it - you keep full credit in the history and contributors graph. That PR adds hardening on top (position-zero anchoring of the staging marker, plus fixes for staged/applied mixed rounds that an adversarial review surfaced) and a CHANGELOG entry crediting you.

Closing this PR in favor of that one. Thanks again for contributing - sorry about the gate timing, and we hope to see more from you.

PR #5771, opened 2026-09-02
fix(connections): read NanoGPT model limits from its detailed catalog

### LINKED ISSUE #5770: [Issue]: NanoGPT model list reports Context 0 / Max Output 0
## Summary

NanoGPT connections show `Context: 0` and `Max Output: 0` for most models, so the Max Context Window field is never auto-filled and has to be set by hand. Marinara requests NanoGPT's plain model catalog, which does not carry any limit fields.

## Expected behavior

Selecting a NanoGPT model fills in that model's context window and output limit, the same way it works for OpenRouter, and Max Context Window is auto-set from the model.

## Actual behavior

The model card shows `Context: 0  Max Output: 0`, Max Context Window stays at whatever was typed manually, and the dropdown label repeats the model id twice because no display name is available.

Only models whose unqualified id also exists in the built-in aggregator fallback (OpenAI, Anthropic, Google, DeepSeek, MiMo, Moonshot, Z.AI, xAI) resolve limits. Any other vendor — for example `inception/mercury-2.5-preview` — resolves to nothing.

## Steps to reproduce

1. Create a NanoGPT text connection and paste an API key.
2. Click **Fetch Models from API**.
3. Select `inception/mercury-2.5-preview` (or any model from a vendor outside the aggregator fallback list).
4. The model card reads `Context: 0  Max Output: 0` and Max Context Window is not auto-set.

## Environment

- Marinara Engine version: 2.4.5 (`staging`)
- Install type: source (git checkout)
- OS + version: Windows 10 22H2
- Browser or app shell: desktop browser

The provider catalog is public, so the difference is reproducible without a key:

`GET https://nano-gpt.com/api/v1/models` — every entry carries only:

    id, object, created, owned_by

`GET https://nano-gpt.com/api/v1/models?detailed=true` — the same models, plus:

    name, description, context_length, max_output_tokens, architecture, capabilities, pricing, ...

For `inception/mercury-2.5-preview` the detailed catalog reports `context_length: 260000` and `max_output_tokens: 65536`.

Measured on the current catalog (622 models): 609 carry a usable context window and 597 a usable output limit in the detailed variant, and none in the plain one.

Cost of the detailed variant, measured with compression as the server requests it:

| Variant | On the wire | Decoded | Request time |
| --- | --- | --- | --- |
| plain | 7.6 KB | 58 KB | 0.66 s |
| detailed | 60 KB | 675 KB | 1.48 s |

## Additional context

`packages/shared/src/constants/model-lists.ts` states the intent explicitly — `nanogpt: [], // NanoGPT aggregator — models fetched dynamically via API` — but the request Marinara sends is the plain catalog, which carries no limits to fetch. The normalizer in `connections.routes.ts` already reads `context_length` and `max_output_tokens`, so no parsing change is needed.

NanoGPT ignores unknown query parameters and answers with the plain catalog, so a future removal of the flag would degrade to today's behavior rather than fail.

## Template check

- [] I DID NOT read this template and provide the requested details.

SonaGat3:
> [!IMPORTANT]

## Linked issue

Closes #5770

## Why this change

- NanoGPT connections report `Context: 0` and `Max Output: 0` for most models, so Max Context Window is never auto-filled and users have to guess the limit by hand. On `inception/mercury-2.5-preview` the real window is 260K, and a user who typed 128K was silently losing half of it.
- `model-lists.ts` already declares the intent — `nanogpt: []` with the comment "models fetched dynamically via API" — but the request Marinara sends is NanoGPT's plain catalog, whose entries carry only `id`, `object`, `created` and `owned_by`. There are no limits in that response to read.
- Today only models whose unqualified id also exists in the built-in aggregator fallback (OpenAI, Anthropic, Google, DeepSeek, MiMo, Moonshot, Z.AI, xAI) resolve limits. Every other vendor resolves to zero, and the only alternative to this fix is adding vendors to a static list by hand, which is what the aggregator design deliberately avoids.

## What changed

- `packages/shared/src/constants/providers.ts`: NanoGPT's `modelsEndpoint` now requests the detailed catalog (`/models?detailed=true`). No other provider is affected.
- No parsing change: `readOpenAICompatibleModelLimits` in `connections.routes.ts` already reads `context_length` and `max_output_tokens`, which are exactly the fields the detailed catalog adds.
- The detailed catalog also supplies `name`, so the model dropdown stops rendering the id twice (`inception/mercury-2.5-preview (inception/mercury-2.5-preview)` becomes `Mercury 2.5 Preview`).
- CHANGELOG entry under `[Unreleased] → Fixed`.

Measured, on the current catalog of 622 models: 609 now resolve a context window and 597 an output limit, against none before.

Notes for reviewers:

- The detailed catalog is larger — 60 KB against 7.6 KB on the wire with compression, 675 KB against 58 KB decoded, and about 0.8 s slower per fetch. Both figures sit well under the 5 MiB cap on the model-list request.
- The same field also feeds the connection test, whose response cap is 2 MiB rather than 5 MiB. The current response is 675 KB, so there is roughly a threefold margin, but the catalog grows over time. If reviewers prefer to remove that exposure, the connection test can keep using the plain catalog since it only checks credentials; that would need a separate field or a special case in the URL builder, so it is deliberately not in this PR.
- Putting a query string in a field named `modelsEndpoint` is new for this file: every other provider uses a bare path. Both consumers concatenate the value onto a base URL with trailing slashes stripped, and both produce a valid URL. Happy to move it to a dedicated field if the convention matters more than the diff size.
- NanoGPT ignores unknown query parameters and answers with the plain catalog, so if the flag ever disappears upstream this degrades to today's behavior instead of breaking the model list.

## Validation

### Manual verification notes

Verified in a desktop browser against a real NanoGPT connection, on this branch:

- **Fetch Models from API** on a NanoGPT text connection loads the catalog as before.
- `inception/mercury-2.5-preview` now reports `Context: 260K` and `Max Output: 65536`, where it previously reported `Context: 0  Max Output: 0`.
- The model dropdown renders `Mercury 2.5 Preview` instead of repeating the id twice.
- **Max Context Window** is auto-filled from the selected model instead of staying at a hand-typed value.
- **Test connection** still succeeds on the same connection, which is the path that also uses this endpoint.
- An untouched provider (OpenRouter) still lists models and limits exactly as before.

`pnpm check` passes locally on this branch (impeccable, localization, format, lint, e2e types, build).

## Docs and release impact

- [ ] No docs changes needed

image
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed NanoGPT model connections so context window and output limits are populated with accurate values.
  * The model picker and automatically filled Max Context Window field no longer display 0 for supported models.
* **Documentation**
  * Added an Unreleased changelog entry describing the NanoGPT connection fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 
 ### Fixed
 
+- NanoGPT connections now read each model's context window and output limit from the provider catalog, so the model picker and the auto-filled Max Context Window show real values instead of 0 for models that are not covered by the built-in aggregator fallback (#5770).
 - Lorebook entries are no longer inserted twice when a preset has more than one lorebook marker (#5716): each world-info position (Before / After) is now placed by the first marker that covers it, so a second "Lorebook Marker (All)" placeholder - or an "All" marker following a "Before" marker - no longer repeats the same entries in the prompt.
 - Mobile connection drag previews now retain the configured Chat Chrome Text Color, and the Roleplay quick switcher context progress bar follows Accent Color (#5758).
 - Professor Mari can no longer answer her own permission question (#5748): once she asks whether to apply a change, that question is binding for the rest of the run - any edit she stages afterwards is held behind the Accept action instead of executing, and a hidden follow-up edit is refused with guidance. Previously an ask that rode alongside an `apply:false` preview left the engine nothing to hold, so she could pivot to applying unasked one round later ("to show you the review card"). Dry-run previews now tell her, truthfully, that the user cannot see them, and "propose your edits" maps to one described-and-held proposal instead of a preview plus a second full generation.
--- a/packages/shared/src/constants/providers.ts
+++ b/packages/shared/src/constants/providers.ts
@@ -136,7 +136,8 @@ export const PROVIDERS: Record<APIProvider, ProviderDefinition> = {
     id: "nanogpt",
     name: "NanoGPT",
     defaultBaseUrl: "https://nano-gpt.com/api/v1",
-    modelsEndpoint: "/models",
+    // Detailed catalog: the plain one omits context_length and max_output_tokens.
+    modelsEndpoint: "/models?detailed=true",
     supportsStreaming: true,
     usesAuthHeader: true,
     apiKeyHeader: null,


SpicyMarinara:
Thank you for your patience. I am back from vacation and can take a look at your PR now.

SpicyMarinara:
Your contribution from this PR was incorporated and adapted in the maintainer-requested consolidated sweep, #6072, now merged into `staging`. The PR description links this source PR, and the implementation commit preserves co-author credit.

Thank you for the patch. Closing this PR as superseded by #6072 so the same work is not merged a second time.

PR #5879, opened 2026-09-05
feat(server): let capability packages register their own file tables

### LINKED ISSUE #5878: [Feature]: Let capability packages register and persist their own file tables

The Slurp capability package needs package-owned storage. Slurp declares its own file tables and cannot persist them.

The file-backed store builds all table metadata once, at module load, from a hardcoded allowlist of the Engine's own tables. Any table name outside that list has no metadata, so `getMeta()` throws:

```
[file-storage] Unsupported table: <name>
```

A downloaded capability package therefore has no supported way to persist its own data. Its only option is to borrow an Engine table that was designed for something else, which means writing package rows into a schema the package does not own, sharing a flush and manifest entry with unrelated Engine data, and having those rows travel in profile export/import as if they were Engine data.

This is an Engine-side gap. The package side already declares tables; the store simply has no entry point to accept them.

## Proposed solution

Add a storage-side API on the live store that registers package-supplied file tables.

Packages activate after the store's `initialize()`, so registration has to work against an already-running store rather than at boot. A registered table should end up with everything a boot-loaded table gets, or it would accept writes and silently lose them: validated metadata, a row container, its existing rows read off disk, membership in the flush loop, and an entry in the snapshot manifest.

Required properties:

- Idempotent, so a package reload re-registering the same table does not reset it.
- Cannot shadow an Engine table; the existing definition wins.
- Failures are per table and logged, not thrown, so one malformed package table definition cannot abort host startup.

Security is the load-bearing part. Table names become filesystem paths (`tables/<name>.json`, `tables/<name>/`) with no escaping anywhere in the store. That was safe only while every name came from the Engine's own hardcoded list. An unvalidated package-supplied name such as `../../.ssh/authorized_keys` would be an arbitrary-write primitive outside the data directory, so a package-supplied name must be validated before it reaches any path join or any registry.

Two further constraints follow from how the store already works:

- Registered tables must be sharded. `SHARDED_TABLES` aliases the file-backed table list, so a name added to that list becomes sharded by array identity whether or not that was intended. The sharded-table lookup set must be kept in step with the list rather than remaining a module-load snapshot, or the load, flush and delete paths would each classify the same table differently.
- A table with no primary key must be refused at registration. Registered tables shard by primary key, so a keyless table would register cleanly and then fail on its first insert with no stable shard key, long after a package author could connect that failure to the table definition.

Profile export/import should deliberately exclude registered tables. A profile snapshot is portable user data that may be restored into a different install, where the owning package may not exist; importing those rows would fail at restore time or leave silent orphans with no schema to validate against. Package data still persists normally and is captured by a full data-directory backup, which is scoped to one install.

## Alternatives considered

- **Add Slurp's tables to the Engine's hardcoded list.** Rejected: it makes the Engine carry schema for an optional downloadable package, and it does not generalise to the next package. It also puts package rows into profile export, which is the wrong portability boundary.
- **Let packages borrow an existing Engine table.** This is the status quo. It stores package rows under a schema the package does not own and gives the package no control over its own data lifecycle.
- **A separate storage mechanism just for packages.** Rejected: it duplicates the crash-safe shard pipeline, the flush scheduling, the manifest and the corruption recovery that the file-backed store already has, and it is the parts that are hard to get right that would be duplicated.
- **Per-store instance registries instead of process-global ones.** This is the cleaner design, but the Engine's table metadata is already process-global and reached through module-level lookups, so instance registries would mean threading a store instance through `getMeta()` and the whole lookup surface. That is a much larger change than the gap warrants while production runs a single store per process.

## Additional context

The Engine currently has no caller that loads package-supplied table objects, so the first change here is the storage-side API only. Wiring a package loader to it is separate work.

## Template check

Gunterlie:
> [!IMPORTANT]
> Contributions target `staging`.

## Linked issue

Closes #5878

## Why this change

The Slurp capability package needs package-owned storage. Slurp declares its own file tables and cannot persist them.

The file-backed store builds every table's metadata once at module load, from a hardcoded allowlist of the Engine's own tables. Any name outside that list has no metadata, so `getMeta()` throws `[file-storage] Unsupported table: <name>`. A downloaded capability package therefore has no supported way to persist its own data; it can only borrow an Engine table that was designed for something else.

This PR closes the Engine-side gap: it adds the storage API that accepts package-declared tables into a live store.

## What changed

**`registerTables` on the store controller.** Packages activate after `initialize()`, so registration runs against an already-running store rather than at boot. A registered table gets everything a boot-loaded table gets, because the alternative is a table that accepts writes and silently loses them: validated metadata, a row container, its existing on-disk rows, membership in the flush loop, and an entry in the snapshot manifest. Registration is idempotent, so a package reload re-registering the same table does not reset it. It cannot shadow a table that already exists — the first definition of a name, always the Engine's own since those are built at module load, keeps the name. Failures are per table and logged rather than thrown, so one malformed package table definition cannot abort host startup.

**Table-name validation.** Table names become filesystem paths (`tables/<name>.json`, `tables/<name>/`) with no escaping anywhere in the store. That was safe only while every name came from the Engine's own hardcoded list; an unvalidated package-supplied name would be an arbitrary-write primitive outside the data directory. `fileBackedTableNameRejection` now gates every registered name and accepts only what the Engine's own names already look like: non-empty, at most 64 characters, matching `/^[a-z][a-z0-9_]*$/`, and not a Windows reserved basename (Windows blocks those with any extension). The length bound keeps `<name>.json` plus a shard filename inside path limits. A rejected name is refused before it reaches any path join or any registry.

The regression asserts this against the real store: it registers a set of hostile names — `../../escaped`, `..`, `nested/table`, `back\slash`, `Uppercase`, `trailing.`, `with space`, `_leading`, `9leading`, `con`, and a 65-character name — and then walks the entire sandbox tree, which is one level above the storage directory so an escape would land somewhere the test can see. It asserts that no rejected name joins the file-backed table list and that not one file or directory was created, inside or outside the data directory.

**Sharding.** Registered tables are sharded, like every Engine table. `SHARDED_TABLES` aliases `FILE_BACKED_TABLES`, so a name pushed onto the list becomes sharded by array identity whether or not that was intended; making registered tables flat would opt a brand-new table into the store's least-exercised path and would still leave the alias lying about them. That aliasing was also a trap worth closing: `SHARDED_TABLE_SET` used to be a module-load snapshot of the list, so after registration the array said "sharded" and the set said "not sharded", and the load, flush and delete paths would each have classified the same table differently. The set is now mutable and registration writes the list and the set together, or neither.

**No primary key is refused at registration.** Registered tables have no `SHARD_KEY_COLUMNS` entry, so they shard by primary key. A keyless table would register cleanly and then throw "no stable shard key" on its first insert, long after a package author could connect the failure to the definition. It is now refused up front, before the name enters any registry.

**Profile export/import.** Registered tables persist exactly like Engine tables and are captured by a full data-directory backup, but they are deliberately excluded from profile export. `profileTableObjects` in `backup.routes.ts` is built from the schema barrel, so a runtime-registered table has no entry and every loop skips it. That exclusion is intentional, not an oversight to fix by consulting the live registry: a profile snapshot is portable user data that may be restored into a different install where the owning package does not exist, so importing those rows would be a restore-time failure or a silent orphan with no schema to validate against. A comment at that site records the reasoning.

**Known constraint — the registries are process-global.** `tableMetasByName`, `FILE_BACKED_TABLE_SET` and `SHARDED_TABLE_SET` are module-level, as the Engine's own table metadata already is. A registered name therefore outlives the store that registered it: a store constructed later in the same process, against a different storage directory, already knows the name, boot-loads it from its own directory, and finds nothing. That degrades to an empty table, never to corruption or a cross-directory read. It is accepted because production runs one store per process, and because the alternative — instance-scoped registries — means threading a store through `getMeta()` and the entire module-level lookup surface. Per-instance registration is the correct fix if the Engine ever runs more than one store at a time. Until then there is a concrete consequence for tests: a test that constructs several stores **must** give each scenario a distinct table name, or its second store silently exercises the boot loader instead of `registerTables`. The regression follows that rule explicitly, with a comment at each case that needs a fresh name.

**No caller is wired yet.** Nothing in the Engine currently loads package-supplied table objects. This PR is the storage-side API only; connecting a package loader to it is separate work.

Also included:

- `FILE_BACKED_TABLES` is now a widened runtime list built from `BUILT_IN_FILE_BACKED_TABLES`, so every consumer that walks "all file-backed tables" sees registered names. The Engine's own tuple keeps its literal-union type so `CASCADES` and other table-name constants stay compile-checked. The list itself is module-private; the only sanctioned way to grow it is `registerTables`, which validates the name and updates every companion registry in the same breath.
- `scripts/regressions/launcher/format-guard.regression.mjs` now compares the launcher's `SHARDED_TABLES` copy against `BUILT_IN_FILE_BACKED_TABLES` rather than the widened list. An offline downgrade script can never know a runtime-registered name.
- `loadRegisteredTableRows` reads a newly registered table's existing shards. It is deliberately thinner than the boot loader: the re-home detection, duplicate arbitration and format-migration repair there exist for histories a table the Engine has never written cannot have. Corrupt-file recovery still applies, because that comes from `parseJsonFile` and applies to any file on disk.
- `buildFileTableMetadata` is extracted so schema-declared and package-registered tables share the same shape and unique-constraint validation.
- New regression `scripts/regressions/package-table-registration.regression.ts` (auto-discovered by the runner).

## Validation

### Manual verification notes

The checkboxes above are left unticked for a human to confirm. This change has no UI surface and no wired caller, so the meaningful verification is the new regression, which drives the real store end to end rather than a mock. It is written to be run as `node ./scripts/run-regressions.mjs --filter package-table-registration` after `pnpm build:shared`, and each case builds a real store over a fresh temporary storage directory:

1. **Hostile names.** Flush twice to a clean baseline (the second flush is what creates the manifest's `.bak`, so normal store bookkeeping stays out of the comparison), snapshot the whole sandbox, register each hostile name plus a keyless table, flush again, and assert the sandbox tree is byte-for-byte unchanged and no rejected name entered `FILE_BACKED_TABLES`.
2. **Register and persist.** Register a valid table, insert a row, re-register the same table and confirm the row survives, flush, and check on disk that a shard directory exists and that no flat `<name>.json` monolith was also created. Then close the store, open a new one over the same directory, register again, and confirm the row is read back with its values intact. Delete the row, flush, and confirm the emptied shard is removed rather than left as litter.
3. **Shadowing.** Insert a real `chats` row, register an impostor table also named `chats` with different columns, and confirm the Engine's rows and column metadata are untouched and the name is not duplicated in the table list.
4. **Recovery on registration.** Plant a corrupt shard primary with a good `.bak`. Registration must recover the rows *and* mark the shard dirty; the test then flushes and reads the file back to confirm it was rewritten canonically, so the next boot does not recover from backup all over again.
5. **Recovery to zero rows.** Plant a corrupt primary with a valid-but-empty `.bak`. There is no shard key to dirty, so the file is marked stale instead; the test confirms the flush removes the primary and its backup rather than leaving a corrupt pair that is re-recovered and re-logged on every boot.
6. **Unrecoverable pair.** Plant a primary and a `.bak` that both fail to parse, and confirm both are quarantined for manual inspection, reported through `getQuarantinedTables()` against the registered table, and left undisturbed by the following flush.

The regression sets `LOG_LEVEL=silent` before importing the store, because these cases deliberately provoke error-level rejections and quarantine reports, and the pretty-print transport worker can otherwise outlive the test and hang the runner.

## Docs and release impact

- [ ] No docs changes needed
- [ ] Translated docs on the `docs-i18n` branch updated to match, or a `[docs-i18n]` follow-up issue opened

A `CHANGELOG.md` entry is added under `[Unreleased]`. No user-facing docs change, because there is no user-facing surface yet. No version bump.

Not applicable; no UI surface.

## Summary by CodeRabbit

* **New Features**
  * Capability packages can register persistent file tables at runtime.
  * Registered tables support sharding, querying, persistence across restarts, and safe backup recovery.
  * Table names are strictly validated to prevent unsafe or conflicting registrations.
  * Re-registering an existing table is handled safely and consistently.

* **Bug Fixes**
  * Improved recovery of damaged or unreadable table data, including quarantining unusable files and rewriting recovered shards canonically.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This file is the release-notes source of truth for Marinara Engine. Reuse these 
 
 ## [Unreleased]
 
+- Added a storage API that lets a downloaded capability package register and persist its own file tables, with strict table-name validation. No caller is wired to it yet.
 - Removed the unfinished Slurp creator-feed material from the Noodle guides.
 - Updated the timeout reference in `.env.example` to use Slurp consistently.
 - Added a reusable Character Schedule Manager for Conversation schedules. It groups characters with and without schedules, supports bulk generation and removal, and provides per-character weekly renewal controls.
--- a/packages/server/src/db/file-backed-store.ts
+++ b/packages/server/src/db/file-backed-store.ts
@@ -215,6 +215,33 @@ export type FileNativeStoreController = {
    * safe, a missed one is not. 0 = never written in this process.
    */
   getTableWriteGeneration: (table: string) => number;
+  /**
+   * Adds capability-package-owned file tables to the live store: host
+   * integration for downloaded packages that declare their own storage.
+   * Packages activate after initialize(), so this integrates
+   * a table into an already-running store rather than at boot. Names that fail
+   * validation, or that a table already claims, are skipped and logged — one bad
+   * package table must never abort host startup.
+   *
+   * Registered tables persist exactly like the Engine's own — they flush, they
+   * appear in the snapshot manifest, and a full data-directory backup captures
+   * them. They are deliberately NOT part of profile export/import; see the note
+   * beside profileTableObjects in backup.routes.ts for why.
+   *
+   * KNOWN CONSTRAINT — the registries are PROCESS-GLOBAL, not per store. A name
+   * registered here stays in tableMetasByName/FILE_BACKED_TABLES for the life of
+   * the process, so a store constructed later against a DIFFERENT storage
+   * directory already knows the name and boot-loads it from its own directory,
+   * where it finds nothing. That degrades to an empty table, never to corruption
+   * or a cross-directory read. It is accepted because production runs one store
+   * per process, and because the alternative — instance registries — means
+   * threading them through getMeta() and the whole module-level lookup surface.
+   * Per-instance registration is the correct fix if the Engine ever runs more
+   * than one store at a time. Until then: a test that constructs several stores
+   * MUST give each scenario a distinct table name, or its second store will
+   * silently exercise the boot loader instead of this method.
+   */
+  registerTables: (tables: readonly unknown[]) => void;
 };
 
 export type FileNativeDB = {
@@ -299,7 +326,7 @@ export const STORAGE_WRITER_LIVENESS_FILENAME = "live.sock";
 const SAVE_DEBOUNCE_MS = 750;
 const SAFETY_SAVE_MS = 10_000;
 
-export const FILE_BACKED_TABLES = [
+const BUILT_IN_FILE_BACKED_TABLES = [
   "chats",
   "messages",
   "message_swipes",
@@ -383,12 +410,28 @@ export const FILE_BACKED_TABLES = [
   "mari_workspace_context",
 ] as const;
 
-type FileBackedTable = (typeof FILE_BACKED_TABLES)[number];
+/**
+ * The Engine's own tables keep their literal-union type so CASCADES and other
+ * table-name constants stay compile-checked. The runtime list is a widened copy
+ * because capability packages register their own tables into a live store
+ * (registerTables): every consumer that walks "all file-backed tables" — the
+ * flush loop, the manifest, the Mari db tooling — must see them, and an
+ * `as const` tuple could never grow.
+ */
+type FileBackedTable = (typeof BUILT_IN_FILE_BACKED_TABLES)[number];
+/**
+ * The registry itself. Module-private because the ONLY sanctioned way to grow
+ * it is registerTables, which validates the name and updates every companion
+ * registry in the same breath; an outside push would land a name in the flush
+ * loop with no metadata and no shard classification.
+ */
+const fileBackedTables: string[] = [...BUILT_IN_FILE_BACKED_TABLES];
+export const FILE_BACKED_TABLES: readonly string[] = fileBackedTables;
 
 // #5302: every file-backed table uses the existing crash-safe shard pipeline.
 // Order remains significant for messages/swipes because swipe ownership is
 // resolved through the parent-message index.
-export const SHARDED_TABLES = FILE_BACKED_TABLES;
+export const SHARDED_TABLES: readonly string[] = FILE_BACKED_TABLES;
 
 /**
  * Child tables group by their stable owner. Every unlisted table uses its
@@ -436,7 +479,12 @@ const SHARD_KEY_COLUMNS: Record<string, string> = {
   memory_chunks: "chatId",
   mari_workspace_context: "chatId",
 };
-const SHARDED_TABLE_SET: ReadonlySet<string> = new Set(SHARDED_TABLES);
+// Deliberately mutable, unlike the arrays it mirrors: SHARDED_TABLES aliases
+// FILE_BACKED_TABLES, so a registered table becomes "sharded" by array identity
+// the instant it is pushed. If this set were the module-load snapshot it once
+// was, the load, flush and delete paths would each classify that same table
+// differently. Registration writes both, together, or neither.
+const SHARDED_TABLE_SET = new Set<string>(SHARDED_TABLES);
 
 /**
  * Chat-unit lazy tier (#5592 Phase 2). These tables no longer load at boot:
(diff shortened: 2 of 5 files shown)


Gunterlie:
Thanks — all three findings reviewed against the code. Pushed 6a274ca.

**1. Malformed file-table metadata read before the guard — valid, fixed.** `isFileTable()` only checks that the `marinara:file-table` symbol is present, so a value carrying `null` metadata passes it and `tableNameOf()` throws in `getFileTableConfig`, outside the per-table handler. That aborted the whole call and stranded every valid table registered after the bad one. The name is now read inside a guard and a bad definition is skipped with a log. The regression registers a broken-metadata object, a string and `null` in the same call and asserts the valid tables around them still register.

**2. Malformed rows dropped without preserving the source — valid, fixed.** With `skipped > 0` the shard was marked for rewrite and the next flush overwrote the primary from the normalized rows. Where no readable `.bak` held a copy, the dropped rows were unrecoverable. `loadRegisteredTableRows` now calls `preserveMalformedRowSourceSync` first and reports the file through `getQuarantinedTables()`, matching the boot loader. New regression case plants a shard with one good row and one malformed entry and asserts a `.corrupt-*` copy exists and still contains the dropped entry before the flush rewrites the shard.

**3. Reload path in `registersAndPersists` — correct observation, declining the suggested change; fixed the misleading comment instead.** You are right that the reopened store boot-loads `package_demo_notes` through `initialize()`, because the registries are process-global, so the re-registration there takes the idempotent path. That case is asserting persistence across a reload, not registration-time shard loading, and its comment claimed otherwise; the comment now says plainly that this goes through the boot loader and why.

`registerTables`' own shard loading is already covered directly, by the four cases that each use a table name this process has never registered: `healsRecoveredShardsOnRegistration` asserts rows already on disk are loaded by registration for a table the store has never seen, and `clearsShardsThatRecoverToNoRows`, `quarantinesUnreadableShardPairs` and the new `preservesMalformedRowsOnRegistration` do the same for the recovery paths. Adding a fifth fresh-name case that only re-asserts the plain load would duplicate coverage that already exists.

Gunterlie:
Valid — applied in 898f5c2, with one placement change.

The mixed call now runs *after* the tree-diff assertion rather than before it, because that assertion requires the sandbox to be byte-for-byte unchanged and `package_demo_after_broken` is meant to register successfully. The call itself is as suggested: a malformed-metadata object, a string and `null` ahead of a valid primary-key table, then an assertion that `package_demo_after_broken` reaches `FILE_BACKED_TABLES`. That is the behaviour that actually matters — not merely that the call does not throw, but that a bad definition does not strand the valid tables behind it in the same package.

Issue #5798, opened 2026-09-03
A way for the GM to write back to an Experience package


kolacheee:
# A way for the GM to write back to an Experience package

## What's missing, in one paragraph

An Experience package can *read* the chat: the host hands it props every turn, and it can
subscribe to a small set of engine-defined events. There is no lane in the other direction. When
the GM narrates "the storm rolls in" or "the innkeeper waves off your coin and gives you the room",
the package never hears about it, so the sky stays fair and the purse stays full. Everything the
narrator does to the world is invisible to the surface the player is standing on, and the only
thing keeping the two in step is that the package writes a few words into the prompt header and
hopes.

What this asks for is the reverse lane: **a package-addressed, structured output channel** — a
small closed vocabulary of named calls, with primitive arguments, validated by the server before
delivery, that reaches the package the way host events already do.

## Why this shape and not a free-form one

The founding rule of Pixelforge is *the model decides what happened; the algorithm decides how it
looks.* A free-form channel would hand the model geometry authority and re-open every seeding and
determinism guarantee the package is built on. A closed enum of verb names with typed arguments
does not: every call is a selection from a list the package published, and the package can refuse
one that does not fit the world it actually compiled.

## The delivery mechanism already exists

The Capability API 1.12 event bus is the pattern. The host dispatches a browser event —
`CAPABILITY_CLIENT_EVENT`, `"marinara-capability-server-event"`
(`packages/client/src/lib/capability-client-events.ts:1`) — and a package's element filters it on
package id and chat id. Pixelforge binds that listener once
(`packages/pixelforge/src/90-element.js:819`, in the Pixelforge repo) and routes it into a handler
that already consumes three event types (`packages/pixelforge/src/50-spatial.js:139`). What it
cannot do today is carry anything the *model* said: the type strings are engine-defined, so there
is no surface a real writer could sit on. That is the verified reason Pixelforge 0.14 ships a fully
working weather override with a browser console as its only writer.

So the ask is not a new transport. It is **the same event bus, with a payload the model can
produce** — concretely, a schema-validated custom tool call whose name is drawn from a closed enum
the package declares.

**Version:** this wants the next Capability API point release. The registry is
`packages/shared/src/schemas/capability-package.schema.ts` — `supportedCapabilityApi`. Two
observations, because the trees disagree: the review worktree at merge `34442e26d` reads
`{major: 1, minor: 13}` (line 184), and `feat/mari-write-readback` at `af2dc003a` reads
[...]

PR #5923, opened 2026-09-07
feat(game): skill checks are rolled by the server, not narrated by the model

### LINKED ISSUE #5949: Game skill checks can disagree with saved narration
## Problem
Game Mode can display a corrected dice result while saving the Game Master’s original inconsistent result in the message. Only the first check is audited in the browser, so additional checks can remain unresolved or inconsistent in saved history.

## Expected behavior
Reuse the existing dice parser and roller to resolve eligible sparse d20 checks on the server before saving. Keep displayed results and saved narration consistent, preserve valid player rolls, and leave unsupported dice systems unchanged.

## Existing work / ownership
Already implemented in PR #5923 by @kolacheee. @SpicyMarinara (Codex) is handling final staging integration and check/review follow-up in that same PR. No duplicate implementation PR is needed. Related planning: #5798.

kolacheee:
> [!IMPORTANT]

> [!IMPORTANT]
> **Stacked PR.** Branch `feat/gm-sparse-checks-server-rolled` sits on top of `fix/dice-notation-consolidation` (one shared dice grammar). Review and merge that one first; the diff shown here against `staging` will include its commit until it lands.
>
> **Scheduling — this must not land in the same playtest window as the GM verb runtime branch (`feat/5798-gm-verb-runtime`, #5798).** Both change what the GM is told to emit and what the engine does with the result. If a playtest night reads badly with both in, nothing tells you which one caused it.
>
> **One merge conflict to expect.** Against `feat/5798-gm-verb-runtime`, this branch conflicts in exactly one place — both insert their post-processing block at the same anchor in `generate.routes.ts`. The conflict is textual only; the two rewrites touch different tag names, so the order they merge in does not matter. Whoever merges second resolves it by hand. Everything else auto-merges, and this branch merges clean against `feat/game-mode-dice-tool`.

## Linked issue

Closes #5949

Tracking issue: #5949. Related Game Mode planning: #5798 (not closed by this PR).

## Why this change

The Game Master invented its own dice results and the engine mostly let it.

- The GM wrote the whole check itself — die, modifier, total, outcome — and the engine only ever second-guessed the **first plain d20 of a turn**, in the browser, after the fact. A turn with three checks audited one.
- **The live bug:** when that audit rejected the GM's arithmetic, the corrected numbers reached the animated dice card and **the invention stayed in the saved text**. The endpoint that rewrites a check decided what to overwrite by searching for `result=`, so a tag reporting a complete roll was never touched. The player saw an honest card; the transcript kept the lie; and on the next turn the model read its own invented number back out of its context as fact.
- Dice pools were pure narration — the engine never audited them and never rolled them, but nothing stopped a resolver from treating a pool it could not read as a plain d20 request and answering a six-die pool with one d20.
- The player has no way to tell an honest roll from a flattering one. That is the whole reason a table trusts dice.

## What changed

**One reader for check tags.** `parseSkillCheckTagBody`, its rolls reader, the tag type and the grammar itself move into `packages/shared/src/utils/skill-check-tag.ts`. The server used to read `skill` and `dc` with its own attribute search and throw the rest away, so an advantage mode or a die the player rolled themselves vanished the moment the server resolved a check. Both sides now reach the same verdict about whether the GM's numbers can be trusted.

**One service for rolling.** The modifier lookup a check needs — the skill bonus from the latest game-state snapshot, and the governing attribute from either the structured stats or the player card's free-form sheet — was 27 lines inlined in the `POST /game/skill-check` handler, so nothing else in the engine could roll a check the way the shipped endpoint rolls one. `skill-check-resolution.service.ts` now owns it; the endpoint is a thin caller; a turn with several checks costs one snapshot read and every check sees the same sheet. Three deliberate behaviour changes came with it: a tag the shared reader cannot read at all is left alone instead of overwritten (the old text search could reach past its target and clobber an unrelated malformed tag); unparseable stored stats now cost the check its modifiers and log a warning instead of failing the whole request, because a resolver that runs during generation must not be able to kill a turn; and a non-numeric skill bonus coerces to 0 instead of being concatenated as a string.

**Resolution at save time, with real randomness.** The prompt convention becomes sparse — both branches of the check line now ask the GM for `skill` and `dc` only, plus the player's own number where the player rolled. Generation post-processing then rolls **every** tag in the turn that still owes a roll and rewrites it in place, immediately before the `content_replace` frame, so the browser renders resolved text and the save carries it. The roller is the engine's own unseeded `Math.random` path, unchanged by the consolidation below it. It is idempotent (what it writes reads back as audited, so a second pass throws no dice), scoped to the newly generated segment on a continue, Game Mode only, never on impersonation, and a failure costs the check rather than the turn.

**The audit is ported server-side and now overwrites.** A self-reported plain d20 check whose arithmetic does not add up is re-rolled and its invented numbers leave the saved text. That closes the live bug above at both ends: the endpoint no longer skips a tag for carrying `result=`, and the decision about what to overwrite is the shared reader's rather than a text search's.

**Pools are never audited and never rolled — that is a rule now, not an accident.** `isEngineRollableSkillCheckTag` is the single place that decides whether the engine implements what a tag names: `resolution=` absent or `sum`, and `dice=` absent or a d20 notation the reader can restate exactly, with the die count agreeing with the declared mode (one die normally, two under advantage or disadvantage). `1d20+3` is refused for the same reason a pool is — it is a d20, but the engine's modifier comes from the character sheet, so rolling it would silently drop the GM's flat bonus and relabel the tag as if it had never asked. **All four doors onto a rewrite ask that same question:** generation post-processing, the `POST /game/skill-check` rewrite, the browser's legacy fallback, and the sparse pre-roll inference (so a pool die is never adopted as the player's own d20).

**The prompt carries the other half.** Both check branches now show the GM the complete pool tag, and say plainly that an incomplete one is left unresolved rather than repaired. Without that, the sparse convention had removed the full form from every template in the file, so the attribute most likely to go missing was exactly the one that used to get a pool converted into a d20.

**The tag grammar reads what the GM wrote, not how they spaced it.** Whitespace around `=` is tolerated at the one reader every door asks — which is also what the server's other check reader has always accepted, so the two stopped disagreeing about the same tag. An attribute written with no value (`mode=`) declares nothing and consumes nothing, so it can no longer reach across the space and swallow the declaration behind it. Advantage and disadvantage are read from `mode=` (compared lowercased, so a capitalised mode counts) or from a bare flag standing outside every `key=value` pair — never from a substring of the whole tag body, which used to let a skill *name* declare a mode.

### Evidence

Everything below was driven through the shipped code with live counters on the die and on the chat read, not asserted about the source:

- A sparse pool, spaced 2,401 different ways: left verbatim every time, zero dice thrown, zero context loads, both declarations read, and both the server's and the browser's predicates refusing it.
- A sparse d20, spaced 343 different ways: exactly one die each — and all 343 produced **one distinct output string**, so spacing is erased rather than merely tolerated.
- The empty-value class — a bare `key=` inserted at every position of six declaration shapes, nine keys, four gap characters, three spacings: **2,700 shapes, none rolled**.
- The advantage rules across 5 skill names × 9 mode declarations × 3 dice labels: 135 shapes, no failures, checking the parsed flags on both sides and the number of dice actually consumed.
- 400 spaced complete pools read through both server readers: no disagreements.
- Mutation sweeps over the shared reader, the resolver, the route, the browser fallback and the prompt: the pins fail when the guards are removed, including the two opposite ways of getting the grammar wrong. Four survivors remain where the shipped code is stricter than the test requires; they are listed as pin gaps, not defects.

## Validation

### Manual verification notes

Run in the review worktree, for information — the boxes above are yours to tick after you have seen this yourself:

- `pnpm check` exited 0 (full chain through the client build). `pnpm version:check` in sync for 2.4.5.
- `gm-skill-check-resolution.regression.ts` passed, and passed again after every mutation sweep restored the tree.
- Neighbours all passed: `dice-notation`, `dice-display`, `conversation-game-slash`, `game-narration-phase`, `game-checkpoint-engine-state`, `game-journal-edit`, `game-metadata-race`, `game-portrait-recovery`, `prompt`, `context-fit`.
- No `console.*` in any touched server file.

Still owed by hand — **this change needs a playtest night, not just a click-through:**

- [ ] Manually verify in the browser: a Game Mode turn where the GM asks for a check comes back with the check resolved in the text, the dice card matching the resolved numbers, and your character's skill and attribute modifiers applied.
- [ ] Manually verify a turn containing three checks resolves all three (the old behaviour resolved the first).
- [ ] Manually verify a check you rolled yourself keeps your number instead of being re-rolled.
- [ ] Manually verify an advantage check throws two dice and keeps the higher, and disadvantage keeps the lower.
- [ ] Manually verify a dice pool written by the GM is left exactly as written — no d20 card, no relabelled tag in the saved text.
- [ ] **Play a session and judge the two-beat narration** (see below). This is the question the change actually turns on.
- [ ] Manually verify the narration log and dice card render correctly in light and dark mode, and that a declined check still shows its system line rather than disappearing.

## Docs and release impact

- [ ] No docs changes needed

`CHANGELOG.md` gains one entry. No version-bearing file is touched. `docs/game/dice-and-skill-checks.md` is **not** updated here and is now incomplete — nothing in it becomes false, but the most player-visible part of this change (the two-beat narration), the pool rule, and the fact that a check can now be declined are undocumented. That edit belongs in the batched `[docs-i18n]` follow-up rather than in this PR.

No new UI. The animated dice card, the narration log line and the check tag all render exactly as they did; what changes is which numbers are in them. Worth a before/after screenshot of a resolved check in the transcript if you want one on the record.

---

### The deliberate trade: the narration is now two beats

The model writes the turn **before the number exists**. So it narrates up to the attempt, and the consequence lands on the following turn, once the resolved check is sitting in its context. That is the honest trade for a roll the player can trust on a connection that cannot call a dice tool, and it is the whole reason this change is worth having — but whether it *reads* well is a prose question that only a playtest night can answer.

This branch is deliberately isolated so that a "no" is one revert rather than an unpicking. If the two-beat rhythm reads badly, revert it and revisit — the separate `feat/game-mode-dice-tool` branch gets the number before the narration on connections that support tools, and this one stays the safety net underneath for the connections that do not.

### Known limitations, named here rather than left for review to find

1. **The audit is arithmetic-only, and a GM can opt out of it with a label.** A check that declares a system the engine does not roll — `resolution="successes"`, or a non-d20 `dice=` — is left verbatim, numbers and all. So a GM that writes a self-serving roll and labels it as a pool keeps it. This is pre-existing behaviour, but it matters more now for two reasons: the headline of this change is "rolled by the engine instead of written by the Game Master", and the prompt now shows the GM the exact attribute pair that exits the audit — which it has to, because that same complete pool template is what stops real pools being mangled.
2. **Even on the plain d20 path the audit only checks the arithmetic.** A natural 20 that adds up correctly is trusted verbatim. A GM that ignores the sparse convention and writes consistent, flattering numbers is not caught.
3. **A tag that declares no system at all is still answered with a d20.** `[skill_check: skill="Intimidation" dc="4" rolls="3|7|9"]` — no `dice=`, no `resolution=` — is rolled as a plain d20 and those three numbers are discarded. Pre-existing and arguably correct: with nothing declared there is nothing to distinguish it from a sparse d20 request, and the prompt tells the GM to write every attribute for a pool. But "pools are left exactly as the GM wrote them" is about *declared* systems, and this is the gap in that sentence.
4. **The advantage rule now reads only `mode=` or a bare flag, which has two edges.** A hyphenated `dis-advantage` reads as *advantage* (word boundaries), so it throws two dice and keeps the higher — no emitter in the app produces that spelling, and the prompt teaches `mode=`. And an **unquoted** multi-word skill name still leaks a flag word: `skill=Press the advantage` parses the skill as `Press` and picks up advantage from the leftover text. The tag is mangled either way in that case; this is inherent to having a bare-flag form at all.
5. **The empty-attribute rule is enforced now, and the obvious alternative is pinned as the thing not to do.** A bare `key=` declares nothing and consumes nothing. Closing that hole by requiring an unquoted value to touch its `=` would reopen the one the grammar was widened to fix — `dice = 6d10` would stop being read, and an unseen declaration is exactly the sparse-looking tag that gets handed to a d20. Both directions have tests.
6. **A check the engine declines leaves no prose in the next turn's context.** The context builder writes a `Skill check result: …` line only for a tag it can read completely; otherwise it drops the tag and writes nothing. So a GM that asks for a pool and writes it badly has no record next turn that it asked. This is strictly better than the old behaviour (a fabricated d20 read back as fact), and player-side it is softer — the narration log still renders the unresolved tag as a plain system line — but it is a real consequence of "leave it standing".
7. **Only the first check of a turn draws the animated dice card.** Every resolved check does appear in the narration log with its full formatted result; the card itself is first-check-only.
8. **A fourth reader of these tags still lives on the server** (`segment-edits.ts`), with its own hand-rolled scanner and no audit at all. The two now agree about spacing and about mode. They still disagree about one shape: on a complete pool written with an empty attribute before `resolution=`, the context line can read `→ 3.` where it should read `→ 3 successes.` Cosmetic, one input class wide, and it belongs to a follow-up about that reader rather than to this PR.
9. **The `POST /game/skill-check` endpoint still rolls a pool it then refuses to rewrite.** It resolves first and asks the guard second, so a caller that posts a pool still gets a d20 result back — only the saved transcript is protected. No shipped caller does that any more; a stale browser bundle would draw a d20 card over an untouched pool tag, which is the safe direction, but it is worth knowing.
10. **Three or four pin gaps where the shipped code is stricter than its tests.** `resolution=""` and `dice=""` are refused deliberately per the code's own comment but nothing pins that; one alternative spelling of the grammar's lookahead, and dropping the trim on `mode=`, also survive. None is a defect in behaviour — all were driven and refuse correctly — they are optional extra pins if you want them.
11. **Pool systems stay untouched.** This PR does not teach the engine to roll them, and deliberately does not widen the audit to cover them.

## Summary by CodeRabbit

* **New Features**
  * Game Mode skill checks are rolled by the engine using the correct character’s skills and attributes.
  * Multiple checks per turn are resolved, including inaccurate GM-provided results.
  * Player-rolled dice are preserved when provided.

* **Improvements**
  * Unsupported or unrollable checks remain saved without invented results.
  * Results appear between attempt and consequence narration.
  * Checks requesting both advantage and disadvantage remain unresolved without a result.
  * Client and server interpret skill-check tags consistently.

## Maintainer staging integration

Updated against current `staging` without conflicts. `pnpm check`, `gm-skill-check-resolution`, and neighboring dice-display, game-narration-phase, capability GM verbs/runtime, prompt, and context-fit regressions pass locally. This is automated proof, not a live-model playtest. The prerequisite dice-grammar PR is already on staging. Translation follow-up for `docs/game/dice-and-skill-checks.md`: #5951. Human validation checkboxes remain unchecked.


SpicyMarinara:
@SpicyMarinara (Codex) is handling staging integration and final check/review follow-up in this existing @kolacheee PR, at the maintainer’s request. Tracking issue: #5949. I will preserve the unsupported-system boundaries and distinguish automated proof from live-model playtesting.

PR #5985, opened 2026-09-09
fix: enforce package chat persistence permissions

### LINKED ISSUE #5899: Five capability permissions are declared in package manifests and enforced nowhere
A capability package manifest can declare nine permissions. Four of them are actually checked before
the matching capability is handed out; the other five — `chat-read`, `chat-write`, `network`,
`storage` and `ui` — are never checked anywhere in the server. They are strings a package writes into
its manifest and a user reads on the install screen, and nothing behind them changes based on what
they say. In particular, any package with a server entrypoint already receives the full persistence
host — read and write, across every chat, not scoped to its own — whether or not its manifest asked
for anything.

Where it is. The permission enum is
`packages/shared/src/schemas/capability-package.schema.ts:9-20`. A repo-wide search for permission
gates returns five call sites covering four permissions: `agent-runtime`
(`packages/server/src/services/capability-packages/capability-agent-runtime.service.ts:15`),
`conversation-actions` and `prompt-context`
(`packages/server/src/services/capability-packages/capability-module-runtime.service.ts:220` and
`:234`), and `routes`
(`packages/server/src/services/capability-packages/capability-route-registration.service.ts:71`,
`packages/server/src/services/capability-packages/package-manager.service.ts:447`). Nothing checks the
other five. The persistence host is attached unconditionally in the same activation context that does
the other gating — `capability-module-runtime.service.ts:98`, `persistence:
createCapabilityPersistenceHost(app.db)` — and that host includes whole-chat metadata writers
(`packages/server/src/services/capability-packages/capability-persistence.service.ts:507-545`). The
permission list is shown to the user in the catalog detail pane before install
(`packages/client/src/components/agents/AgentCatalogView.tsx:731-742`) and nowhere after it. Packages
do declare these strings in good faith — the Pixelforge build script declares `chat-read`,
`chat-write`, `ui` with a comment saying the point is to disclose access to users
(`scripts/build-pixelforge-package.mjs:126-130` in the Pixelforge package repo).

Why it matters. The install screen makes a promise the engine does not keep. A user reading a package's
permissions is being told what it can reach, and for five of the nine that list is decoration — a
package that declares nothing can still write every chat's metadata. It also means the disclosure runs
backwards: the honest packages that list `chat-write` look more dangerous than a package that lists
nothing and has identical access.

The smallest honest fix is probably not "enforce all five at once" — `ui` and `network` in particular
would need a definition of what enforcement even means for a client bundle — but to close the gap
between what the list claims and what the list does, one permission at a time, starting with the ones
that gate a real server-side capability. `chat-read` and `chat-write` are the two with an obvious
enforcement point: the persistence host could be built per-package from the manifest rather than handed
out whole. Two things are worth deciding alongside that. First, every gate that exists today guards a
*server entrypoint registration*, so a gate on a package with no server code at all would be a new kind
of check, not the existing kind applied once more. Second, permissions are displayed pre-install only,
so any change that widens what an already-granted permission means reaches existing users with no
re-consent — the write-back channel in #5798 will be the first thing to actually enforce `chat-write`,
and it widens the meaning from "this package's own code writes chat data" to "the engine will execute
writes on this package's behalf, from model output". That widening is worth an answer here rather than
inside that PR.

SpicyMarinara:
## Linked issue

Closes #5899.

## Why this change

A package with no declared chat permissions still received unrestricted chat persistence methods. The catalog could also show a newer version's permissions as though they described the installed version.

## What changed

Package activation builds the persistence host from its declared permissions. Chat/message/game-state and spatial-snapshot reads require chat-read; chat/message/event and snapshot writes require chat-write. Transactions retain the same checks, and package chat locks require at least one chat permission. Trusted Engine callers retain their existing host.

The existing catalog detail view shows installed permissions and, when different, the catalog version's permissions separately. It explains the API checks and model-command write scope while explicitly identifying network, storage and UI as trusted-code access declarations. These checks do not sandbox package JavaScript or change the existing exact-version/checksum install approval flow.

## Validation

Passed `pnpm check`, `capability-persistence-permissions.regression.ts`, the package lifecycle/readiness regression, and the Game Master verb runtime regression. The new proof activates four real local package modules through registry loading, file digest verification, runtime snapshots and service registration. It tests no permissions, read-only, write-only and both permissions against actual file storage; denied methods, nested transaction sessions, chat locks, persisted positive writes and manifest-array mutation are covered. The same proof fails against the original activation/persistence code because a denied read succeeds. Updated the existing readiness fixtures to declare the chat reads they perform.

Passed 6 browser cases across desktop Chromium, mobile Chromium and mobile WebKit, covering empty/read-only installed permissions, differing update permissions and reloads. Captured both themes and inspected mobile screenshots. The UI cases use catalog/installed API fixtures; the separate server proof uses actual package activation and persistence. Screenshots are attached to the browser test artifacts.

- [ ] Manually verify an installed third-party package declares every chat permission its persistence calls require.
- [ ] Manually verify the permission display on physical mobile hardware.

## Docs and release impact

Updated the optional-package contributor guide, English UI copy and Unreleased changelog. Translation follow-up: #5986. Packages that use undeclared chat APIs must add the relevant permission to their manifest; this is an API enforcement change, not a JavaScript sandbox.

## Summary by CodeRabbit

- **New Features**
  - Package persistence now enforces declared `chat-read` and `chat-write` permissions for chat and spatial snapshot access, including transactions.
  - Agent details distinguish permissions for installed and catalog versions.
  - Permission explanations clarify API access gates versus trusted package-code access.

- **Documentation**
  - Updated Capability API and package installation guidance to reflect permission requirements and approval behavior.

- **Tests**
  - Added coverage for permission enforcement, catalog display, theme rendering, and state persistence.


SpicyMarinara:
@coderabbitai review

SpicyMarinara:
@coderabbitai review

SpicyMarinara:
CodeRabbit completed its hosted review of `2f22933b6fdc3c85aa383edc7996b4651849b3ad` with no actionable findings and no established merge-blocking issue. The subsequent merge of staging (`8713607e0`) only combines already reviewed staging changes; the permission implementation is unchanged. Both Unreleased entries were retained when resolving the changelog conflict.

After that update, `pnpm check`, the real installed-module permission matrix, and the package lifecycle/readiness regression all passed. The earlier complete Linux CI and six browser cases also passed. Required GitHub checks are running for the updated branch and remain required before merge.

SpicyMarinara:
Post-merge CI follow-up: the optional desktop browser matrix finished with one timeout in the existing Background Library selection-marker fixture (`ux-feedback-sweep.e2e.ts:283`). The required checks had passed before merge. The fixture and Background Library implementation are unchanged by this PR, and the exact case passes on the current staging tree (one desktop Chromium case, 13 seconds including setup). The package permission regression and its dedicated browser cases remain green; no permission implementation change is warranted by this unrelated locator timeout.

PR #6114, opened 2026-09-13
fix(mari-db): resolve capability package tables registered after load

### LINKED ISSUE #6113: Professor Mari fails on capability package tables: Unknown file-backed table
## Bug
With a package that registers its own tables (e.g. Slurp2 `slurp2_*`), every Professor Mari DB command on those tables fails:

```
Unknown file-backed table: slurp2_account_subscriptions
  at getMeta (mari-db.service.js)
```

## Cause
`mari-db.service.ts` builds `TABLE_METAS` once at module load. Packages call `db._fileStore.registerTables` later, so Mari never sees them. The store itself registers them fine; the package works, only Mari is blind.

## Impact
Existing users with Slurp2 installed: Mari cannot read/write Slurp2 data. Data on disk is intact; no migration needed once fixed.

Gunterlie:
Fixes #6113

- `file-backed-store.ts`: export `getRegisteredFileTable(name)` (live registry lookup).
- `mari-db.service.ts`: `getMeta` falls back to the live registry and caches the meta. Unregistered names still throw.
- Regression: `package-table-registration.regression.ts` lists a registered package table through Mari.

Verified: server `tsc --noEmit` clean; regression passes with fix, fails without it with the exact prod error.

Existing users: update Engine and restart; no data repair.

## Summary by CodeRabbit

- **Bug Fixes**
  - Package-registered database tables are now recognized when accessed after startup.
  - The database listing command can now read and display data from tables registered by capability packages.
  - Unknown table requests continue to return an error.

### DIFF
--- a/packages/server/src/db/file-backed-store.ts
+++ b/packages/server/src/db/file-backed-store.ts
@@ -865,6 +865,11 @@ const tableMetasByObject = new WeakMap<object, TableMeta>();
 const columnMetasByObject = new WeakMap<object, ColumnMeta>();
 const tableMetasByName = new Map<string, TableMeta>();
 
+/** Live lookup, so consumers that snapshot the schema at load still see package tables registered later. */
+export function getRegisteredFileTable(name: string): Table | undefined {
+  return tableMetasByName.get(name)?.table;
+}
+
 function tableNameOf(table: Table): string {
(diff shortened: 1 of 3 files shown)


Gunterlie:
@coderabbitai review

PR #6190, opened 2026-09-14
fix(capability): keep restart-pending agents visible

Gunterlie:
## Why

Uninstalling one capability package can refresh the capability-agent registry and hide another package that is still active while its update waits for restart. This was reported with Slurp Legacy and Slurp Remastered: Slurp Remastered remained active on Home and installed in Download Agents, but disappeared from Installed Agents.

## What changed

- Keep the previous agent definitions for a `restart-required` package until the Engine restarts.
- Reuse the same servable-package fallback already used by active client surfaces.
- Add regression coverage for the registry hydration contract.

## Validation

- `pnpm exec prettier --check packages/server/src/services/capability-packages/package-manager.service.ts scripts/regressions/agent-registry-hydration.regression.ts`
- `pnpm build:shared`
- `node scripts/run-regressions.mjs --filter scripts/regressions/agent-registry-hydration.regression.ts`
- `pnpm check` passed lint and type checks. The client build exceeded the local 120-second command timeout.

## Scope

- Engine only. No Marinara-Agents package change is required.
- Manual browser verification remains to be performed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Agent definitions now remain visible while a package update is waiting for a restart.
  - During this transition, the system continues using the previous runtime version until the restart is completed.
  - Packages without an available runtime remain excluded as before.

- **Tests**
  - Added regression coverage to verify that agent definitions are retained during restart-required updates.


Gunterlie:
Replaced with an organization-owned branch and PR.

Issue #6119, opened 2026-09-13
Pressing Ctrl+C in the Windows launcher window kills the server outright instead of shutting it down


kolacheee:
On Windows, stopping the app with Ctrl+C in the launcher console ends the server abruptly rather than shutting it down cleanly. The console never prints "Shutdown complete", and the next start warns that your previous session "ended without shutting down" and was force-quit or lost power - even though you stopped it the normal way. Anything saved in the last fraction of a second before Ctrl+C can also be lost even though the app told you it saved.

## How to reproduce (try it yourself)

This only happens on Windows. Run the same steps on macOS or Linux with `./start.sh` as a control - there everything works correctly.

1. Check out `staging` at 52799569c and build as usual.
2. Double-click `start.bat`, or run it from a `cmd.exe` window. Wait until the console prints `Starting Marinara Engine on http://0.0.0.0:7860` and then the server's own `Marinara Engine server listening` line.
3. Let it sit idle for about ten seconds. You do not need to create anything - the bug does not depend on a pending save.
4. Press Ctrl+C once in the console.
5. Look through the console output for the line `Shutdown complete`.
6. Run `start.bat` again and read the first few seconds of console output.

**What you'll see:** at step 5 there is no `Shutdown complete` line. At step 6 the new session prints `Previous session (pid NNN) ended without shutting down: ... Nothing was logged, so the process was ended from outside - ... on desktop, a force-quit or power loss. See Support Diagnostics for the record.` That warning appears after every ordinary Ctrl+C.

**What should happen:** the console prints `Received SIGINT; shutting down Marinara Engine` and then `Shutdown complete`, exits cleanly, and the next start prints no force-quit warning - exactly what happens today on macOS and Linux.

Do not use the `Received SIGINT; shutting down Marinara Engine` line as your test: the Windows console also sends the stop key to the server directly, so that first line often prints a millisecond before the server is killed. Only the missing `Shutdown complete` and the next-boot warning reliably show the problem.

Worth doing in the same pass, because it would make this worse: start a local model so a sidecar process is running, note its process id in Task Manager, press Ctrl+C, then check whether the sidecar is still alive and still holding its port. Nothing reaps it if the clean shutdown is skipped.

[...]

PR #6195, opened 2026-09-14
fix(server): shut down cleanly when the terminal window closes

### LINKED ISSUE #6183: Closing the terminal window ends the server abruptly instead of shutting it down, leaving the storage writer lease behind
On macOS, Linux, and Termux, closing the terminal window that runs `start.sh` (or losing the SSH session it runs in) ends the server on the spot instead of shutting it down. Nothing is flushed or released: the storage writer lease stays on disk, up to about three quarters of a second of saves the app already confirmed can be lost, a running local-model sidecar is not reaped, and the next start reports that the previous session "ended without shutting down" as if the process had been force-quit or lost power - even though the user only closed the window. Pressing Ctrl+C first avoids all of it, which is not something users know to do.

The lease itself recovers: the next start reclaims it through the same-host process check and logs a warning. The abrupt ending is the bug, and the leftover lease is its most visible symptom (users see the `.writer-lease` folder and the reclaim warning and reasonably assume something is wrong).

## How to reproduce (try it yourself)

Linux or macOS, any recent `staging` (verified at f2219d815).

1. Open a terminal window and run `./start.sh`. Wait for the server's `Marinara Engine server listening` line.
2. Close the terminal window with its close button (do not press Ctrl+C). The same thing happens if you run `kill -HUP <pid of the shell>` from another terminal.
3. Look at `packages/server/data/storage/`.
4. Run `./start.sh` again and read the first few seconds of output.

**What you'll see:** at step 3 the `.writer-lease` folder is still there. At step 4 the server logs `[file-storage] Reclaimed the writer lease after confirming its same-host PID exited`, and Support Diagnostics records the session as ended without shutting down (the `Previous session (pid N) ended without shutting down ... a force-quit or power loss` warning appears once the session is old enough to have a heartbeat).

**What should happen:** the same clean stop as Ctrl+C. `Shutdown complete` is reached (it cannot be printed - the terminal is gone - but the lease folder is removed, the session heartbeat is stamped `clean`, and the next start prints no reclaim warning and no force-quit report).

## Where it is

At staging f2219d815:

- `packages/server/src/index.ts:116-121` - the only signal handlers are `SIGTERM` and `SIGINT`. There is no `SIGHUP` listener, so Node applies the default action (terminate immediately) when the hangup arrives.
- `scripts/run-server.mjs:9-20` - the launcher supervisor already relays `SIGHUP` to the server (`child?.kill(signal)`), but relaying does not help: when a terminal closes, the shell hangs up the whole foreground job, so the server receives `SIGHUP` directly at the same moment and dies before the relay lands.
- `packages/server/src/lib/logger.ts:17-20` and `packages/server/src/app.ts:94-98` - the two Pino instances write to stdout through Pino's default asynchronous stream. Pino's own guard (`buildSafeSonicBoom` in `pino/lib/tools.js`) only tolerates `EPIPE`; a hung-up terminal answers every write with `EIO`, which Pino re-emits as an unhandled `'error'`. And Pino's exit-time `flushSync` (`sonic-boom/index.js`, the `flushSync` loop) retries a failed write every 100 ms without end for any error that is not `EAGAIN`/`EBUSY`.

## What's happening

Two layers, and the second only shows up once the first is fixed.

1. No `SIGHUP` handler, so the process is terminated by the signal itself. The debounced store flush, the `onClose` hook that releases the writer lease, the sidecar reaper, and the session exit stamp (`process.on("exit")` hooks) never run. That is exactly what the next start reports.

2. Simply adding a `SIGHUP` handler that runs the existing `shutdown()` is not enough, and this is the part worth knowing before fixing it. By the time the shell forwards `SIGHUP`, the pty is already gone, so stdout returns `EIO`. The graceful close finishes within a couple of milliseconds - before the failed asynchronous write for the "Received SIGHUP" line has even reported back - so Pino never learns the stream is dead. At `process.exit`, Pino's exit hook calls `flushSync` on the buffered "Shutdown complete" line, gets `EIO`, and retries forever. Measured with a real pty: the store flushed and the lease was released, but the process then sat in that loop for 10.0 s until the launcher's SIGKILL backstop killed it (exit status 137), so the session exit stamp was never written and the next start still reported a force-quit. With the terminal alive (`kill -HUP` on the shell while the pty stays open) the same handler works perfectly, which is why this is easy to get wrong in a pipe-based test.

## Suggested fix

- In `index.ts`, on non-Windows platforms, treat `SIGHUP` like `SIGTERM`: `process.on("SIGHUP", () => void shutdown("SIGHUP"))`. Windows raises `SIGHUP` for a closed console through a different mechanism (#6119) and should be left alone until someone can verify it there.
- In `logger.ts`, extend Pino's dead-pipe tolerance to a dead terminal, for both Pino instances (the shared logger and Fastify's `app.log`): on an `EIO`/`EPIPE` `'error'` from the underlying stream, replace `write`/`end`/`flushSync`/`destroy` with no-ops the way Pino already does for `EPIPE`; and register a `process.on("exit")` hook before any Pino instance exists that checks "stdout was a terminal at startup and `isatty(1)` is false now" and silences the streams, so Pino's exit flush has nothing to retry. `isatty(1)` reliably flips to `false` after a pty hangup (verified).
- Coverage: a regression that hangs up a real pty, not just a pipe - the pipe case passes without the second half of the fix. `python3`'s `pty` module or util-linux `script` can provide the pty on Linux CI; the pipe-based `SIGHUP` case can run everywhere the existing `server-signal-shutdown` lane runs.

## Related report

A Podman user on the v2.4.4 image reported having to delete `data/storage/.writer-lease` before almost every launch. That is a different, already-fixed cause: on v2.4.4 a `podman stop` with any browser connection still open exceeds the 10 s stop window and is SIGKILLed (fixed on staging by #5838), and the lease it leaves can never be reclaimed after a host reboot because v2.4.4 has no boot-id proof for a container's `hostId: null` lease (fixed on staging by #5744). Both were verified against the `latest` (v2.4.4) and `staging` GHCR images under rootless Podman 5.8.4: `latest` is blocked after a simulated reboot and leaves the lease on a stop with an in-flight request; `staging` reclaims the lease in every case (SIGKILL, container recreate, reboot) and stops cleanly in 4.3 s. That report needs the next release, not a code change; this issue is only about the terminal-close path.

## Context

Found while investigating that Podman report and re-verified on Fedora Linux with Node v24.21.0 against staging f2219d815 on 2026-09-13, using a Python pty harness (interactive bash in a pty running `node scripts/run-server.mjs dist/index.js`, then closing the pty master). The two-layer diagnosis above comes from instrumenting the exit hooks in the built server; the numbers (10.0 s, exit 137) are measured, not estimated. Please verify on macOS before closing; the Linux behavior is confirmed.

kolacheee:
## Linked issue

Closes #6183

## Why this change

Closing the terminal window that runs `start.sh` or `start-termux.sh` (or losing that SSH session) hangs up the whole launcher job. The server had no `SIGHUP` handler, so Node ended it on the spot: the storage writer lease stayed on disk for the next start to reclaim (users see the `.writer-lease` folder and the reclaim warning and assume something broke), up to ~750 ms of already-confirmed saves could be lost, a running local-model sidecar was not reaped, and Support Diagnostics reported the session as a force-quit. Only Ctrl+C avoided it.

Handling the signal alone is not enough, which is the part worth knowing when reviewing: by the time the shell forwards `SIGHUP` the pty is gone and stdout answers `EIO`. Pino only tolerates `EPIPE`, and its exit-time `flushSync` retries a buffered line every 100 ms without end. A close that finishes before the failed async write reports back (the normal case, ~2 ms) therefore released the lease but then hung until the launcher's ten-second SIGKILL and never stamped its exit - measured at 10.0 s / exit 137 with a real pty before the second half of this fix.

## What changed

- `packages/server/src/index.ts`: `SIGHUP` now runs the same `shutdown()` as `SIGTERM`/`SIGINT` on non-Windows platforms. Windows raises `SIGHUP` for a closed console through a different mechanism (#6119) and is deliberately left as is.
- `packages/server/src/lib/logger.ts`: `guardLoggerAgainstLostOutput()` extends Pino's dead-pipe tolerance to a dead terminal for a Pino instance - on an `EIO`/`EPIPE` `'error'` from the underlying stream it stops writing the way Pino already does for `EPIPE` (other errors still surface as before). A `process.on("exit")` hook registered before any Pino instance exists silences the guarded streams when stdout was a terminal at startup and `isatty(1)` is false at exit, so Pino's exit flush has nothing to retry. Applied to the shared logger here and to Fastify's own `app.log` in `app.ts`.
- `scripts/regressions/server-hangup-shutdown.regression.ts` (new): stage 1 hangs up a pipe-backed launcher job with a process-group `SIGHUP` the way a shell does and asserts the shutdown lines, launcher exit 0, lease released, and a `clean` session stamp; stage 2 hangs up a real pty through `python3`'s `pty` module (skipped with a log line when `python3` is missing) and additionally asserts the job ends well under the SIGKILL backstop. Red on unchanged `staging` (missing shutdown line), green with this change (pty stage completes in ~55 ms).
- `docs/TROUBLESHOOTING.md`: the "Restart Server does not return" section now says closing the terminal on macOS/Linux/Termux stops the server the same way Ctrl+C does.
- `CHANGELOG.md`: Unreleased entry.

Not changed: `scripts/run-server.mjs` already relays `SIGHUP`; that relay was never the problem (the shell delivers `SIGHUP` to the server directly at the same moment). No `nohup` special-casing: Node resets signal dispositions at startup, so a `nohup`'d Node server never ignored `SIGHUP` before this change either (verified: `nohup node -e 'setInterval(()=>{},1000)' &` dies on `kill -HUP`).

Known limitation, shared with `SIGTERM`/`SIGINT`: the handler is registered where those two are, after `buildApp()` returns, because `shutdown()` needs the app and the env watcher. A hangup that lands during the few seconds of startup still ends the process abruptly, exactly as `SIGTERM` does today; the next start reclaims the lease. Moving all three signals ahead of `buildApp()` with a pending-signal state is a separate change.

## Validation

### Manual verification notes

Automated evidence from the authoring machine (Fedora Linux, Node v24.21.0), not a claim of manual testing on other platforms:

- `pnpm check` passed on this branch.
- New lane `node scripts/run-regressions.mjs --filter scripts/regressions/server-hangup-shutdown.regression.ts`: passes here (pipe stage and real-pty stage); fails on unchanged `origin/staging` at the "Received SIGHUP; shutting down" assertion.
- Existing lanes `server-signal-shutdown`, `shutdown-deadline`, `restart-supervisor`, `file-backed-shutdown` still pass.
- Real-terminal reproduction with a Python pty harness (interactive bash in a pty running `node scripts/run-server.mjs dist/index.js`, then closing the pty master), on the built `dist`: before this change the server died instantly with the lease left behind and no exit stamp; with only the `SIGHUP` handler it released the lease but hung 10.0 s until the launcher's SIGKILL (exit 137) with no exit stamp; with this change it exits in 0.1 s, the lease is gone, the heartbeat reads `exitKind: "clean"`, and the next start prints no reclaim warning and no force-quit report. Same result at the default `LOG_LEVEL=warn` and at `LOG_LEVEL=info`. A hangup with the terminal still alive (`kill -HUP` on the shell) prints both shutdown lines normally.

Please verify before merging:

- [ ] Manually verify on macOS: run `./start.sh`, close the Terminal window, confirm `packages/server/data/storage/.writer-lease` is gone and the next `./start.sh` prints no "Reclaimed the writer lease" or "ended without shutting down" line.
- [ ] Manually verify on Termux: run `start-termux.sh`, swipe the session closed, confirm the same two things.
- [ ] Manually verify a local model is stopped when the terminal closes (start the sidecar first, close the window, check no `llama-server` remains).
- [ ] Manually verify Windows behavior is unchanged (this PR does not register `SIGHUP` on `win32`).

## Docs and release impact

- [ ] No docs changes needed

`docs/TROUBLESHOOTING.md` changed one paragraph; the `docs-i18n` mirror is pending a batched `[docs-i18n]` follow-up.

Not applicable: server-side shutdown behavior only.


kolacheee:
Superseded by #6193, which carries the same fix for #6183 (SIGHUP routed through normal shutdown, Pino stdout guarded against a hung-up terminal, exit-time flush silenced when the terminal is gone) plus macOS-verified terminal detection and development pino-pretty coverage. Closing to avoid two implementations of the same mechanism conflicting in logger.ts and app.ts. The troubleshooting note from this PR (closing the terminal now stops the server the same way Ctrl+C does) can follow separately.

Issue #6208, opened 2026-09-14
Conversation mode renders {{user}} in a character's message as the character's own name


kolacheee:
In a Conversation chat, `{{user}}` inside a character's message is displayed with the character's own name (or the literal "Assistant" when the chat has no character) instead of the user's name. The model receives the right text; only what is shown on screen is wrong.

Found while investigating an active-persona report; re-verified on `staging` at 14de2d1ad, and the same code is in v2.4.4.

## How to reproduce (try it yourself)

1. Create a persona "Alice" and select it in a Conversation chat (or set it as active and leave the chat without a persona).
2. Get a character message that contains `{{user}}`. The quickest way without a model: give the character a greeting like `Nice to meet you, {{user}}.` and start a Conversation with it, or edit an existing character message to contain `{{user}}`.
3. Compare the rendered message with Peek Prompt.

**What you'll see:** the message on screen reads "Nice to meet you, <character name>." With no character in the chat it reads "Nice to meet you, Assistant." Peek Prompt shows the same message as "Nice to meet you, Alice."

**What should happen:** `{{user}}` renders as the user's persona name (Alice), the way Roleplay messages already do.

## Where it is

At staging 14de2d1ad, `packages/client/src/components/chat/ConversationMessage.tsx`, the `macroContext` memo at lines 333-345: `userName: displayName` (line 335) is the cause. `displayName` (lines 302-308) is the message author's label, so for any non-user message it is `primaryCharInfo?.name ?? "Assistant"`. `{{user}}` resolves from `userName` (`packages/client/src/lib/chat-macros.ts`, `user: userName ?? persona?.name ?? "User"`), so a character message substitutes the speaker's name. The `persona.name: displayName` on line 337 is shadowed by `userName` and does not matter; the persona description/personality/backstory/appearance/scenario fields on the following lines still come from the real persona, so only the name macros are wrong: `{{user}}`, `{{userName}}`, and the phonetic variants that fall back to the user name. The same context feeds the display-side regex scripts, so a Find/Replace using `{{user}}` matches the character name in Conversation mode too. The Roleplay equivalent (`ChatMessage.tsx`) passes the user's name as `userName`, and the server (`macro-context.ts`) always uses the persona name. In v2.4.4 the same line is at 321.

## Suggested fix

[...]

Issue #6207, opened 2026-09-14
There is no way to clear the active persona once one has been set


kolacheee:
Once you set a persona as active in the Personas panel, there is no way to go back to having no active persona. The check button ("Set as active") only appears on personas that are not active, the active one has no "unset" control, no persona editor or settings screen offers one, and the persona API has no call for it. The only things that clear it are deleting that persona or restoring a backup.

Reported by a user on v2.4.4 ("is there a way to have no active persona in ME after setting one? like clear it completely"). Re-verified on `staging` at 14de2d1ad. Companion issue about what "active" actually does and how the UI hides it: #6206.

## How to reproduce (try it yourself)

1. Open the Personas panel and create two personas, for example "Alice" and "Bob".
2. Hover "Alice" and click the check button ("Set as active"). Alice gets the highlight ring and the check badge.
3. Hover "Alice" again and look for a way to turn that off. Hover "Bob": the only choice is to make Bob active instead.
4. Open a persona's editor and the Library view: the state is displayed, never toggled.

**What you'll see:** one persona is always active from now on. Switching to Bob just moves the flag; nothing removes it.

**What should happen:** a way to return to "no active persona", the same way the in-chat persona switcher offers "None". Either the check button on the active persona toggles it off, or a "Clear active persona" action exists, with API support.

Why it matters: the active persona is silently used as the identity for every Conversation-mode chat that has no persona selected, and choosing "None" in such a chat's switcher falls straight back to it. So a user who tried the feature once cannot get an anonymous Conversation chat again without deleting the persona.

## Where it is

At staging 14de2d1ad:

- `packages/client/src/components/panels/PersonasPanel.tsx` - the row actions render the check button only when the row is not active (`{!active && (<button ... activatePersona.mutate(persona.id)`, around line 1239). This is the only caller of `useActivatePersona` in the client; bulk selection offers export and delete only.
- `packages/server/src/routes/characters.routes.ts` - `PUT /personas/:id/activate` (line 2538) is the only persona route that touches the flag, and it requires an existing persona id (a stale id returns 404 without clearing, by design since #3041). `PATCH /personas/:id` (line 2386) validates with `personaUpdateInputSchema` (`packages/shared/src/schemas/persona.schema.ts` line 296), which has no `isActive` field, so `{"isActive": false}` is stripped, the body is empty, and the route returns the unchanged persona with 200 (verified).
[...]

PR #6213, opened 2026-09-14
docs: refresh slash command and persona guides in all language packs

### LINKED ISSUE #6206: Setting a persona as active looks like it does nothing: docs promise it applies to every chat, code applies it to Conversation only, and the chat UI reports None
Setting a persona as active in the Personas panel looks like it does nothing, and the app contradicts itself about what it does. The documentation says the active persona is used by any chat that has no persona of its own (Game Mode excepted), but the code only applies it to Conversation-mode chats; Roleplay and Game chats never use it, and their setup defaults to "Stay anonymous". In the one place it does apply, the chat UI says the opposite: the persona switcher shows "None - No persona selected" as checked and the composer avatar shows "?", while the prompt sent to the model and the name on your own messages use the active persona. The only visible feedback in the Personas panel is a highlight ring and a check badge; the persona also moves to the top when the list is sorted by Recent, but only because activating it refreshes its "updated" time.

Reported by a user on v2.4.4 ("having an active persona also doesn't really do anything ... it doesn't change current persona in a chat (at least in RP mode) and doesn't auto-select it when making a new chat, all i can see that it does is moves it to the top of the folder"). Re-verified on `staging` at 14de2d1ad.

## How to reproduce (try it yourself)

1. Create a persona "Alice" and set it as active in the Personas panel.
2. Start a new Conversation chat and do not pick a persona. Open the persona switcher next to the composer, then send a message or use Peek Prompt.
3. Start a new Roleplay chat. Look at the persona step in the setup, then open the same switcher and Peek Prompt in that chat.
4. In that Roleplay chat, type `Hi, I am {{user}}` in the composer and send it.

**What you'll see:** in the Conversation chat the switcher shows "None - No persona selected" as checked and the avatar shows "?", yet your message is labeled "Alice" and the prompt contains Alice's persona block. In the Roleplay chat the persona step starts on "Stay anonymous", the switcher shows None, and the prompt has no persona at all. At step 4 the stored message reads "Hi, I am Alice" (the composer expanded the macro with the active persona) while the model sees a persona-less chat.

**What should happen:** the documentation, the prompt, and the chat UI should agree. If Conversation-only is the intended rule (it was chosen deliberately in #3715), the docs and the Personas panel should say so, and a persona-less Conversation chat should show "Alice (default)" instead of "None". If the documented rule is the intended one, Roleplay setup should preselect the active persona with "Stay anonymous" one click away. Either way the composer must not expand `{{user}}` with a persona the server will not use.

## Where it is

At staging 14de2d1ad:

- `packages/shared/src/utils/chat-persona.ts` - `resolveChatPersonaCandidate` falls back to the active persona only when `chatMode === "conversation"`. Deliberate: CHANGELOG entry from #3715 ("only Conversation mode may fall back to the globally active Persona"), used by `packages/server/src/services/chat-user-identity.ts` for prompts and prompt previews.
- `docs/characters/choosing-your-persona.md` lines 11 and 64-70, and `docs/characters/personas.md` line 13 - still describe the active persona as the fallback for "any chat that has no persona of its own", excepting only Game Mode. This text predates #3715 and was never updated, so users are told to expect exactly what the reporter expected.
- `packages/client/src/components/chat/ChatArea.tsx` - `useActivePersona(... chatMode === "conversation")` at line 872 and the identity memo at line 1082 label the user's messages with the active persona in Conversation chats.
- `packages/client/src/components/chat/QuickPersonaSwitcher.tsx` - `activePersonaId = chat?.personaId ?? null` with no fallback, so the switcher and the composer avatar report None for the same chat. Choosing "None" there sets `personaId` to null, which in a Conversation chat falls straight back to the active persona, so "None" cannot mean "no persona" while one is active (see the companion issue about clearing it).
- `packages/client/src/lib/chat-macros.ts` - `selectActivePersona` uses `allowGlobalFallback = chat?.mode !== "game"`, so the Roleplay composer (`ChatInput.tsx`, `resolveInputMacros` applied to the outgoing text) expands `{{user}}` with the active persona and stores it, unlike the server. It should reuse the shared `resolveChatPersonaCandidate`.
- `packages/client/src/components/panels/PersonasPanel.tsx` - the "Set as active" tooltip, the row highlight, and the avatar check badge (around line 1208) are the only feedback; there is no explanation of scope. The move-to-top comes from `setActivePersona` bumping `updatedAt` and the list's default updated-at ordering.
- `packages/client/src/components/chat/ChatSetupWizard.tsx` and the persona picker - the persona step starts on "Stay anonymous" (`onChange(null)`), never on the active persona.

Other readers of the flag that are not mode-gated and keep working regardless: tracker card color defaults (`TrackerCardColorSettings.tsx`), DJ Mari playlists, and the `mari personas active` command.

## What's happening

The server and the chat header treat the active persona as a Conversation-only default, the switcher, avatar, and setup wizard only look at the chat's explicit persona, the composer uses a third rule of its own, and the docs describe a fourth. The mode the reporter tried it in (Roleplay) is excluded by design without anything saying so.

## Suggested fix

- Decide the rule (product call): keep Conversation-only, or restore the documented "any chat without its own persona" behavior. Update `docs/characters/choosing-your-persona.md` and `docs/characters/personas.md` to match.
- Client: in Conversation chats without a persona, resolve the effective persona through the shared `resolveChatPersonaCandidate` and show it in the switcher and avatar as a default (a "default" badge, not a chat-level selection). Make `chat-macros.ts` use the same shared helper so the composer matches the server.
- Personas panel: say what active means in the tooltip and, if there is room, a short caption.

## Context

Found while investigating a user report against v2.4.4; the behavior is unchanged on staging. Verified with the running app and the API (prompt preview) and a nine-way adversarial code review, not from a single reading. Please verify before fixing.

### LINKED ISSUE #6211: [docs-i18n] docs/chats/slash-commands.md, docs/chats/messages.md

The English slash-command guides are being updated in #6210. Translated documentation on `docs-i18n` must match so users can discover command arguments and the correct visibility syntax.

## Proposed solution

Update every translated language pack on `docs-i18n` for:

- `docs/chats/slash-commands.md`
- `docs/chats/messages.md`

Mirror the command formats, required versus optional placeholder explanation, examples for `/as` and targeted `/guided`, and range-first `/hide` with optional character name. Preserve the documented legacy name-first syntax and distinguish `/unhide` global restoration from the character visibility chooser.

## Alternatives considered

This follow-up tracks translation work rather than copying English into language packs without a responsible translation review.

## Additional context

English source change: #6210. No renames or deletions are planned.

## Template check

SpicyMarinara:
## Linked issue

Closes #6211

Companion implementation: #6212 (resolves #6206, #6207, and #6208).

## Why this change

Translated guides must show the command arguments introduced in #6210 and accurately describe the per-chat persona selection used after #6212. Users should receive the same instructions regardless of their documentation language.

## What changed

- Update five guides in every documentation pack: slash commands, messages, persona selection, persona editing, and the Noodle account-switcher clarification.
- Cover required/optional arguments, range-first and legacy targeted hiding, quoted numeric names, and global versus character-specific undo behavior.
- Retire global active-persona instructions while preserving the explanation of per-chat selection, None, saved message identities, and Noodle's separate account selection.
- Align the Noodle navigation literal with English. Remove the retired Set as active UI key from the two community packs that contained it.
- Regenerate all ten documentation manifests and the UI manifest. Documentation provenance: Engine commit cec53fb6e8db544436144d0fa27495acefb35404.

## Validation

- All 50 changed guides pass exact inline/fenced-code and link-target comparisons, heading structure, NFC/whitespace checks, and retired-control checks.
- All ten documentation packs pass build-manifest and validate-pack (125 guides each).
- All eleven UI packs pass validation; unrelated existing stale keys remain unchanged and are ignored by Engine.
- Independent meaning review covers the changed Japanese, Korean, Chinese, Russian, and Hindi guides; the Latin-language changes were also reviewed against English.
- `git diff --check` passes. No application code, dependencies, or version files are changed.

- [ ] Translations reviewed against English
- [ ] Pack manifests regenerated and validated

## Manual verification notes

Manually verify downloaded guides in the app. This documentation-only PR targets docs-i18n, as required for downloadable language packs, and should land after the companion Engine PR #6212.


SpicyMarinara:
@coderabbitai review

Please review the current translation commit 5861118063b4ef4dcad69f2643d78582956d92fe. Automatic review is disabled for the docs-i18n target, but this PR needs a completed review before merge. The body records the English reference and successful pack validation.

